PR #1, opened 2023-08-14
feature: add endpoint to retrieve botanix gateway address

arminsabouri:
Tested using rpc example: ` cargo run --example rpc-db`

And `curl -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"eth_getGatewayAddress","params":[ "0xA65812Bac44dadb79C3e4930dBD98d5a75376b2A", 100, "02e2af4a49570e224fdddc6443863281ff9d96e6311547943a7628ed925e767a7a"],"id":1}' 127.0.0.1:8545`

Result: 
```json
{"jsonrpc":"2.0","result":"tb1pauwffg9tcaqx4ncf7lkmthtj8gy39tsf722k4q0nppgpach2sxsqq0hljg","id":1}
```

### DIFF
--- a/crates/rpc/rpc/src/eth/api/mod.rs
+++ b/crates/rpc/rpc/src/eth/api/mod.rs
@@ -53,7 +53,7 @@ pub trait EthApiSpec: EthTransactions + Send + Sync {
         &self,
         eth_address: Address,
         nonce: u64,
-    ) -> std::result::Result<bitcoin::Address, GatewayAddressRPCError>;
+    ) -> std::result::Result<(bitcoin::Address, secp256k1::PublicKey), GatewayAddressRPCError>;
 
     /// Returns a list of addresses owned by provider.
     fn accounts(&self) -> Vec<Address>;
@@ -253,9 +253,9 @@ where
         &self,
         eth_address: Address,
         nonce: u64,
-    ) -> std::result::Result<bitcoin::Address, GatewayAddressRPCError> {
-        let address = self.inner.botanix_provider.get_gateway_address(eth_address, nonce).await?;
-        Ok(address)
+    ) -> std::result::Result<(bitcoin::Address, secp256k1::PublicKey), GatewayAddressRPCError> {
+        let pegin_info = self.inner.botanix_provider.get_gateway_address(eth_address, nonce).await?;
+        Ok(pegin_info)
     }
 
     /// Returns the current info for the chain
--- a/crates/rpc/rpc/src/eth/api/mod.rs
+++ b/crates/rpc/rpc/src/eth/api/mod.rs
@@ -17,7 +17,6 @@ use reth_provider::{BlockReaderIdExt, EvmEnvProvider, StateProviderBox, StatePro
 use reth_rpc_types::{SyncInfo, SyncStatus};
 use reth_tasks::{TaskSpawner, TokioTaskExecutor};
 use reth_transaction_pool::TransactionPool;
-use secp256k1::PublicKey;
 use std::{future::Future, sync::Arc};
 use tokio::sync::oneshot;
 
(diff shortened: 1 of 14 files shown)


crates/rpc/rpc/src/eth/api/mod.rs
```diff
@@ -30,6 +31,18 @@ mod transactions;
 
 pub use transactions::{EthTransactions, TransactionSource};
 
+use btc_wallet::address::gateway_address;
+
+lazy_static::lazy_static! {
+    static ref SECP: secp256k1::Secp256k1<secp256k1::All> = secp256k1::Secp256k1::new();
```
stevenroose: You're creating one like this in many places now, right? I think in btc-wallet there's already two.

I mean for a server it doesn't matter aaall that much, but it's better to have like 1 at the application level (like here) and then none at the library level. Though that might mean taking a secp ctx ref on many library functions..

crates/rpc/rpc/src/eth/api/mod.rs
```diff
@@ -44,6 +57,15 @@ pub trait EthApiSpec: EthTransactions + Send + Sync {
     /// Returns provider chain info
     fn chain_info(&self) -> Result<ChainInfo>;
 
+    /// Returns gateway address
+    fn get_gateway_address(
+        &self,
+        eth_address: Address,
+        nonce: u64,
+        // TODO Hex encoded string because Bitcoin public key doesnt implement deserialize
+        aggregate_public_key: String,
```
stevenroose: It does, you need to enable the `serde` feature on the crate. But it also implements `FromStr` that is hex.

Both secp::PublicKey and bitcoin::PublicKey are json-serialized to hex with serde (they might be byte-serialized when not using the serde "human-readable" flag that is used for JSON).

crates/rpc/rpc/src/eth/botanix_config.rs
```diff
@@ -0,0 +1,84 @@
...
+    pub fn new(config: BotanixConfig) -> Self {
+        Self { botanix_rpc_config: config }
+    }
+
+    /// Returns the configuration of botanix provider
+    pub fn config(&self) -> &BotanixConfig {
+        &self.botanix_rpc_config
+    }
+
+    pub async fn get_gateway_address(
```
stevenroose, line 69: Just curiosity: when is this actually going to get called?

crates/rpc/rpc/src/eth/botanix_config.rs
```diff
@@ -0,0 +1,84 @@
...
+
+        let response = client.get_public_key(request).await.unwrap();
+        let pk_hex = response.into_inner().publickey;
+
+        let pk_vec = hex::decode(pk_hex)
+            .map_err(|e| GatewayAddressRPCError::FailedToDecodeAggregatePublicKey(e))?;
+
+        let pk = PublicKey::from_slice(pk_vec.as_slice()).map_err(|_e| {
+            GatewayAddressRPCError::InvalidParam("Failed to derive aggregate public key from input")
+        })?;
```
stevenroose, line 85: Again here you can use `.parse()` or `PublicKey::from_str`.

crates/rpc/rpc/src/eth/botanix_config.rs
```diff
@@ -0,0 +1,84 @@
...
+        let network = self.botanix_rpc_config.bitcoin_network;
+        let address = gateway_address(&SECP, &pk, &eth_address, network, nonce)
```
stevenroose: nit: personally in this case, I'd find the full path of the function `btc_wallet::address::gateway_address` more readable. I had to go grep to see if you defined another version of this method here locally.

stevenroose (commented):
I think my comments for your first commit already are outdated and removed. So ignore them :)

I review per commit usually.

Then I think the only real worthy change is optimizing that hex parsing to use the rust-bitcoin built-in parsing.

crates/rpc/rpc/src/eth/api/mod.rs
arminsabouri: Was able to get around this. Actually didn't need it as a RPC argument

crates/rpc/rpc/src/eth/botanix_config.rs
arminsabouri: This is called by any node operator that wants to display their pegin address or a pegin address for anyone else.

crates/rpc/rpc/src/eth/api/mod.rs
arminsabouri: left TODO. Reth sets up a different contexts and providers which often get provided to functions like RPC handlers. I think it would be appropriate to have secp in a crypto_context or include it in a existing provider

PR #3, opened 2023-08-29
feature: add get merkle proof rpc

arminsabouri:
```bash
curl -X POST -H "Content-Type: application/json"  --data '{"jsonrpc":"2.0","method":"eth_merkleProof","params":["4a862efb76670f57c9c2e17e0372d66f4589e5e3d39c067547f799a5f4c70314", "000000000000000b35403746160f90021e197cedea24e498b65509bdd55223b6"],"id":1}' 127.0.0.1:8545 | jq .
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100   571  100   374  100   197    658    346 --:--:-- --:--:-- --:--:--  1014
{
  "jsonrpc": "2.0",
  "result": "0x0e00000005ce88336dc1340fed95a5f334a536b459d9af3aa3f44eb7b64de3a75e01812f021403c7f4a599f74775069cd3e3e589456fd672037ee1c2c9570f6776fb2e864a3e06d4e3858fdfa9f987053290aac66ef9b7c28fcaf3d3d64724d65a5fc11a2365557cde0d0e465dbfa1f2730617416133b2347ca5170fc1c0dd86f019356acf331d671f862a2841476864ef8639511b9e6f29c25b3c150680b626f9c185be81022f00",
  "id": 1

}

```

### DIFF
--- a/crates/rpc/rpc/src/eth/api/mod.rs
+++ b/crates/rpc/rpc/src/eth/api/mod.rs
@@ -30,7 +30,7 @@ mod transactions;
 
 pub use transactions::{EthTransactions, TransactionSource};
 
-use super::botanix_config::{Botanix, GatewayAddressRPCError};
+use super::botanix_config::{Botanix, GatewayAddressRPCError, MerkleProofRPCError};
 
 
 
@@ -54,6 +54,13 @@ pub trait EthApiSpec: EthTransactions + Send + Sync {
         eth_address: Address,
         nonce: u64,
     ) -> std::result::Result<(bitcoin::Address, secp256k1::PublicKey), GatewayAddressRPCError>;
+
+    /// Returns the merkle proof for a given block hash
+    async fn get_merkle_proof(
+        &self,
+        txid: String,
+        block_hash: String,
+    ) -> std::result::Result<Vec<u8>, MerkleProofRPCError>;
 
     /// Returns a list of addresses owned by provider.
     fn accounts(&self) -> Vec<Address>;
@@ -255,6 +262,15 @@ where
         nonce: u64,
     ) -> std::result::Result<(bitcoin::Address, secp256k1::PublicKey), GatewayAddressRPCError> {
         let pegin_info = self.inner.botanix_provider.get_gateway_address(eth_address, nonce).await?;
+        Ok(pegin_info)
+    }
+
+    async fn get_merkle_proof(
+        &self,
+        txid: String,
+        block_hash: String,
+    ) -> std::result::Result<Vec<u8>, MerkleProofRPCError> {
+        let pegin_info = self.inner.botanix_provider.get_merkle_proof(txid, block_hash).await?;
         Ok(pegin_info)
     }
 
--- a/crates/rpc/rpc/src/eth/api/server.rs
+++ b/crates/rpc/rpc/src/eth/api/server.rs
@@ -115,11 +115,11 @@ where
     }
 
     /// Handler from `eth_getMerkleProof`
-    async fn merkle_proof(&self, txid: String, block_hash: String) -> Result<Option<Bytes>> {
+    async fn merkle_proof(&self, txid: String, block_hash: String) -> Result<Bytes> {
         trace!(target: "rpc::eth", ?txid, ?block_hash, "Serving eth_getMerkleProof");
         let merkle_proof = match EthApi::get_merkle_proof(self, txid, block_hash).await {
-            Ok(value) => Some(Bytes::from(value)),
-            Err(_) => None,
+            Ok(value) => Bytes::from(value),
+            Err(e) => return Err(internal_rpc_err(e)),
         };
         Ok(merkle_proof)
     }
--- a/crates/rpc/rpc/src/eth/api/server.rs
+++ b/crates/rpc/rpc/src/eth/api/server.rs
@@ -114,6 +114,16 @@ where
         Ok(address)
     }
 
+    /// Handler from `eth_getMerkleProof`
+    async fn merkle_proof(&self, txid: String, block_hash: String) -> Result<Option<Bytes>> {
+        trace!(target: "rpc::eth", ?txid, ?block_hash, "Serving eth_getMerkleProof");
+        let merkle_proof = match EthApi::get_merkle_proof(self, txid, block_hash).await {
+            Ok(value) => Some(Bytes::from(value)),
+            Err(_) => None,
+        };
+        Ok(merkle_proof)
+    }
+
     /// Handler for: `eth_getBlockTransactionCountByHash`
     async fn block_transaction_count_by_hash(&self, hash: H256) -> Result<Option<U256>> {
         trace!(target: "rpc::eth", ?hash, "Serving eth_getBlockTransactionCountByHash");
--- a/crates/rpc/rpc/src/eth/botanix_config.rs
+++ b/crates/rpc/rpc/src/eth/botanix_config.rs
@@ -1,6 +1,6 @@
 //! Defines structure for botanix RPC configurables and business logic
 
-use std::str::FromStr;
+use std::{fmt, str::FromStr};
 
 use bitcoin::consensus::Encodable;
 use btc_wallet::block_source::BlockSource;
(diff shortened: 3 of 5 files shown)


stevenroose:
nice
```
$ hal merkle proof-check 0e00000005ce88336dc1340fed95a5f334a536b459d9af3aa3f44eb7b64de3a75e01812f021403c7f4a599f74775069cd3e3e589456fd672037ee1c2c9570f6776fb2e864a3e06d4e3858fdfa9f987053290aac66ef9b7c28fcaf3d3d64724d65a5fc11a2365557cde0d0e465dbfa1f2730617416133b2347ca5170fc1c0dd86f019356acf331d671f862a2841476864ef8639511b9e6f29c25b3c150680b626f9c185be81022f00
4a862efb76670f57c9c2e17e0372d66f4589e5e3d39c067547f799a5f4c70314
```
hehe

crates/rpc/rpc/src/eth/api/mod.rs
```diff
@@ -55,6 +55,13 @@ pub trait EthApiSpec: EthTransactions + Send + Sync {
         nonce: u64,
     ) -> std::result::Result<(bitcoin::Address, secp256k1::PublicKey), GatewayAddressRPCError>;
 
+    /// Returns the merkle proof for a given block hash
+    async fn get_merkle_proof(
+        &self,
+        txid: String,
+        block_hash: String,
+    ) -> std::result::Result<Vec<u8>, MerkleProofRPCError>;
```
stevenroose, line 63: Any specific reason you use the full `std::result::Result` path instead of just `Result`?

stevenroose: Also, I admire your good practice, but I would personally not really bother to create a error type for each RPC call :sweat_smile:  But definitely doesn't hurt.

crates/rpc/rpc/src/eth/api/server.rs
```diff
@@ -114,6 +114,16 @@ where
         Ok(address)
     }
 
+    /// Handler from `eth_getMerkleProof`
+    async fn merkle_proof(&self, txid: String, block_hash: String) -> Result<Option<Bytes>> {
+        trace!(target: "rpc::eth", ?txid, ?block_hash, "Serving eth_getMerkleProof");
+        let merkle_proof = match EthApi::get_merkle_proof(self, txid, block_hash).await {
+            Ok(value) => Some(Bytes::from(value)),
+            Err(_) => None,
```
stevenroose: Why do you return None here instead of an error? I'm a bit lost in the amount of times you redefine this method lol. It seems that because of this if the internal thing errors, you won't see an RPC error, but just empty bytes, is that desired? Like that it seems even more weird to define an entire Error type with nice variants and stuff and then just ignore it instead of exposing it to the user..

crates/rpc/rpc/src/eth/botanix_config.rs
```diff
@@ -90,4 +120,36 @@ impl Botanix {
...
+        txid: String,
+        block_hash: String,
+    ) -> std::result::Result<Vec<u8>, MerkleProofRPCError> {
+        let tx_id: bitcoin::Txid = bitcoin::Txid::from_str(txid.as_str())
+            .map_err(|_e| MerkleProofRPCError::InvalidTxId)?;
+        let mempool =
+            btc_wallet::block_source::MempoolSpace::new(self.config().mempool_space_url.clone());
+
+        let txids = mempool
+            .get_txids(bitcoin::BlockHash::from_str(&block_hash).unwrap())
```
stevenroose: You have an error variant for the "txid" being wrong, but here the program will panic if the "block_hash" is wrong. At least be consistent, IMO you should have an error case here too. Programs shouldn't be able to crash because of user input.

crates/rpc/rpc/src/eth/botanix_config.rs
```diff
@@ -90,4 +120,36 @@ impl Botanix {
...
+            .await
+            .map_err(|_e| MerkleProofRPCError::FailedToGetTxIds)?;
+        if !txids.contains(&tx_id) {
+            return Err(MerkleProofRPCError::TxIdNotInBlock)
+        }
+
+        let matches = txids.iter().map(|txid| txid == &tx_id).collect::<Vec<bool>>();
```
stevenroose: nit: Usually, for using `collect`, you can hide the inner type, like so `.collect::<Vec<_>>()`. Ofc being more complete never hurts.

crates/rpc/rpc/src/eth/botanix_config.rs
```diff
@@ -90,4 +120,36 @@ impl Botanix {
...
+
+        let pmt = bitcoin::merkle_tree::PartialMerkleTree::from_txids(&txids, &matches);
+        let mut bytes = Vec::new();
+        pmt.consensus_encode(&mut bytes).map_err(|e| {
```
stevenroose: nit: You can use `bitcoin::consensus::serialize(&pmt)` which returns a `Vec<u8>` at once and cannot error, so you can skip this error case (which is actually impossible).

stevenroose: So everything from `let mut bytes` down can be replaced with simply `Ok(bitcoin::consensus::serialize(&pmt))` ;)

stevenroose (commented):
Mostly nits, no real blockers.

stevenroose:
Maybe one relevant comment is that the RPC I talked about in Elements is entirely stateless and takes block hex + txid and doesn't use a block source. My suggestion for this RPC was to do something similar, so that the ethRpc node doesn't have to rely on mempool.space, but the sidecar will do that. Either way works, though, no worries.

crates/rpc/rpc/src/eth/api/mod.rs
arminsabouri: You got it 👍

arminsabouri: Still getting use to best rust practices around this stuff

crates/rpc/rpc/src/eth/botanix_config.rs
arminsabouri: Good catch!

crates/rpc/rpc/src/eth/api/server.rs
arminsabouri: Yeah you're completely right. 
Fixed this
```bash
curl -X POST -H "Content-Type: application/json" -H "Authorization: Bearer d533e5418d7ece42f12f6552d5d259ca05c87907e79f8fa2746abe8adf938bbb" --data '{"jsonrpc":"2.0","method":"eth_merkleProof","params":["4a862efb76670f57c9c2e17e0372d66f4589e5e3d39c067547f799a5f4c70314", "00000000000b35403746160f90021e197cedea24e498b65509bdd55223b6"],"id":1}' 127.0.0.1:8545 | jq .
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   274  100    81  100   193  11339  27019 --:--:-- --:--:-- --:--:--  267k
{
  "jsonrpc": "2.0",
  "error": {
    "code": -32603,
    "message": "Malformed block hash"
  },
  "id": 1
}
```

crates/rpc/rpc/src/eth/botanix_config.rs
arminsabouri: good call

arminsabouri:
> Maybe one relevant comment is that the RPC I talked about in Elements is entirely stateless and takes block hex + txid and doesn't use a block source. My suggestion for this RPC was to do something similar so that the ethRpc node doesn't have to rely on mempool.space, but the sidecar will do that. Either way works, though, no worries.

Does the blockHex include transactions + witnesses? If so isn't that a huge payload? I'm going to merge + rebase. And revisit this suggestion

crates/rpc/rpc/src/eth/api/server.rs
stevenroose: curl wizard, very nice!

stevenroose:
> > Maybe one relevant comment is that the RPC I talked about in Elements is entirely stateless and takes block hex + txid and doesn't use a block source. My suggestion for this RPC was to do something similar so that the ethRpc node doesn't have to rely on mempool.space, but the sidecar will do that. Either way works, though, no worries.
> 
> Does the blockHex include transactions + witnesses? If so isn't that a huge payload? I'm going to merge + rebase. And revisit this suggestion

Yes it does. So yeah it can be big. Like up to 4MB. Alternatively you can just take a list of all the txids in the block (like `hal` also does). That's also what bitcoin core's rpc can give you and what most block explorers will give you.

arminsabouri:
Yeah I agree. A list of txids makes the most sense

PR #5, opened 2023-09-08
Arm/genesis contract prefund

arminsabouri:
- [x] Pull agg pk from crate
- [x] Call btc server to add utxo to db

### DIFF
--- a/crates/consensus/auto-seal/src/task.rs
+++ b/crates/consensus/auto-seal/src/task.rs
@@ -1,5 +1,7 @@
 use crate::{mode::MiningMode, Storage};
-use botanix_lib::mint_validation::{parse_reth_log_topic, GenesisContractEvents};
+use botanix_lib::mint_validation::{
+    parse_pegin_reth_log_topic, parse_pegout_reth_log_topic, GenesisContractEvents,
+};
 use futures_util::{future::BoxFuture, FutureExt};
 use reth_beacon_consensus::{BeaconEngineMessage, ForkchoiceStatus};
 use reth_interfaces::consensus::ForkchoiceState;
@@ -22,7 +24,7 @@ use tokio::sync::{mpsc::UnboundedSender, oneshot};
 use tokio_stream::wrappers::UnboundedReceiverStream;
 use tracing::{debug, error, info, warn};
 
-use client::{BtcServerClient, NotifyPeginRequest};
+use client::{BtcServerClient, MakeTxRequest, NotifyPeginRequest};
 
 /// A Future that listens for new ready transactions and puts new blocks into storage
 pub struct MiningTask<Client, Pool: TransactionPool> {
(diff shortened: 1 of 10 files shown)


crates/revm/src/executor.rs
```diff
@@ -239,6 +251,67 @@ where
...
+                    Ok(pegin_data) => {
+                        let block_hash = pegin_data.meta.block_header.block_hash();
+                        let (sender, mut receiver) = mpsc::channel(1);
+                        // Create a Tokio runtime
+                        let rt = tokio::runtime::Handle::current();
+
+                        let manager = TaskManager::new(rt.clone());
+                        let executor = manager.executor();
+
+                        let _task = TaskSpawner::spawn_blocking(
```
stevenroose, line 269: This doesn't use tokio? With tokio it would be `tokio::spawn(|| {})` to enter the current runtime.

crates/revm/src/executor.rs
```diff
@@ -239,6 +251,67 @@ where
...
+                        // TODO ideally this function runs in the same runtime as the caller.
+                        // This is easier said than done, so it may be easier to have a async task in a seperate runtime
+                        // this task will recieve jobs from the caller and send the result back to the caller.
+                        let mut counter = 0;
+                        let ten_millis = time::Duration::from_millis(10);
+                        loop {
+                            if counter == 10_000 {
+                                return Err(BlockExecutionError::FailedToGetBitcoinHeader);
+                            }
+                            if let Ok(block_header) = receiver.try_recv() {
```
stevenroose, line 290: The btc block header should come from the eth block header, no? Or is this for mempool?

crates/consensus/auto-seal/src/task.rs
```diff
@@ -157,6 +154,49 @@ where
...
+                                                eth_address: hex::encode(
+                                                    pegin_data.meta.address.to_vec(),
+                                                ),
+                                                output: bitcoin::consensus::serialize(
+                                                    &pegin_data
+                                                        .meta
+                                                        .tx
+                                                        .output
+                                                        .get(pegin_data.meta.outpoint.vout as usize)
+                                                        .unwrap(),
```
stevenroose: Ugg these crazy verbose lines :o

stevenroose:
I'm confused why you're involving mempool.space in consensus :sweat_smile:  That doesn't seem right. Is this for mempool? I'm pretty sure there must be some way to get a "eth block header context" or something for mempool validation. The same way the mempool execution would know what the block height would be etc so that it can validate things like CHECKLOCKTIMEVERIFY (but then the eth variant).

crates/revm/src/executor.rs
arminsabouri: Reth has its own wrapper for tokio. Tends to work well with other reth things

crates/consensus/auto-seal/src/task.rs
arminsabouri: 🤮

arminsabouri:
> I'm confused why you're involving mempool.space in consensus 😅 That doesn't seem right. Is this for mempool? I'm pretty sure there must be some way to get a "eth block header context" or something for mempool validation. The same way the mempool execution would know what the block height would be etc so that it can validate things like CHECKLOCKTIMEVERIFY (but then the eth variant).

Its not mempool.space necessarily but a block source. 
For the single block producer model we aren't going to store the headers in the botanx block headers but they need to be available at execution checktime. I agree a bitcoinBlockHeader Context would be more appropriate. Drafting till I have that

PR #7, opened 2023-09-21
add minting.sol bytecode in config

scottmillner:


arminsabouri (approved):

stevenroose:
Nice! @scottmillner this is what Armin and I were wondering about for a while initially haha, cool to see this happen!

PR #8, opened 2023-09-22
fix: include c-kzg serde feature

arminsabouri:
Builds are broken on Debian


scottmillner (approved):

arminsabouri:
This ended up breaking the builds on macos. More investigation needed.

PR #9, opened 2023-09-25
update bytecode with sats to wei conversion logic

scottmillner:
Bytecode in config now included sats to wei conversion logic
Contract is initialized with 21 million ether


arminsabouri (approved):

stevenroose:
Oh it's unfortunate that the contract and the bytecode don't live in the same repo, so that they can be updated in atomic commits :/

PR #10, opened 2023-09-25
111 feature miner task should take btc server and block source details from config

arminsabouri:


scottmillner (approved):

arminsabouri:
Pegout seem to be broken. 
drafting for now

PR #11, opened 2023-09-26
add getBtcFees endpoint

scottmillner:


arminsabouri (approved):
🙌 🙌 🙌

stevenroose:
What does this method actually return? A static fee value? Or a feerate? How is the feerate expressed, if so? The documentation reads "Returns the Bitcoin fees for a pegout transaction" which doesn't really mean anything.

arminsabouri:
Yes its a static value until we move the fee to the botanix block headers. 
>The documentation reads "Returns the Bitcoin fees for a pegout transaction" which doesn't really mean anything.

Agreed. Documentation should specify fee rate

stevenroose:
I'll add it. In other places could be done too.

stevenroose:
https://github.com/botanix-labs/Macbeth/pull/15

PR #13, opened 2023-10-06
feature: create async bitcoin worker

arminsabouri:
Created a seperate bitcoin worker background thread to write bitcoin block headers to a shared reader/writer locked datastructure

- [x] Read block source URL from cli


bin/reth/src/node/mod.rs
```diff
@@ -251,6 +253,35 @@ impl<Ext: RethCliExt> NodeCommand<Ext> {
             BtcServerClient::connect(self.rpc.btc_server.clone()).await.expect("connect to btc_server");
         info!(target: "reth::cli", "Btc server connected");
 
+        let bitcoin_block_headers: Arc<RwLock<Vec<bitcoin::block::Header>>> =
+            Arc::new(RwLock::new(Vec::new()));
```
stevenroose, line 258: Should be enough to just have the "latest", no?

crates/consensus/auto-seal/src/task.rs
```diff
@@ -49,7 +49,10 @@ pub struct MiningTask<Client, Pool: TransactionPool> {
     canon_state_notification: CanonStateNotificationSender,
     /// The pipeline events to listen on
     pipe_line_events: Option<UnboundedReceiverStream<PipelineEvent>>,
+    /// BTC Server client
     btc_server: BtcServerClient<tonic::transport::Channel>,
+    /// Recent bitcoin block headers
+    bitcoin_block_headers :Arc<RwLock<Vec<bitcoin::block::Header>>>,
```
stevenroose: nit: typo in the position of the space here

crates/consensus/auto-seal/src/task.rs
```diff
@@ -125,11 +130,13 @@ where
                 let events = this.pipe_line_events.take();
                 let canon_state_notification = this.canon_state_notification.clone();
                 let mut btc_server = this.btc_server.clone();
+                let bitcoin_block_headers = this.bitcoin_block_headers.clone();
                 // Create the mining future that creates a block, notifies the engine that drives
                 // the pipeline
                 this.insert_task = Some(Box::pin(async move {
                     let block_source =
                         MempoolSpace::new("https://mempool.space/testnet".to_string());
+                    let recent_block_headers= bitcoin_block_headers.read().await.clone();
```
stevenroose: nit: another type of the spacing here

stevenroose (commented):
Should we maybe have a call to re-hash the pegin flow and proof system?

Maybe I'm on the wrong page and ideas have changed since our last convos.

Or maybe I'll go over the pegin spec document again to see if there are things different than what's in my head.

bin/reth/src/node/mod.rs
arminsabouri: >Should we maybe have a call to re-hash the pegin flow and proof system?

I think we should. In the first iteration of pegin (live now) the user does not provide multiple block headers.

This is changing in the next 2 weeks. Where we store only the 6th deep block hash and the user has to provide all the block headers that link to the 6th deep header

stevenroose: Ooh, ok, yeah let's talk on monday.

scottmillner (approved):

PR #14, opened 2023-10-10
fix: pass `B160` as vec<u8> to btc_wallet

arminsabouri:
Related to https://github.com/botanix-labs/botanix/pull/143

we no longer use the reth address types. This was causing a circular dependency


scottmillner (approved):

arminsabouri:
Fixing this on #13

PR #15, opened 2023-10-11
rpc: Rename RPC method getBtcFees to GetBtcFeeRate and change usage

stevenroose:
Depends on https://github.com/botanix-labs/botanix/pull/151.

Does this affect SideCar at all? I don't fully grasp who is using these RPC methods, they seem like they might be used internally only..

### DIFF
--- a/crates/consensus/auto-seal/src/task.rs
+++ b/crates/consensus/auto-seal/src/task.rs
@@ -215,14 +215,14 @@ where
                                             }
                                             Ok(GenesisContractEvents::BurnEvent) => {
                                                 // TODO (armins): obv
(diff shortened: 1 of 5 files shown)


arminsabouri (approved):
ACK
Thanks for making this change
This will break the Dapp I believe. @scottmillner

scottmillner:
> ACK Thanks for making this change This will break the Dapp I believe. @scottmillner



> ACK Thanks for making this change This will break the Dapp I believe. @scottmillner

it'll be small update in the dApp...thanks for letting me know!

arminsabouri (approved):
Nice

PR #12, opened 2023-09-28
executor checks if btc address is valid

scottmillner:
@0xBEEFCAF3 first pass...on the right track?

PR #16, opened 2023-10-11
executor: Fix various issues in Mint contract validation

stevenroose:
Closes https://github.com/botanix-labs/botanix/issues/154, https://github.com/botanix-labs/botanix/issues/155.

Depends on https://github.com/botanix-labs/botanix/pull/166.

Triggered by reading https://github.com/botanix-labs/Macbeth/pull/12, I found some bugs in the validation code.

I tried to break it into various commits, but it was so many details that I'm going to leave comments in the code instead. That way we can also have discussions in these comment threads.

This code is so critical for this system, it should really be reviewed carefully.

### DIFF
--- a/crates/interfaces/src/executor.rs
+++ b/crates/interfaces/src/executor.rs
@@ -24,6 +24,8 @@ pub enum BlockValidationError {
     BlockPreMerge { hash: H256 },
     #[error("Missing total difficulty")]
     MissingTotalDifficulty { hash: H256 },
+    #[error("Invalid Mint contract invocation")]
+    MintContractViolation,
 }
 
 /// BlockExecutor Errors
@@ -58,6 +60,7 @@ pub enum BlockExecutionError {
     #[error("Execution unavailable for tests")]
     UnavailableForTest,
 
+    //TODO(stevenroose) this should be removed in later versions
     /// Failed to get bitcoin header
     #[error("Failed to get bitcoin header")]
     FailedToGetBitcoinHeader,
--- a/crates/revm/src/executor.rs
+++ b/crates/revm/src/executor.rs
@@ -6,7 +6,9 @@ use crate::{
     stack::{InspectorStack, InspectorStackConfig},
     to_reth_acc,
 };
-use botanix_lib::mint_validation::{parse_pegin_topic, parse_pegout_topic, MINT_TOPIC, BURN_TOPIC};
+use botanix_lib::mint_validation::{
+    parse_pegin_topic, parse_pegout_topic, MINT_CONTRACT_ADDRESS, MINT_TOPIC, BURN_TOPIC,
+};
 use reth_consensus_common::calc;
 use reth_interfaces::executor::{BlockExecutionError, BlockValidationError};
 use reth_primitives::{
@@ -19,11 +21,11 @@ use revm::{
     db::{AccountState, CacheDB, DatabaseRef},
     primitives::{
         hash_map::{self, Entry},
-        Account as RevmAccount, AccountInfo, ResultAndState,
+        Account as RevmAccount, AccountInfo, ExecutionResult, ResultAndState,
     },
     EVM,
 };
-use tracing::warn;
+use tracing::{error, warn};
 use std::{
     collections::{BTreeMap, HashMap},
     sync::Arc,
@@ -59,6 +61,47 @@ where
         let evm = EVM::new();
         Executor { chain_spec, evm, stack: InspectorStack::new(InspectorStackConfig::default()) }
     }
+}
+
+fn botanix_mint_contract_checks(
+    result: &ExecutionResult,
+    recent_block_headers: Option<Vec<bitcoin::block::Header>>,
+) -> Result<(), BlockExecutionError> {
+    for log in result.logs() {
+        let block_source_clone = BLOCK_SOURCE.to_owned();
+
+        if log.topics.get(0) == Some(&MINT_TOPIC) {
+            let pegin_data = parse_pegin_topic(&log).map_err(|e| {
+                error!("Failed to parse pegin topic! {:?}", e);
+                BlockValidationError::MintContractViolation
+            })?;
+
+            let blockhash = pegin_data.meta.block_header.block_hash();
+            let pegin_header = recent_block_headers
+                //TODO(stevenroose) under what circumstances would we not have them?
+                //we could alternatively return the same error as on violation
+                .expect("can't pegin without recent headers known")
+                .iter().find(|h| h.block_hash() == blockhash)
+                .ok_or_else(|| {
+                    warn!("Failed pegin attempt Could not find block header in list of recent headers");
+                    BlockExecutionError::FailedToGetBitcoinHeader
+                })?;
+
+            if let Err(e) = pegin_data.validate(&SECP, &pegin_header) {
+                warn!("Failed pegin attempt! {:?}", e);
+                return Err(BlockValidationError::MintContractViolation.into());
+            }
+        }
+
+        if log.topics.get(0) == Some(&BURN_TOPIC) {
+            if let Err(e) = parse_pegout_topic(&log) {
+                error!("Failed to parse pegout topic! {:?}", e);
+                return Err(BlockValidationError::MintContractViolation.into());
+            }
+        }
+    }
+
+    Ok(())
 }
 
 impl<DB> Executor<DB>
@@ -191,6 +234,7 @@ where
         &mut self,
         transaction: &TransactionSigned,
         sender: Address,
+        recent_block_headers: Option<Vec<bitcoin::block::Header>>,
     ) -> Result<ResultAndState, BlockExecutionError> {
         // Fill revm structure.
         fill_tx_env(&mut self.evm.env.tx, transaction, sender);
@@ -209,6 +253,14 @@ where
             // main execution.
             self.evm.transact()
         };
+
+        // Botanix mint contract validation
+        if let Ok(ResultAndState { ref result, .. }) = out {
+            if result.is_success() && transaction.to() == Some(*MINT_CONTRACT_ADDRESS) {
+                botanix_mint_contract_checks(&result, recent_block_headers)?;
+            }
+        }
+
         out.map_err(|e| BlockValidationError::EVM { hash, message: format!("{e:?}") }.into())
     }
 
@@ -240,8 +292,6 @@ where
         let mut cumulative_gas_used = 0;
         let mut post_state = PostState::with_tx_capacity(block.number, block.body.len());
         for (transaction, sender) in block.body.iter().zip(senders) {
-            let mut pegin_fail = false;
-            let mut pegout_fail = false;
             // The sum of the transaction’s gas limit, Tg, and the gas utilised in this block prior,
             // must be no greater than the block’s gasLimit.
             let block_available_gas = block.header.gas_limit - cumulative_gas_used;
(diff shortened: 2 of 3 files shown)


crates/interfaces/src/executor.rs
```diff
@@ -24,6 +24,8 @@ pub enum BlockValidationError {
     BlockPreMerge { hash: H256 },
     #[error("Missing total difficulty")]
     MissingTotalDifficulty { hash: H256 },
+    #[error("Invalid Mint contract invocation")]
+    MintContractViolation,
```
stevenroose, line 28: Added an error variant for this because failing the mint contract check is a consensus violation and should fail block validation.

crates/revm/src/executor.rs
```diff
@@ -61,6 +63,80 @@ where
     }
 }
 
+fn botanix_mint_contract_checks(result: &ExecutionResult) -> Result<(), BlockExecutionError> {
```
stevenroose: Bundling all our stuff into a single method (or file even better), makes it a lot easier lateron to merge in upstream changes because it minimizes the actual diff inside the upstream code.

crates/revm/src/executor.rs
```diff
@@ -61,6 +63,80 @@ where
...
+    for log in result.logs() {
+        let block_source_clone = BLOCK_SOURCE.to_owned();
+
+        if log.topics[0] == *MINT_TOPIC {
```
stevenroose: Made this an explicit check for the first topic. Can someone confirm that all logs have at least one topic? Because if that is not the case, this code could potentially panic.

crates/revm/src/executor.rs
```diff
@@ -61,6 +63,80 @@ where
...
+                                }
+                                thread::sleep(time::Duration::from_millis(10));
+                                continue;
+                            }
+                        }
+                    }
+                }
+                Err(e) => {
+                    error!("Failed to parse pegin topic! {:?}", e);
+                    return Err(BlockValidationError::MintContractViolation.into());
```
stevenroose: This error case was previously just a `continue`, which would mean that the error would simply be ignored. Now, these errors are a bit special, they shouldn't really happen and they would likely mean that our own interpretation of the log structure is wrong. But ignoring is definitely not a good idea. It's hard to evaluate which of these errors can be triggered by users.

crates/revm/src/executor.rs
```diff
@@ -61,6 +63,80 @@ where
...
+                            Ok(Some(header)) => {
+                                if let Err(e) = pegin_data.validate(&SECP, &header) {
+                                    warn!("Failed pegin attempt! {:?}", e);
+                                    return Err(BlockValidationError::MintContractViolation.into());
+                                }
+                                break
+                            }
+                            Ok(None) | Err(mpsc::error::TryRecvError::Disconnected) => {
+                                return Err(BlockExecutionError::FailedToGetBitcoinHeader);
+                            }
```
stevenroose: Without this block the the above `let ret = block_source_clone.get_block_header(block_hash).await.ok();`, which turns the channel value into an `Option`, the block source failing would cause that spinloop to go on for 100 seconds and then timing out instead of erroring out early.

crates/revm/src/executor.rs
```diff
@@ -61,6 +63,80 @@ where
...
+                    // Create a Tokio runtime
+                    let rt = tokio::runtime::Handle::current();
+
+                    let manager = TaskManager::new(rt.clone());
+                    let executor = manager.executor();
+
+                    //TODO(stevenroose) this should be removed in later stages and be retrieved
+                    // directly from the execution context (botanix block header) and then this
+                    // method can simply return a boolean and the error type specified at the
+                    // callsite
```
stevenroose: This spinloop thing should really really not be here. I think you do it because for the first testnet you don't want to add any data to the botanix headers, do I remember that correctly?

crates/revm/src/executor.rs
```diff
@@ -61,6 +63,80 @@ where
...
+                                    return Err(BlockValidationError::MintContractViolation.into());
+                                }
+                                break
+                            }
+                            Ok(None) | Err(mpsc::error::TryRecvError::Disconnected) => {
+                                return Err(BlockExecutionError::FailedToGetBitcoinHeader);
+                            }
+                            Err(mpsc::error::TryRecvError::Empty) => {
+                                counter += 1;
+                                if counter == 10_000 {
```
stevenroose: Is a 100 seconds not a lot of time?

crates/revm/src/executor.rs
```diff
@@ -209,6 +285,14 @@ where
             // main execution.
             self.evm.transact()
         };
+
+        // Botanix mint contract validation
+        if let Ok(ResultAndState { ref result, .. }) = out {
+            if result.is_success() && transaction.to() == Some(*MINT_CONTRACT_ADDRESS) {
```
stevenroose, line 259: So two things here: (1) most obviously, we only do our checks if this tx is a call to our mint contract. Any other contract we don't care about. But (2) this code misinterpreted the concept of "tx success"; check also the changed line below in the Receipt struct.

The previous code would just change the receipt of the tx. This literally has no effect on state. That line below basically disregards all our checks and just makes the receipt show unsuccesful but that has absolutely no effect. Any user could just pegin any amount in the system without a valid proof 😅

What the success flag means is whether the solidity execution of the tx was succesful. If not, it means no state changes have happened. So basically if an error already occurred in Solidity, the state is reverted and we also don't have to do any further checks. (Tbh I would presume no logs would be produced either so that's probably not a problem.)

crates/revm/src/executor.rs
```diff
@@ -209,6 +285,14 @@ where
...
+                botanix_mint_contract_checks(&result)?;
```
stevenroose: This is maybe the most important change. The error return here. Before, mint check failure would just set the success flag in the receipt to false, but txs would still be accepted. To actually invalidate the tx, we need to somehow upstream indicate that a consensus rule was violated. That's what this error return does here.

crates/interfaces/src/executor.rs
```diff
@@ -58,6 +60,7 @@ pub enum BlockExecutionError {
     #[error("Execution unavailable for tests")]
     UnavailableForTest,
 
+    //TODO(stevenroose) this should be removed in later versions
```
arminsabouri, line 63: What is expected from putting your username by a TODO? Are you signifying you made it or your going to solve it?

arminsabouri: Regardless I think we need this variant independent of where we are retrieving block headers

crates/revm/src/executor.rs
arminsabouri: This was removed in #13

crates/revm/src/executor.rs
arminsabouri: Again removed in #13

crates/revm/src/executor.rs
arminsabouri: These are two different things
If you return an Err from the executor the whole bundle of txs will be rejected. But still in the mempool.

If you set the tx receipt to false, other txs can me mined while this one is dropped from the mempool

crates/revm/src/executor.rs
arminsabouri: >most obviously, we only do our checks if this tx is a call to our mint contract.

This check is in peginData.validate(). Do we need it again here?

crates/revm/src/executor.rs
arminsabouri: This mechanism was removed in #13

crates/revm/src/executor.rs
arminsabouri: >Any user could just pegin any amount in the system without a valid proof 😅

This isn't entirely true if the tx fails execution (receipt.success == false) it will not be mined.

arminsabouri (changes requested):
There are some good points being made but largely this code conflicts with #13 
Let's merge #13 in first and revisit this.

crates/revm/src/executor.rs
stevenroose: > > most obviously, we only do our checks if this tx is a call to our mint contract.
> 
> This check is in peginData.validate(). Do we need it again here?

I don't think this is true. peginData.validate has no context to make this check

stevenroose: > > Any user could just pegin any amount in the system without a valid proof 😅
> 
> This isn't entirely true if the tx fails execution (receipt.success == false) it will not be mined.

How so? Failed txs in ethereum are mined, they don't have effect on state, but the gas is paid and they are part of the block. AFAIK

crates/interfaces/src/executor.rs
stevenroose: > What is expected from putting your username by a TODO? Are you signifying you made it or your going to solve it?

Oh, I think it's a convention, this happens automatically if I type "todo". I think it serves mostly to indicate who made the todo so that if you have questions about it, you know who to talk with.

stevenroose: > Regardless I think we need this variant independent of where we are retrieving block headers

Could you explain in what kind of scenario that would be?

crates/revm/src/executor.rs
stevenroose: Hmm, I might not correctly understand the context here then. Does the mempool call the `.transact(tx, sender)` method of the multi `.execute_transactions(block)` method? Ok I checked and it seems the `.transact()` method is only called from within `.execute_transactions()` so my second commit might not be needed. I have to check better what context this code is run in. Because Ethereum for sure mines transactions that fail execution because they have to pay gas.

crates/revm/src/executor.rs
scottmillner: that's correct, all logs have at least one topic
an anonymous event won't have a signature topic but still has at least one topic

stevenroose: Anonymous events? I didn't know about those. How do they work?

stevenroose: Thanks for checking, btw!

scottmillner: np! 
anonymouse events are rarely used (ive never used one)
if you put the `anonymous` keyword after the event in the contract, the signature topic is omitted so a client can't search for it easily...not sure the benefit of that lol

stevenroose: But then if you do `anonymous event NoArguments();`, won't that have 0 topics?

scottmillner: good point! yeah best to revert this change

stevenroose: changed

stevenroose:
Updated this with the changes you just merged, Armin.

crates/interfaces/src/executor.rs
arminsabouri: When #18 gets merged this won't be an issue anymore.
Resolving

stevenroose: I can rebase this after that.

crates/revm/src/executor.rs
```diff
@@ -61,6 +63,47 @@ where
...
+            let pegin_data = parse_pegin_topic(&log).map_err(|e| {
+                error!("Failed to parse pegin topic! {:?}", e);
+                BlockValidationError::MintContractViolation
+            })?;
+
+            let blockhash = pegin_data.meta.block_header.block_hash();
+            let pegin_header = recent_block_headers
+                //TODO(stevenroose) under what circumstances would we not have them?
+                //we could alternatively return the same error as on violation
+                .expect("can't pegin without recent headers known")
```
arminsabouri, line 83: I can update this code when after #18  gets merged

crates/revm/src/executor.rs
```diff
@@ -209,6 +253,14 @@ where
             // main execution.
             self.evm.transact()
         };
+
+        // Botanix mint contract validation
```
arminsabouri, line 257: Nit: these checks cover both mint and withdrawals

arminsabouri (approved):
Nice! I can rebase #18 after this gets merged.

crates/revm/src/executor.rs
arminsabouri: On second thought I don't think this Mr should make this change the whole file calls the genesis contract a "mint contract". So we'll refactor the names as somepoint

PR #18, opened 2023-10-17
refactor(async_worker): only store most recent bitcoin block header

arminsabouri:
This is actually the original design as defined in https://github.com/botanix-labs/botanix/blob/master/docs/pegin.md
User provides n block headers. The nth block header should be the tip that we store in reth. For v0 we only care about 1 conf. Going forward it will be m confs.

in a future PR. We will mine the block with the mth bitcoin block header in the botanix block header.

### DIFF
--- a/bin/reth/src/node/mod.rs
+++ b/bin/reth/src/node/mod.rs
@@ -254,8 +254,8 @@ impl<Ext: RethCliExt> NodeCommand<Ext> {
                 .expect("connect to btc_server");
         info!(target: "reth::cli", "Btc server connected");
 
-        let bitcoin_block_headers: Arc<RwLock<Vec<bitcoin::block::Header>>> =
-            Arc::new(RwLock::new(Vec::new()));
+        let bitcoin_block_headers: Arc<RwLock<Option<bitcoin::block::Header>>> =
+            Arc::new(RwLock::new(None));
         let bitcoin_block_headers_clone = bitcoin_block_headers.clone();
         let block_source = MempoolSpace::new(self.rpc.btc_block_source.to_string().clone());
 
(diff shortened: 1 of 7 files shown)


stevenroose:
After https://github.com/botanix-labs/botanix/pull/178 this can just be block hash, right?

bin/reth/src/node/mod.rs
```diff
@@ -272,10 +272,10 @@ impl<Ext: RethCliExt> NodeCommand<Ext> {
                         let block_hash = block_source.get_block_hash(current_tip).await.unwrap();
                         let block_header = block_source.get_block_header(block_hash).await.unwrap();
 
-                        header_write.push(block_header);
+                        // TODO (armins) in v1 we will need the nth deep block header not tip
+                        *header_write = Some(block_header);
```
stevenroose, line 276: You want the one that is 6 deep here, though, right? Instead of the tip..

stevenroose (changes requested):
This seems to be already incorrect currently, but you want the 6-deep block hash right, not the current tip?

bin/reth/src/node/mod.rs
arminsabouri: For v0 we want the tip. This is a product decision. V1 we will update that to 6 blocks deep

stevenroose: ah ok sorry didn't realize that. :+1:

stevenroose (approved):

bin/reth/src/node/mod.rs
arminsabouri: All good All good 👍 I left a comment above it

PR #20, opened 2023-10-25
Scott/195 feature proper signing of blocks to authorize

arminsabouri:


arminsabouri:
The last commit will get replaced with one that implements the block signing code in authority consensus crate

arminsabouri:
closing infavor of #23

PR #23, opened 2023-10-30
Arm/dyna fed voting

arminsabouri:


crates/consensus/authority/src/voting.rs
```diff
@@ -0,0 +1,156 @@
...
+        let sig_hash = secp256k1::Message::from_slice(
+            create_authority_sighash(&header, &extra_data_header).unwrap().as_slice(),
+        )
+        .map_err(|e| GetVotesError::FailedToDeserializeBlockHeaderExtraData(e))?;
+        let authority_that_votes = extra_data_header
+            .authority_signature
+            .expect("valid signature")
+            .recover(&sig_hash)
+            .map_err(|e| GetVotesError::FailedToRecoverAuthority(e))?;
+        // Already keeping track of this authority
```
scottmillner, line 132: - `auth_vote.contains(&authority_to_vote_on)` will this work?
- this is `Vec<AuthorityVote>` searching for `PublicKey`

this could just be my limited rust knowledge

arminsabouri: > will this work?

No, it won't 
I keep running into things like this so I started to abstract `Vec<AuthorityVote>` to a new struct in voting. I will probably replace this

crates/consensus/authority/src/voting.rs
```diff
@@ -0,0 +1,125 @@
...
+    Add,
+    Remove,
+}
+
+impl TryFrom<u64> for Vote {
+    type Error = &'static str;
+    fn try_from(value: u64) -> Result<Self, Self::Error> {
+        match value {
+            NONCE_AUTH => Ok(Vote::Add),
+            NONCE_DROP => Ok(Vote::Remove),
```
stevenroose, line 29: Seems weird to me that these constants exists in EIP spec but the code for vote counting doesn't?

crates/consensus/authority/src/voting.rs
```diff
@@ -0,0 +1,125 @@
...
+    // Structure to keep track of all votes that occured in this block range
+    let mut auth_vote: Vec<AuthorityVote> = Vec::new();
+
+    for header in headers {
+        if header.is_empty() {
+            continue
+        }
+        // Check if there is a authority being voted on in the extra data
+        let extra_data_header = ExtraDataHeader::deserialize(header.extra_data.as_slice())
+            .map_err(|e| GetVotesError::FailedToDeserializeBlockHeaderExtraData(e))?;
```
stevenroose, line 102: So did this get merged implicitly with this MR? Does that mean that this MR is also merged implicitly? https://github.com/botanix-labs/botanix/pull/198 Or are you merging into a feature branch? :thinking:

crates/consensus/authority/src/voting.rs
```diff
@@ -0,0 +1,125 @@
...
+        }
+
+        let authority = extra_data_header.authority_vote.expect("valid authority vote");
+        // Already keeping track of this authority
+        if auth_vote.contains(&authority) {
+            // Need to recover the authority that signed the block from the signature
+            let sig_hash = secp256k1::Message::from_slice(
+                create_authority_sighash(&header, &extra_data_header).unwrap().as_slice(),
+            )
+            .map_err(|e| GetVotesError::FailedToDeserializeBlockHeaderExtraData(e))?;
```
stevenroose: `Message::from_slice` doesn't produce meaningful errors if the input data is hashed (not user-inputted) and is 32 bytes.

So in this case, you can just `unwrap`/`expect` it instead of throwing the error up.

That's why there is `Message::from_32_byte_hash` method as well that doesn't throw an error but lets you run in hoops to implement certain types. It's a bit annoying, but I think you should just unwrap the error here.

stevenroose: Sorry for the delay in review, btw :/ my bad

crates/consensus/authority/src/voting.rs
```diff
@@ -0,0 +1,125 @@
...
+            Vote::Add => add_votes += 1,
+            Vote::Remove => remove_votes += 1,
+        }
+    }
+
+    if add_votes > remove_votes {
+        Vote::Add
+    } else {
+        Vote::Remove
+    }
```
stevenroose, line 169: Wait how does the voting worked? I thought `ADD` was to add someonne and you need to have a certain threshold of the epoch to vote add in order to get added and `REMOVE` was to remove someone that was already in and you'd need a certain threshold of votes to get removed.. Like, here it seems like it's `ADD` _**against**_ `REMOVE`..

crates/consensus/authority/src/voting.rs
arminsabouri: Ah ok noted.

crates/consensus/authority/src/voting.rs
arminsabouri: can you elaborate? Not entirely sure what you mean

crates/consensus/authority/src/voting.rs
arminsabouri: Into a feature branch

crates/consensus/authority/src/voting.rs
arminsabouri: Yeah this was code for an earlier design that we are no longer following. In the newer design you need a threshold of votes to either add OR remove. There aren't votes against or for

crates/consensus/authority/src/voting.rs
stevenroose: You can use `.iter().find()` or something which is pretty much the same.

crates/consensus/authority/src/voting.rs
stevenroose: It seems that `NONCE_AUTH` and `NONCE_DROP` are voting-related constants specified in some ethereum-native library, but you are implementing the logic for counting votes etc. Wouldn't that be part of the same EIP as the constants so I would expect some implementation to exist alongside the constants.

crates/consensus/authority/src/voting.rs
stevenroose: oooh, makes sense, ok :pray:

crates/consensus/authority/src/voting.rs
stevenroose: But this got merged right? So it's changed again already? I'll catch up with MRs, so feel free to ignore this if it's already changed.

crates/consensus/authority/src/voting.rs
arminsabouri: These constants live ni the added in the eip-255 constants file which we created

stevenroose: Aaah, you created the file? Ok, sorry then it makes sense.

PR #29, opened 2023-11-06
feature: Networking

cdxker:
Updated networking to merge with the new base

### DIFF
--- a/crates/net/network/src/import.rs
+++ b/crates/net/network/src/import.rs
@@ -1,6 +1,15 @@
 use crate::message::NewBlockMessage;
-use reth_primitives::PeerId;
-use std::task::{Context, Poll};
+use reth_beacon_consensus::BeaconEngineMessage;
+use reth_eth_wire::NewBlock;
+use reth_interfaces::consensus::{Consensus, ConsensusError};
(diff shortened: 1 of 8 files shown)


crates/net/network/src/import.rs
```diff
@@ -64,3 +73,57 @@ impl BlockImport for ProofOfStakeBlockImport {
...
+
+    /// Vaidates a header on block import
+    fn validate_header(&mut self, _header: Header) -> Result<(), ConsensusError> {
+        // TODO (networking) This will need to be updated with more checks
+        Ok(())
+    }
+
+    /// Validates a block on block import
+    fn validate_new_block(&mut self, block: Arc<NewBlock>) -> Result<(), ConsensusError> {
+        if block.td != DIFF_INTURN && block.td != DIFF_NOTURN && block.td != DIFF_NOVOTE {
```
arminsabouri, line 100: This logic is fine here for now, we have most of the block/ block header validation logic in crates/consensus/authority/lib.rs

arminsabouri: In the future we probably want to just call into that

arminsabouri (approved):

PR #31, opened 2023-11-15
feature(authority): write block import messages on a shared queue

arminsabouri:
PoA consensus will consume block import messages and notify the beacon engine of new blocks. As well as update its own cache and prune the mempool


crates/consensus/authority/src/task.rs
```diff
@@ -105,6 +115,79 @@ where
...
+                                        let body = BlockBody {
+                                            transactions: block.body,
+                                            ommers: vec![],
+                                            withdrawals: None,
+                                        };
+
+                                        storage.insert_new_block(block.header, body);
+                                        let mining_pool = self.pool.clone();
+                                        // Lastly remove confirmed txs from the mempool
+                                        mining_pool.remove_transactions(
```
stevenroose, line 251: nit: why do you clone the pool Arc instead of just do `self.pool.remove_transactions()`?

crates/consensus/authority/src/task.rs
```diff
@@ -105,6 +115,79 @@ where
...
+                    loop {
+                        let block = new_block.block.block.clone();
+                        // send the new update to the engine, this will trigger the engine
+                        // to download and execute the block we just inserted
+                        let (tx, rx) = oneshot::channel();
+                        let _ = self.to_engine.send(BeaconEngineMessage::NewPayload {
+                            payload: block.clone().seal_slow().into(),
+                            cancun_fields: None,
+                            tx,
+                        });
```
stevenroose, line 136: I think you should panic here on the error for `send`. It means that the receiving thread has stopped (probably paniced, but might be gracefully shutting down). I'm not 100% sure but otherwise the `await` down there might block this thread forever.

crates/consensus/authority/src/task.rs
```diff
@@ -105,6 +115,79 @@ where
         loop {
             let mut storage = self.storage.write().await;
 
+            match self.block_import_rx.try_recv() {
+                Ok(new_block) => {
```
stevenroose, line 121: Just a style thing, but I think it's easier to follow the code if the error cases from this method are not like 100 lines back down. Also, leaving the `match` asap saves you 2 levels of indentation. I would write this like
```rust
            let new_block = match self.block_import_rx.try_recv() {
                Ok(b) => b,
                Err(TryRecvError::Empty) => {
                    debug!(target: "consensus::authority", "No new blocks from peers");
                    break;
                },
                Err(TryRecvError::Disconnected) => {
                    error!(target: "consensus::authority", "Block import channel disconnected");
                    return ();
                },
            };
```            
And then the rest of the code can just be below that instead of inside the `Ok() => {    }` block.

crates/consensus/authority/src/task.rs
```diff
@@ -105,6 +115,79 @@ where
...
+                        // send the new update to the engine, this will trigger the engine
+                        // to download and execute the block we just inserted
+                        let (tx, rx) = oneshot::channel();
+                        let _ = self.to_engine.send(BeaconEngineMessage::NewPayload {
+                            payload: block.clone().seal_slow().into(),
+                            cancun_fields: None,
+                            tx,
+                        });
+
+                        match rx.await.unwrap() {
```
stevenroose, line 138: Same thing here the error handling can be earlier and can save you 2 levels of indentation.

crates/consensus/authority/src/task.rs
stevenroose: I think it helps that while reading the `try_recv` you immediatelly see how errors are handled instead of having to scroll down and find them. Also in this care you have 3 following error handling things following each other so you'd also have to track indentation to see which belongs to which:
image

stevenroose: Did this here https://github.com/botanix-labs/Macbeth/pull/72

crates/consensus/authority/src/task.rs
stevenroose: Was this ignored on purpose?

arminsabouri: Not purposfully. Sorry I missed your comments on this PR

arminsabouri: Not purposfully. Sorry I missed your comments on this PR

arminsabouri: Yeah thats a good point. If the engine API is down or cleaning up there isnt much we can do in the Consensus layer anyways

arminsabouri: Yeah thats a good point. If the engine API is down or cleaning up there isnt much we can do in the Consensus layer anyways

PR #32, opened 2023-11-15
refactor: lower initial base fee by factor of 1000

scottmillner:


crates/primitives/src/chain/spec.rs
```diff
@@ -362,7 +361,7 @@ impl ChainSpec {
     /// Get the initial base fee of the genesis block.
     pub fn initial_base_fee(&self) -> Option<u64> {
         // If London is activated at genesis, we set the initial base fee as per EIP-1559.
-        (self.fork(Hardfork::London).active_at_block(0)).then_some(EIP1559_INITIAL_BASE_FEE)
+        (self.fork(Hardfork::London).active_at_block(0)).then_some(BOTANIX_INITIAL_BASE_FEE)
```
arminsabouri: Is there a way to conditionalize this based on the specific chain? i.e if the spec is being used for botanix chain vs some other eth chain

Thinking this will simplify upstream rebases

crates/consensus/common/src/validation.rs
```diff
@@ -311,7 +311,7 @@ pub fn validate_header_regarding_parent(
 
         let expected_base_fee =
             if chain_spec.fork(Hardfork::London).transitions_at_block(child.number) {
-                constants::EIP1559_INITIAL_BASE_FEE
+                constants::BOTANIX_INITIAL_BASE_FEE
```
arminsabouri: Same comment as below

arminsabouri (approved):

PR #36, opened 2023-12-04
Txs which fail mint contract validation should revert

scottmillner:
tested and working as expected:
- bad tx is reverted and caller is charged
- confirmed this in metamask: tx shows as failed and my address debited gas costs

### DIFF
--- a/crates/revm/src/executor.rs
+++ b/crates/revm/src/executor.rs
@@ -253,7 +253,12 @@ where
                         Ok(()) => out,
                         Err(e) => Ok({
                             error!("Botanix mint contract validation failed: {:?}", e);
-                            let new_result = ExecutionResult::Revert { gas_used: result.gas_used(), output: bytes::Bytes::new() };
+                            let output_match = match result {
+                                ExecutionResult::Success { output, .. } => output.clone().into_data().clone(),
+                                ExecutionResult::Revert { output, .. } => output.clone(),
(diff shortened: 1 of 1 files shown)


crates/revm/src/executor.rs
```diff
@@ -245,11 +246,26 @@ where
...
-            if result.is_success() && transaction.to() == Some(*MINT_CONTRACT_ADDRESS) {
-                botanix_mint_contract_checks(&result, recent_block_header)?;
+        let out = match out {
+            Ok(ResultAndState { ref result, ref state }) => {
+                if result.is_success() && transaction.to() == Some(*MINT_CONTRACT_ADDRESS) {
+                    match botanix_mint_contract_checks(&result, recent_block_header) {
+                        Ok(()) => out,
+                        Err(e) => Ok({
+                            error!("Botanix mint contract validation failed: {:?}", e);
+                            let new_result = ExecutionResult::Revert { gas_used: result.gas_used(), output: bytes::Bytes::new() };
```
arminsabouri: Do you know how output is being used?

scottmillner: sorry should've used Output from the result

fix made

and `Output` is the bytes for the contract call and is used during evm inspection to invoke the call which has already happened at this point

arminsabouri (approved):

PR #39, opened 2023-12-11
fix: revert mint contract txs that fail validation

scottmillner:
@0xBEEFCAF3 
after the upstream rebase, `revm.executor.transact` is no longer called but `revm.processor.transact` instead

so i added the same logic from `executor` to `processor` but didn't remove anything from `executor` in case reth team goes back to that implementation

### DIFF
--- a/crates/consensus/auto-seal/src/task.rs
+++ b/crates/consensus/auto-seal/src/task.rs
@@ -216,7 +216,7 @@ where
                         })
                         .unzip();
 
-                    if transactions.len() == 0 {
+                    if transactions.is_empty() {
                         return None;
                     }
                     // execute the new block
--- a/crates/revm/src/executor.rs
+++ b/crates/revm/src/executor.rs
@@ -71,10 +71,10 @@ fn botanix_mint_contract_checks(
             match pegin_data.validate(&SECP, &recent_block_header.expect("valid header")) {
                 Ok(aggregate_value) => {
                     tracing::trace!("Pegin aggregate value: {}", aggregate_value);
-                    // if aggregate_value != pegin_data.amount {
-                    //     warn!("Failed pegin attempt! Aggregate value does not match pegin amount!");
-                    //     return Err(BlockValidationError::MintContractViolation.into())
-                    // }
+                    if aggregate_value != pegin_data.amount {
+                        warn!("Failed pegin attempt! Aggregate value does not match pegin amount!");
+                        return Err(BlockValidationError::MintContractViolation.into())
+                    }
                 },
                 Err(e) => {
                     warn!("Failed pegin attempt! {:?}", e);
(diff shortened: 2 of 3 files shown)


crates/revm/src/executor.rs
```diff
@@ -71,10 +71,10 @@ fn botanix_mint_contract_checks(
             match pegin_data.validate(&SECP, &recent_block_header.expect("valid header")) {
                 Ok(aggregate_value) => {
                     tracing::trace!("Pegin aggregate value: {}", aggregate_value);
```
scottmillner, line 73: although `executor` is **not** currently being used, i uncommented the block here just in case this implementation gets used again

crates/consensus/auto-seal/src/task.rs
```diff
@@ -216,7 +216,7 @@ where
                         })
                         .unzip();
 
-                    if transactions.len() == 0 {
+                    if transactions.is_empty() {
```
arminsabouri, line 219: Just curious, whats the difference between these

crates/revm/src/processor.rs
```diff
@@ -243,6 +250,44 @@ impl<'a> EVMProcessor<'a> {
...
+                    BlockValidationError::MintContractViolation
+                })?;
+    
+                match pegin_data.validate(&SECP, &recent_block_header.expect("valid header")) {
+                    Ok(aggregate_value) => {
+                        tracing::trace!("Pegin aggregate value: {}", aggregate_value);
+                        if aggregate_value != pegin_data.amount {
+                            warn!("Failed pegin attempt! Aggregate value does not match pegin amount!");
+                            return Err(BlockValidationError::MintContractViolation.into())
+                        }
```
arminsabouri, line 271: Im going to create an issue to move this specific logic to the botanix_lib. That way we can keep all validation in the same crate

arminsabouri (approved):

crates/consensus/auto-seal/src/task.rs
stevenroose: Nothing. It's just a fancier name :)

crates/revm/src/processor.rs
```diff
@@ -243,6 +250,44 @@ impl<'a> EVMProcessor<'a> {
         Ok(())
     }
 
+    /// Performs additional checks on mint contract transactions.
+    fn botanix_mint_contract_checks(
```
stevenroose, line 254: Isn't this copied here from the code above? This changeset looks weird.

crates/revm/src/processor.rs
```diff
@@ -277,6 +323,33 @@ impl<'a> EVMProcessor<'a> {
...
+                        Ok(()) => out,
+                        Err(e) => Ok({
+                            error!("Botanix mint contract validation failed: {:?}", e);
+                            let output_match = match result {
+                                ExecutionResult::Success { output, .. } => output.clone().into_data().clone(),
+                                ExecutionResult::Revert { output, .. } => output.clone(),
+                                ExecutionResult::Halt { .. } => Bytes::new(),
+                            };
+                            let new_result = ExecutionResult::Revert { gas_used: result.gas_used(), output: output_match };
+                            ResultAndState { result: new_result, state: state.clone() }
```
stevenroose, line 340: But this is definitely a good catch!

PR #40, opened 2023-12-14
Move btc-server into macbeth as a crate

scottmillner:


bin/btc-server/Cargo.toml
```diff
@@ -0,0 +1,38 @@
+[package]
+name = "btc-server"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+reth-btc-wallet = { path = "../../crates/btc-wallet" }
```
arminsabouri: Referencing the work space would be prefered over a path here

bin/btc-server/key.hex
```diff
@@ -0,0 +1 @@
+e9e3969ceeb49c13f018acba996c79af35f920d2520bd2a36df03208936cc717
```
arminsabouri: Lets remove this file. We should only need key.template.hex and the instructions have the developer copy it over to key.hex

arminsabouri (approved):
This is awesome! Just had a couple review items

bin/btc-server/Cargo.toml
scottmillner: made change

bin/btc-server/key.hex
scottmillner: made change

PR #42, opened 2023-12-18
Scott/261 remove macbeth as submodule by creating crates

scottmillner:
What was done:
- pulled everything over from botanix (all directories)
- GH workflow: replaced reth actions with ours (btc-server and reth-node)
- updated dockerfiles for btc-server and node

I added a `compose.yml` for local docker testing (it creates btc-server and node container)
- images build but having issue making https calls (to mempool.space) in local node container
- this shouldn't be an issue for container on the box

Tested everything locally and worked as expected

Enabled Actions for the Macbeth repo so workflows should happen


arminsabouri (commented):
I noticed we deleted all the integration/unit/lint jobs? Seems useful to have those run per commit especially as we continue to add our own tests to this repo

scottmillner:
> I noticed we deleted all the integration/unit/lint jobs? Seems useful to have those run per commit especially as we continue to add our own tests to this repo

added back workflows

BatiGencho (approved):

PR #45, opened 2023-12-20
278: added proper error handling in btc server

BatiGencho:
This PR aims to introduce better error handling in the btc-server and remove unwarps

### DIFF
--- a/bin/btc-server/src/main.rs
+++ b/bin/btc-server/src/main.rs
@@ -1,6 +1,4 @@
 #[macro_use]
-extern crate anyhow;
-#[macro_use]
 extern crate log;
 
 mod database;
@@ -42,6 +40,22 @@ lazy_static::lazy_static! {
     static ref SECP: secp256k1::Secp256k1<secp256k1::All> = secp256k1::Secp256k1::new();
 }
 
+#[derive(Debug, Error)]
+pub enum Error {
+    #[error("Database error: {0}")]
+    Db(#[from] DbError),
+    #[error("Coin Selection error: {0}")]
+    CoinSelection(#[from] BdkError),
+    #[error("Pbst error: {0}")]
+    Pbst(#[from] PsbtError),
+    #[error("Invalid resulting transaction")]
+    InvaildResultingTx,
+    #[error("Failed to sign pbst")]
+    FailedToSignPbst,
+    #[error("PSBT finalization failed : {0:?}")]
+    PbstFinalizationFailed(Vec<PsbtError>),
+}
+
 struct App {
     db: database::Db,
     key: KeyPair,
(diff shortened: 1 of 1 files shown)


scottmillner (approved):
nice! 
🔥

bin/btc-server/src/main.rs
```diff
@@ -42,6 +40,22 @@ lazy_static::lazy_static! {
     static ref SECP: secp256k1::Secp256k1<secp256k1::All> = secp256k1::Secp256k1::new();
 }
 
+#[derive(Debug, Error)]
+pub enum Error {
+    #[error("Database error: {0}")]
+    Db(#[from] DbError),
+    #[error("Bdk error: {0}")]
+    Bdk(#[from] BdkError),
```
arminsabouri: Nit: Since we only use Bdk for coin selection can we rename this `CoinSelection`

arminsabouri (approved):
Nice! Just one Nit, but good to go!

bin/btc-server/src/main.rs
BatiGencho: renamed, no worries :)

bin/btc-server/src/main.rs
```diff
@@ -247,7 +261,8 @@ impl rpc::BtcServer for App {
             .require_network(self.network)
             .map_err(|e| badarg!("address for wrong network: {}", e))?;
         let output = TxOut { script_pubkey: address.script_pubkey(), value: req.value };
-        let fee_rate = FeeRate::from_sat_per_vb(req.fee_rate as u64).expect("can't overflow");
+        let fee_rate =
+            FeeRate::from_sat_per_vb(req.fee_rate as u64).ok_or(_internal!("overflowed value"))?;
```
stevenroose, line 265: This one couldn't panic. If you wanted to make sure it would not accidentally be able to panic if we changed the `req.fee_rate` to u64, f.e., you could have added a line above `let fee_rate: u64 = req.fee_rate;` and then the unwrap was safe.

bin/btc-server/src/main.rs
```diff
@@ -316,7 +331,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         tx_lock: Arc::new(Mutex::new(())),
     };
 
-    let addr = "0.0.0.0:8080".parse().unwrap();
+    let addr = "0.0.0.0:8080".parse().context("Unparsable address")?;
```
stevenroose, line 334: `0.0.0.0:8080` is a valid address, why would you introduce error code paths here?

bin/btc-server/src/main.rs
```diff
@@ -115,11 +129,12 @@ impl App {
...
                 self.change_script.as_script(), // drain_script
             )
-            .context("coin selection")?;
+            .map_err(Error::CoinSelection)?;
         let selected = selection
             .selected
             .iter()
-            .map(|u| utxos.get(&OutPoint::from_bdk(u.outpoint())).unwrap())
+            .map(|u| utxos.get(&OutPoint::from_bdk(u.outpoint())))
+            .filter_map(|s| if s.is_some() { s } else { None })
```
stevenroose, line 137: The like is exactly equivalent to `.filter_map(|s| s)`. Also, you could have changed the line above as such to have the exact same effect:
```rust
            .filter_map(|u| utxos.get(&OutPoint::from_bdk(u.outpoint())).ok())
```

That being said, the code above is broken if that `utxos.get` ever returns `None` (we put the utxos in the map ourselves and then get some of them back out), so we would really want this code to panic in that case, it means either out code or the bdk dependency is broken and we better know this during testing than that we just ignore it.

So maybe a change to an `.expect("we put it in the map")` or something would be more appropriate.

stevenroose:
I know it's after the fact, just giving some feedback. Sorry I'm behind :)

PR #47, opened 2023-12-20
build: remove ci token secret from workflow

scottmillner:


.github/workflows/deploy-reth-node.yml
```diff
@@ -12,9 +12,6 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v2
-        with:
-          submodules: 'true'
-          token: ${{ secrets.CI_TOKEN }}
```
arminsabouri, line 17: Note: this was only needed for pulling the submodule. Safe to delete.

arminsabouri (approved):

PR #50, opened 2023-12-26
Arm/enforce block height greater

arminsabouri:
Ported over from https://github.com/botanix-labs/botanix/pull/260

Open TODOs

- [x] rebase base branch
- [x] update genesis contract byte code


arminsabouri:
@scottmillner i fixed some broken tests in test/faucet.sol
Ready to go

arminsabouri:
Whoa. Does this mean Im a solidity developer now
Screenshot 2024-01-05 at 1 57 01 PM

scottmillner:
> @scottmillner i fixed some broken tests in test/faucet.sol Ready to go

ah nice, i forgot to update those, thanks!!

scottmillner:
> Whoa. Does this mean Im a solidity developer now Screenshot 2024-01-05 at 1 57 01 PM

do you feel gross? 🤣 

and we all know the dark side is waaaaay cooler!! welcome Darth Arminis!

scottmillner (approved):
thanks for doing this! 🔥 🔥

stevenroose:
What was the motivation for this change? I think I recall there was a good reason this was allowed.

See https://github.com/botanix-labs/botanix/issues/240#issuecomment-1888656883.

scottmillner:
> What was the motivation for this change? I think I recall there was a good reason this was allowed.
> 
> See [botanix-labs/botanix#240 (comment)](https://github.com/botanix-labs/botanix/issues/240#issuecomment-1888656883).

What we're trying to allow is multiple pegins in the same block (sending a list of proofs to the node) but prevent the replay of a single pegin.  We noticed when testing we could just replay the same pegin over and over until the block height increases.  This does assume users are going through our dApp (hence using sidecar service). So sidecar should package multiple pegins in the same block for the same address in the payload to the node but an attacker wouldn't be able to replay any of those pegins bc block height needs to increase. We updated the node parse/validation logic to expect the pegin metadata to be a list.

stevenroose:
Yeah you're right. I forgot. When 2 pegins confirm in the same BTC block, you can claim them in the same BTX tx together and then the height can be required to always increase. I forgot :)

PR #55, opened 2024-01-02
btc network config taken from cli flag

BatiGencho:
This PR aims at providing the user with the ability to tell which btc network the reth cli is to use as the latter is needed for pegins and pegouts and it used to be hardcoded in the code. The cli arg is completely optional and if not used, the signet shall be used as the default one.

### DIFF
--- a/bin/reth/src/args/rpc_server_args.rs
+++ b/bin/reth/src/args/rpc_server_args.rs
@@ -194,6 +194,10 @@ pub struct RpcServerArgs {
     /// The metrics will be served at the given interface and port.
     #[arg(long, value_name = "BITCOIN_BLOCK_SOURCE", value_parser = parse_url, help_heading = "Btc_block_source")]
     pub btc_block_source: Url,
+
+    /// The bitcoin network to operate on.
+    #[arg(long, value_name = "BITCOIN_NETWORK", help_heading = "Btc_network", required = true)]
+    pub btc_network: bitcoin::Network,
 }
 
 impl RpcServerArgs {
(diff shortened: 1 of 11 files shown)


bin/reth/src/args/rpc_server_args.rs
```diff
@@ -194,6 +194,10 @@ pub struct RpcServerArgs {
     /// The metrics will be served at the given interface and port.
     #[arg(long, value_name = "BITCOIN_BLOCK_SOURCE", value_parser = parse_url, help_heading = "Btc_block_source")]
     pub btc_block_source: Url,
+
+    /// The bitcoin network to operate on.
+    #[arg(long, value_name = "BITCOIN_NETWORK", help_heading = "Btc_network", required = false)]
+    pub btc_network: Option<bitcoin::Network>,
```
arminsabouri: this should be a non-optional field. A default that uses signet would be a bit confusing for a node-runner.

arminsabouri (commented):
Awesome work! I just had one comment

bin/reth/src/args/rpc_server_args.rs
BatiGencho: yep, makes sense. Made it non-optional. Thx

crates/consensus/auto-seal/src/task.rs
```diff
@@ -255,7 +260,7 @@ where
                                                 {
                                                     Ok(GenesisContractEvents::MintingEvent) => {
                                                         info!("Parsing and sending minting event to btc_server");
-                                                        let pegin_data = parse_pegin_reth_log_topic(&log).expect(
```
scottmillner: do we have plans to use `btc_network` for pegins?
only pegouts have a network field so do we need to pass it here?

BatiGencho: I believe not, so I have removed it for the time being. If needed, we could always reintroduce it.

arminsabouri (approved):

scottmillner (approved):

PR #57, opened 2024-01-02
use timestamp to determine if in turn

scottmillner:


BatiGencho (approved):

crates/consensus/authority/src/epoch_manager.rs
```diff
@@ -48,8 +48,16 @@ impl EpochManager {
         let random_delay = self.random_delay.take();
         let storage = self.storage.inner.read().await;
         let signer_index = storage.signer_index;
+        let timestamp = SystemTime::now().duration_since(UNIX_EPOCH).unwrap_or_default().as_secs();
+        // sanity check timestamp is greater than previous best block timestamp
+        let prev_best_block_timestamp = storage.headers.get(&storage.best_block).expect("checked empty").timestamp;
+        if timestamp <= prev_best_block_timestamp {
```
arminsabouri: This is a good sanity check. This really means that our system clock is behind.
Maybe out of scope for this PR but there should be some async task that regularly syncs with an NTP server.

https://en.wikipedia.org/wiki/Network_Time_Protocol

crates/consensus/authority/src/epoch_manager.rs
```diff
@@ -48,8 +48,16 @@ impl EpochManager {
...
+            error!("Timestamp is not greater than previous best block timestamp");
+            return Poll::Pending
+        }
+
         let is_inturn =
-            storage.best_block % (storage.authorities.len() as u64) == signer_index as u64;
+            timestamp % (storage.authorities.len() as u64) == signer_index as u64;
```
arminsabouri: Is the time denominated in seconds? If so, it seems like the allocation isn't completely round-robin. Consider this: with 5 federation members, at t = 1704291331, t % N = 1. Later, after processing blocks, let's say 3 seconds have passed (best case scenario, unlikely due to pre-commitment). At t = 1704291334, t % N = 4. Then, after another second, at t = 1704291336, t % N = 1 again.

If this pattern persists, certain signers might face block production starvation. An alternative approach could involve using minutes and adjusting the timer in the epoch_manager to trigger every minute, allowing for a check to determine if it's their turn.

arminsabouri (changes requested):
This is awesome 🔥 
I had comment that needs some change. 
Additionally we should also update the consensus validation to check the correct signer was in_turn.

I know we said this ticket https://github.com/botanix-labs/botanix/issues/289
But could make sense to include this as well.

Since this is super involved, happy to hop on call to white board some ideas out

crates/consensus/authority/src/epoch_manager.rs
stevenroose: It can also mean that the previous block maker put on a timestamp too far into the future. I don't know how Ethereum regularly checks that. In Bitcoin it's kinda lax because it's not so critical, but if you're going to use the timestamps more critically, then node validators should validate block timestamps as well.

It's a bit non-trivial though. FWIW in Liquid there is no time check on consensus, the time is only used for leader picking in the block making protocol which is totally independent from consensus.

crates/consensus/authority/src/epoch_manager.rs
stevenroose: We discussed this in our call last week, you should first divide the time by N and then to mod N. Otherwise it's indeed never round robin. Imagine you have a 10-second round robin and also 10 block makers, the same one would always be picked.

stevenroose: Using block height to do the round robin seems a lot more robust, tbh

bin/reth/src/poa/mod.rs
```diff
@@ -236,6 +261,31 @@ impl<Ext: RethCliExt> PoaNodeCommand<Ext> {
...
         raise_fd_limit();
 
+        // async task that checks system clock is in sync with NTP server
+        ctx.task_executor.spawn_critical(
+            "async system clock sync with ntp task",
+            Box::pin(async {
+                let sleep_sec = tokio::time::Duration::from_secs(15);
+                let acceptable_drift_sec = 1;
+                loop {
+                    // TODO (scott) pass in ntp url as arg
```
arminsabouri, line 271: No changes now but Once this comes an arg we should use a strong type like Address or Url

bin/reth/src/poa/mod.rs
```diff
@@ -236,6 +261,31 @@ impl<Ext: RethCliExt> PoaNodeCommand<Ext> {
...
+                    match Self::ntp_unix_timestamp("time.cloudflare.com").await {
+                        Ok(ntp_timestamp) => {
+                            let system_timestamp = utils::unix_timestamp();
+                            if (ntp_timestamp as i64 - system_timestamp as i64).abs() > acceptable_drift_sec {
+                                error!("System clock is not in sync with NTP server. System timestamp: {}, NTP timestamp: {}", system_timestamp, ntp_timestamp);
+                            }
+                            info!("System clock is in sync with NTP server. System timestamp: {}, NTP timestamp: {}", system_timestamp, ntp_timestamp);
```
arminsabouri: Should this log be wrapped in an else?

crates/consensus/authority/src/lib.rs
```diff
@@ -69,6 +69,24 @@ impl AuthorityConsensus {
...
+    /// Returns true if the authority is in turn
+    pub fn is_inturn(authorities_len: u64, signer_index: u64) -> bool {
+        // use minutes as time unit to determine in turn
+        let timestamp = utils::unix_timestamp() / 60;
+
+        (timestamp / authorities_len) % authorities_len == signer_index
+    }
+
+    /// Validates that the authority was in turn when producing the block
+    pub fn validate_inturn(block_timestamp: u64, authorities_len: u64, signer_index: u64) -> Result<(), ConsensusError> {
```
arminsabouri, line 82: Is the timestamp on the eth blocks denominated in minutes or seconds? Do we convert min -> sec somewhere

crates/consensus/authority/src/task.rs
```diff
@@ -136,6 +137,26 @@ where
...
                         let block = new_block.block.block.clone();
                         info!(target: "consensus::authority", ?block, "Recieved new block from peer");
+
+                        // extract signer pub key
+                        let signer = utils::recovery_authority(&block.header).expect("valid signer");
+                    
+                        let storage_inner = self.epoch_manager.storage.inner.read().await;
+                        let authorities = storage_inner.authorities.clone();
+                        let signer_index = authorities.iter().position(|pk| *pk == signer).expect("valid signer");
+                        drop(storage_inner);
```
arminsabouri, line 146: Could move this drop up a line. ideally we release the lock as soon as we are done using it

arminsabouri (commented):
Just some nit comments and questions. Over all this looks 🔥

bin/reth/src/poa/mod.rs
scottmillner: yep! nice catch, thanks!

done

crates/consensus/authority/src/lib.rs
scottmillner: ah i forget to convert:
the block_timestamp is in seconds and need to convert to minutes

done

crates/consensus/authority/src/task.rs
scottmillner: done

crates/consensus/authority/src/task.rs
```diff
@@ -136,6 +137,26 @@ where
...
+                        let signer_index = authorities.iter().position(|pk| *pk == signer).expect("valid signer");
+                        match AuthorityConsensus::validate_inturn(
+                            block.header.timestamp,
+                            authorities.len() as u64,
+                            signer_index as u64,
+                        ) {
+                            Ok(_) => {}
+                            Err(err) => {
+                                error!(target: "consensus::authority", ?err, "Block import failed in turn check");
+                                return
```
BatiGencho: Should not we be doing a `continue` here in order to move up in the loop {} where we are blocked waiting on a channel to receive ?

scottmillner: yep! nice catch

done

arminsabouri (approved):
tACK [8dfa92f](https://github.com/botanix-labs/Macbeth/pull/57/commits/8dfa92f3dc5d08520a4d912b61a92bf5d304bc24)

Tested in_turn logic ✅ 
Validating incoming blocks is failing but we know the fix for that is on #63

PR #58, opened 2024-01-03
botanix_mint_contract_checks and transact unit tests for pegin/pegout

BatiGencho:
Ths PR adds multiple unit tests for the `botanix_mint_contract_checks` function. It also adds a serialization method for the PeginMeta and exposes some test utils from another crate. The method `transact`  has not been tested separately as it basically wraps the `botanix_mint_contract_checks` and adds some optimism checks on top of it which are hard to test in isolation. It is also a private method.


crates/revm/src/processor.rs
```diff
@@ -1076,4 +1082,245 @@ mod tests {
...
+        let data = Bytes::from(encoded_params);
+
+        // create a success result object
+        let result = ExecutionResult::Success {
+            reason: Eval::Return,
+            gas_used: 0,
+            gas_refunded: 0,
+            logs: vec![Log {
+                address: Address::from(*MINT_CONTRACT_ADDRESS),
+                topics: vec![*MINT_TOPIC, *MINT_TOPIC],
```
arminsabouri: Can you remind me why we need `MINT_TOPIC` here twice?

crates/revm/src/processor.rs
```diff
@@ -1076,4 +1082,245 @@ mod tests {
...
+                data,
+            }],
+            output: Output::Call(Bytes::new()),
+        };
+
+        // run checks on the output
+        let res = EVMProcessor::botanix_mint_contract_checks(&result, Some((*header, 1_u32)));
```
arminsabouri: Should we have some kind of is_ok or expect() check here?

arminsabouri (commented):
Great stuff. Just had a couple questions

crates/revm/src/processor.rs
BatiGencho: That is the example I need some more clarifications from @scottmillner (this is not fully finished)

crates/revm/src/processor.rs
BatiGencho: There is a check that always checks that the logs.lenght is = = 2. Any idea why ?

BatiGencho: We discussed that already. The second topic is the indexed address. I have added that to the tests now.

crates/revm/src/processor.rs
BatiGencho: The serialization here was missing, I added it. the test is now using is_ok()

scottmillner (approved):
awesome work!! 🔥

arminsabouri (approved):

PR #59, opened 2024-01-04
Arm/poa cleanup

arminsabouri:
Cleaning up double code in task. Mainly around inserting and persisting new blocks and calling fork choice update.
There are tons of warnings after the upstream rebase. I tried to squash as many as I can here related to consensus. I'll grab the other ones in a future commit. Hopefully soon the lint job on gh will pass 💯 

Note: I separated out the refactor commits from the warning fix commits for ease of review

### DIFF
--- a/crates/consensus/authority/src/task.rs
+++ b/crates/consensus/authority/src/task.rs
@@ -8,7 +8,7 @@ use reth_eth_wire::NewBlock;
 use reth_interfaces::consensus::ForkchoiceState;
 use reth_network::{message::NewBlockMessage, NetworkHandle};
 use reth_primitives::{
-    hex, Block, BlockBody, ChainSpec, IntoRecoveredTransaction, SealedBlockWithSenders,
+    hex, Block, BlockBody, ChainSpec, IntoRecoveredTransaction, Log, SealedBlockWithSenders,
     TransactionSigned, U256,
 };
 use reth_provider::{CanonChainTracker, CanonStateNotificationSender, Chain, StateProviderFactory};
@@ -32,6 +32,18 @@ use tracing::{debug, error, info, warn};
 use url::Url;
 
 use client::{BtcServerClient, MakeTxRequest, NotifyPeginRequest};
+
+/// Repersents an error while processing a botanix log
+#[derive(Debug, thiserror::Error)]
+enum ProcessBotanixLogError {
+    /// Failed to notify btc server about pegin
+    #[error("Failed to notify btc server about pegin")]
+    FailedToNotifyPegin(tonic::Status),
+    #[error("Failed to broadcast pegout tx")]
+    FailedToBroadcastPegout,
+    #[error("Failed to make pegout tx")]
+    FailedToMakePegoutTx(tonic::Status),
+}
 
 pub struct BlockProductionTask<Client, Pool: TransactionPool> {
     /// The configured chain spec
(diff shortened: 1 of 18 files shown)


crates/consensus/common/src/utils.rs
```diff
@@ -46,15 +49,21 @@ pub fn read_staker_balance(
 }
 
 #[derive(Debug)]
+/// Error that can occur while recovering the authority list
```
BatiGencho, line 51: Dont we want to use thiserror here or displaydoc with macros for better error message handling, esp. with nested errors ?

crates/consensus/authority/src/task.rs
```diff
@@ -102,7 +114,9 @@ where
             pipe_line_events: None,
             btc_server,
             bitcoin_block_header,
-            bitcoin_block_source_address,
+            bitcoin_block_source: MempoolSpace::new(
+                bitcoin_block_source_address.clone().to_string(),
```
BatiGencho: do we need to clone it here, the to_string() should construct and return a new owned String right ?

crates/consensus/authority/src/task.rs
```diff
@@ -171,92 +182,14 @@ where
                                             // TODO (armins) remove expect
                                             .expect("block is valid");
```
BatiGencho: do we want to handle such expects ?

crates/consensus/common/src/utils.rs
```diff
@@ -76,12 +85,15 @@ pub fn recovery_authority(header: &Header) -> Result<secp256k1::PublicKey, Recov
 }
 
 #[derive(Debug)]
+/// Errors that can occur while reading the authority list from the block header
 pub enum GetAuthoritiesError {
+    /// Failed to deserialize the extra data
```
BatiGencho, line 89: same comment here reg. the error handling - displaydoc / thiserror ?

crates/consensus/common/src/utils.rs
arminsabouri: You have more Rust experience I'll follow your recommendation. What do you think is the best way to move forward?

crates/consensus/authority/src/task.rs
arminsabouri: Good point! Added error handling to this

crates/consensus/authority/src/task.rs
arminsabouri: Good catch! Fixed [22ba968](https://github.com/botanix-labs/Macbeth/pull/59/commits/22ba968aa10db41d2a45868c8f5b57bb02569fd2)

scottmillner (approved):

crates/consensus/common/src/utils.rs
arminsabouri: @BatiGencho Im going to merge the clean up. But want to make sure this is addresses. Lmk what the best path going forward is and I can make a change in a future PR

BatiGencho: I went through it, seems clean to me. Good that you merged it!

crates/consensus/authority/src/task.rs
BatiGencho: Great!

crates/consensus/authority/src/task.rs
BatiGencho: Great!

PR #63, opened 2024-01-07
296 refactor list of authority signers should be optional

arminsabouri:
Resolves: https://github.com/botanix-labs/botanix/issues/296
and #297
Also, refactored optional fields to use a u8 bitmask. Alternatively, we could RLP encode/decode the whole thing but that may be space inefficient. And headers do need to be as lean as possible. 

lmk your thoughts on this.


arminsabouri:
Drafting for now
Going to change poa to only add authorities during epoch blocks

arminsabouri:
Good to go!

crates/botanix-lib/src/extra_data_header.rs
```diff
@@ -205,12 +197,14 @@ impl ExtraDataHeader {
         };
 
         let signature = if optional_fields & (1u8 << HAS_SIGNATURE_POS) != 0 {
+            let recovery_id = RecoveryId::from_i32(i32::consensus_decode(reader)?).unwrap();
```
arminsabouri, line 200: remove unwrap!

scottmillner (approved):
🔥 🔥 🔥

PR #66, opened 2024-01-09
298 fix remove uses of h160 and h256 in favor of b265 h265

arminsabouri:
Another cleanup pr. Reducing warnings and removing usued structures from eip255

### DIFF
--- a/crates/consensus/common/src/validation.rs
+++ b/crates/consensus/common/src/validation.rs
@@ -2,7 +2,6 @@
 use reth_interfaces::{consensus::ConsensusError, RethResult};
 use reth_primitives::{
     constants::{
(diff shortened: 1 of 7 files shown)


crates/consensus/common/src/validation.rs
```diff
@@ -547,13 +546,6 @@ pub fn validate_header_with_total_difficulty(
         return Err(ConsensusError::TheMergeOmmerRootIsNotEmpty)
     }
```
scottmillner: thanks for removing this...i meant to do it in my pr but missed it ☹️

scottmillner (approved):
thanks for removing the inturn/outofturn stuff...forgot to do it in my pr

PR #68, opened 2024-01-11
feature: nodes should propagate txs in mempool

BatiGencho:
This PR handles https://github.com/botanix-labs/botanix/issues/291.
The gossiping protocol indeed works and nodes do share txs when they arrive in the mempool. The blocking nature of the epoch manager was not allowing the fed nodes to share txs as soon as they arrive as the one out of turn never really drained the mempool or even got to it. I have reworked this a bit and also added the possibility for peers to prune their mempool after new blocks are confirmed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -264,3 +264,41 @@ start-reth-server:
 	--auto-mine \
 	--btc-server localhost:8080 \
 	--btc-block-source "https://mempool.space/signet/api"
+
+start-poa-server-1:
+	cd ./bin/reth && \
+	cargo run --bin reth -- poa \
+	--chain botanix_testnet \
+	--datadir ${NODE_1_DIR} \
+	--http \
+	--http.corsdomain "*" \
+	--http.port 8545 \
+	--http.addr "127.0.0.1" \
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -263,4 +263,42 @@ start-reth-server:
...
 	--btc-server localhost:8080 \
-	--btc-block-source "https://mempool.space/signet/api"
\ No newline at end of file
+	--btc-block-source "https://mempool.space/signet/api"
+
+start-poa-server-1:
+	cd ./bin/reth && \
+	cargo run --bin reth -- poa \
+	--chain botanix_testnet \
+	--datadir "/home/evgeni/Documents/BOTANIX/node1" \
```
arminsabouri: Maybe we can use `DB_DIR_NODE1` ?

arminsabouri (approved):

Makefile
BatiGencho: Yep, sure, I renamed them to use NODE_1_DIR and NODE_2_DIR. I presume jwt, db and discovery could all be placed for convenience in one folder

scottmillner (approved):
🔥

PR #69, opened 2024-01-11
feature: base fee burn should go to block producer

BatiGencho:
This PR makes it possible for the base fee not to be burned as it currently happens under Eth 1559, but to rather be added to the prio fee so we can avoid lost BTC on Botanix.

### DIFF
--- a/crates/primitives/src/transaction/mod.rs
+++ b/crates/primitives/src/transaction/mod.rs
@@ -374,8 +374,9 @@ impl Transaction {
         let fee = max_fee_per_gas - base_fee;
 
         // Compare the fee with max_priority_fee_per_gas (or gas price for non-EIP1559 transactions)
+        // NOTE: Add the priority_fee to the base fee as we have to avoid lost BTC on Botanix
         if let Some(priority_fee) = self.max_priority_fee_per_gas() {
-            Some(fee.min(priority_fee))
+            Some(fee.min(priority_fee) + base_fee)
         } else {
             Some(fee)
         }


arminsabouri (approved):

crates/primitives/src/transaction/mod.rs
```diff
@@ -374,8 +374,9 @@ impl Transaction {
         let fee = max_fee_per_gas - base_fee;
 
         // Compare the fee with max_priority_fee_per_gas (or gas price for non-EIP1559 transactions)
+        // NOTE: Add the priority_fee to the base fee as we have to avoid lost BTC on Botanix
         if let Some(priority_fee) = self.max_priority_fee_per_gas() {
-            Some(fee.min(priority_fee))
```
scottmillner, line 378: shouldn't this be `Some(fee.min(priority_fee) + base_fee)`? 

 so `gasTipCap` + `baseFee`
 
 bc currently the changes are the same as `max_fee_per_gas - base_fee + priority_fee`

BatiGencho: yeah, you are correct. MIssed that, thx

BatiGencho: fixed

scottmillner (approved):

PhoenixCodeCraft:
Is there a way to test this? Maybe create a test to check total BTC on spiderchain = total BTC on EVM?

stevenroose:
Is this function actually what is used in consensus to determine the fees? It's a method on the primitives.

Also, the documentation of the function didn't change.

I have trouble understanding what is going on in this function.

---

```rust
        // update add to total fees
        let miner_fee = tx
            .effective_tip_per_gas(Some(base_fee))
            .expect("fee is always valid; execution succeeded");
        total_fees += U256::from(miner_fee) * U256::from(gas_used);
```
Ok so `fee per gas` and `base_fee` are actually fee rates and not fees.

In that case it might be sane to add the base fee with the tip together in this method.

I'm wondering though, the `fee.min(max_priority_fee_per_gas)` still somehow caps the feerate. What if the feerate is higher than the max, there is still bitcoin being lost.

And does this thresholding in no way accidentally assign too much fee? Like can the feerate eat into the base fee somehow?

I was looking for the code where the fee is actually taken from the sender's account, but it seems to be burried deeply inside the EVM VM codebase. It would be really valuable to have unit tests that do some txs and assert that the fee we thing was spent is actually the same amount taken from the sender's account. I doubt if it'd be easy to write such a test.

Alternatively, inside the code we could fetch the address' balance before and after execution and assert that the diff is equal to our calculation of the fee. If this is guarded by `cfg!(debug_assertions)`, this would only be calculated while running tests.

BatiGencho:
> Is this function actually what is used in consensus to determine the fees? It's a method on the primitives.
> 
> Also, the documentation of the function didn't change.
> 
> I have trouble understanding what is going on in this function.
> 
> ```rust
>         // update add to total fees
>         let miner_fee = tx
>             .effective_tip_per_gas(Some(base_fee))
>             .expect("fee is always valid; execution succeeded");
>         total_fees += U256::from(miner_fee) * U256::from(gas_used);
> ```
> 
> Ok so `fee per gas` and `base_fee` are actually fee rates and not fees.
> 
> In that case it might be sane to add the base fee with the tip together in this method.
> 
> I'm wondering though, the `fee.min(max_priority_fee_per_gas)` still somehow caps the feerate. What if the feerate is higher than the max, there is still bitcoin being lost.
> 
> And does this thresholding in no way accidentally assign too much fee? Like can the feerate eat into the base fee somehow?
> 
> I was looking for the code where the fee is actually taken from the sender's account, but it seems to be burried deeply inside the EVM VM codebase. It would be really valuable to have unit tests that do some txs and assert that the fee we thing was spent is actually the same amount taken from the sender's account. I doubt if it'd be easy to write such a test.
> 
> Alternatively, inside the code we could fetch the address' balance before and after execution and assert that the diff is equal to our calculation of the fee. If this is guarded by `cfg!(debug_assertions)`, this would only be calculated while running tests.

It is indeed hard, but I am going to build in a few tests in  a separate MR to prove this point. In a separate MR I have implemented the feature of any fees being sent out to the producing block beneficiary, so I it would be easy to test out the fee allocation. I will add you as a reviewer for sure.

stevenroose:
Cool. Sorry for the delayed response. But yeah this is pretty critical code and it being so vague is not great.

PR #70, opened 2024-01-11
300 feature ensure in turn is only producing one block per time slot

arminsabouri:
This PR achieves a couple of different things:
* empose signer limit for in turn fed member
* Removed duplicate block assertions -> causing recover authority to fail
* Adds auth pk to storage inner on new()
* And removes some unwrap() from persist_new_block()

The authority task is getting a bit hairy. Mainly b/c this task does 2 main things
1. Handle block import
2. Handle create new block

And within these tasks there are duplicate code and functions that should in an async util. Such as calling the engine API
 I think its fine to have the two things in the same tasks but this code could use refactor and cleaning up

### DIFF
--- a/crates/consensus/authority/src/epoch_manager.rs
+++ b/crates/consensus/authority/src/epoch_manager.rs
@@ -47,30 +47,25 @@ impl EpochManager {
     {
         let storage = self.storage.inner.read().await;
         let signer_index = storage.signer_index;
-        println!("signer_index: {}", signer_index);
-
+        let signer_pk = storage.authority;
         let authority_len = storage.authorities.len() as u64;
-        let signer_index = storage.signer_index as u64;
 
         // Check if the last signer was us
         // If so nothing to do anymore until the next timeslot
         let latest_header = storage.headers.get(&storage.best_block).expect("best block");
+        // Skip over genesis
         if latest_header.number != 0 {
             let latest_signer = utils::recovery_authority(&latest_header).unwrap();
-            let latest_signer_index =
-                storage.authorities.iter().position(|pk| pk == &latest_signer).unwrap() as u64;
-            println!("signer_index: {},  latest_signer_index: {}", signer_index, latest_signer_index);
-            if latest_signer_index == signer_index {
-                debug!(
-                    "latest signer index {} != signer index {}",
-                    latest_signer_index, signer_index
-                );
+            let current_ts = utils::unix_timestamp();
+            if let Err(_) = utils::validate_current_signer_against_last(
+                (latest_signer, latest_header.timestamp / 60),
+                (signer_pk, current_ts / 60),
+            ) {
                 return Poll::Pending
             }
         }
 
         drop(storage);
-
         let is_inturn = AuthorityConsensus::is_inturn(authority_len, signer_index);
         info!("is_inturn: {}", is_inturn);
 
--- a/crates/consensus/authority/src/epoch_manager.rs
+++ b/crates/consensus/authority/src/epoch_manager.rs
@@ -13,8 +13,6 @@ use std::{sync::Arc, task::Poll, time::Duration};
 
 #[derive(Debug)]
 /// Manages the block production epochs
-/// Signing time is the parent timestamp + BLOCK_PERIOD
-/// If the signer is inturn then we broadcast the block
 ///
 /// Blocks will be rejected by consensus if
 /// 1. The signer is not in the federation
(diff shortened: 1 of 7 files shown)


crates/consensus/authority/src/lib.rs
```diff
@@ -152,7 +155,11 @@ impl Storage {
             best_block: header.number,
             authorities,
             signer_index,
-            ..Default::default()
+            authority: pk,
+            headers: HashMap::new(),
```
BatiGencho, line 159: cannot we just keep the authority: pk, and use Default::default() for the rest as they seem to be stil defaulted.

BatiGencho:
@0xBEEFCAF3 looks like after merging my MRs there are some conflicts, do you mind resolving them pls ? Thank you.

scottmillner (approved):

arminsabouri:
> @0xBEEFCAF3 looks like after merging my MRs there are some conflicts, do you mind resolving them pls ? Thank you.

Done ✅

crates/consensus/authority/src/lib.rs
arminsabouri: Yeah good point

arminsabouri: Yeah good point

arminsabouri: Actually, I remembered why I needed to do this
secp::public key doesn't impl default. So then I cant use `..Default.default`

Is there a good way around this?

BatiGencho: I meant keep the authority: pk as it is, and make the rest default via Default::default() - they seem to be all defaultable, right ?

crates/consensus/common/src/utils.rs
```diff
@@ -152,6 +152,59 @@ pub fn validate_poa_extra_data_header(
...
+    if parent.number == 0 {
+        return Ok(())
+    }
+    let parent_signer = recovery_authority(&parent)
+        .map_err(|e| ValidateAgainstParrentError::FailedToDerserializeExtraData(e))?;
+    let current_signer = recovery_authority(&current)
+        .map_err(|e| ValidateAgainstParrentError::FailedToDerserializeExtraData(e))?;
+    let is_same_signer = parent_signer == current_signer;
+    // Check if the parent block was mined in a different turn
+    let parent_ts = parent.timestamp / 60;
```
stevenroose, line 205: Should probably take some kind of chain config and take these values from there (also "assuming 1 minute block times"). Also the math below seems to be quite handwavy 
```
    // Last block should be greater that 1 minute in the worst cast
    // Even in the case of > 2 federation members the worst case time between blocks for the same
    // Signer should be 1 minute. Assuming 1 minute block times
```

crates/consensus/authority/src/epoch_manager.rs
```diff
@@ -53,6 +51,24 @@ impl EpochManager {
 
         let authority_len = storage.authorities.len() as u64;
         let signer_index = storage.signer_index as u64;
```
stevenroose: This seems to be fixed while your code below seems to be "within this epoch". Does this function only get called on newly mined tips or also on all historical blocks?

crates/consensus/authority/src/task.rs
```diff
@@ -500,10 +500,6 @@ where
             ommers: vec![],
             withdrawals: None,
         };
-        let mut storage = self.storage.write().await;
-        storage.insert_new_block(sealed_block.header.header.clone(), body.clone());
-        drop(storage);
```
stevenroose: Where is this done otherwise? I find `insert_new_block` in lib.rs, but I can't see how these two callsites relate.

stevenroose (commented):
Left a few questions

crates/consensus/authority/src/lib.rs
arminsabouri: They are. This is probably me being a rust noob but doesn't the entire struct need to be defaultable to use Default::default()

crates/consensus/common/src/utils.rs
arminsabouri: I agree. Not chain config b/c most chains won't have this config and the rebases will be a mess. BUt this should be in the PoA consts file

arminsabouri: What I meant by the fn desc is that **at most**  one `BLOCK_TIME` amount of time must have passed before a fed member can produce another block. Granted they might not be `in_turn` but consensus will validate that.

crates/consensus/authority/src/task.rs
arminsabouri: yup in lib.rs
Its important to note that `insert_new_block` is just updating the consensus layer cache NOT the database. 
It should be the responsibility of build and execute to update the cache and authority task should just be concerned with producing what it thinks is valid blocks.

So it's a seperation of concerns

crates/consensus/authority/src/epoch_manager.rs
arminsabouri: `poll` only gets called to mine new blocks when the signer is `in_turn`

crates/consensus/authority/src/lib.rs
BatiGencho: yep, i believe I did not realize that you did not have the Default trait implemented on it. all good then, sorry

BatiGencho (approved):

PR #73, opened 2024-01-12
304 feature utility to communicate to engine api

arminsabouri:
@scottmillner Drafting until I refactor tasks to use these.
Wanted to put up these commits for visibility on the interface

### DIFF
--- a/crates/consensus/authority/src/engine_util.rs
+++ b/crates/consensus/authority/src/engine_util.rs
@@ -3,28 +3,30 @@
 // engine_newPayloadV2 Interacting with the engine API can be done via RPC or the engine API shared
 // queue
 
-use reth_beacon_consensus::BeaconEngineMessage;
+use reth_beacon_consensus::{BeaconEngineMessage, BeaconOnNewPayloadError, ForkchoiceStatus};
 use reth_primitives::{SealedBlock, SealedHeader};
-use reth_rpc_types::engine::{PayloadStatus, PayloadStatusEnum};
-use tokio::sync::mpsc::UnboundedSender;
+use reth_rpc_types::engine::{ForkchoiceState, PayloadStatusEnum};
+use tokio::sync::{mpsc::UnboundedSender, oneshot};
 
-use tracing::debug;
+use tracing::{debug, error};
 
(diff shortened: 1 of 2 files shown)


crates/consensus/authority/src/engine_util.rs
```diff
@@ -0,0 +1,177 @@
...
+    let state = ForkchoiceState {
+        head_block_hash: new_header.hash,
+        finalized_block_hash: new_header.hash,
+        safe_block_hash: new_header.hash,
+    };
+    let mut ctr = 0;
+    loop {
+        if ctr == 5000 {
+            return Err(SendForkChoiceUpdateError::Timeout)
+        }
```
arminsabouri, line 95: Note for reviewers: there are two things we can do when the engine API isnt responding

If they are offline (the channel cannot be open) -> Should we panic?
What happens if they are syncing for a long time -> panic? wait? timeout?

scottmillner: i think we should panic if the channel can't be opened bc that seems critical and no point continuing until that is resolved

syncing for a long time: this could be for different reasons (unstable peers to download headers from, connectivity issues) so maybe timeout and retry then panic over retry threshold

arminsabouri:
Removed the last commit which changes the authority task. There are currently too many PRs that change that file. I will update once #76 is merged

scottmillner (approved):

crates/consensus/authority/src/engine_util.rs
```diff
@@ -0,0 +1,177 @@
...
+                    }
+                }
+            }
+            Err(err) => {
+                error!(target: "consensus::authority", ?err, "Authority fork choice update failed");
+                return Err(SendForkChoiceUpdateError::InvalidPayload)
+            }
+        }
+        ctr += 1;
+        std::thread::sleep(std::time::Duration::from_millis(500));
```
stevenroose, line 126: 500ms is a _really_ long time in CPU world, though.. This could easily be 50.

Also, you can't use std::thread::sleep in async functions, that will actually hold up the runtime thread. Should use `tokio::time::sleep().await`.

OK, I see these two things are already addressed ;) good job :)

PR #75, opened 2024-01-15
feature: Ensure EVM TX fees are being deposited to federation members account

BatiGencho:
This PR adds poa block beneficiary inclusion, checks and testing.

Ticket description here: https://github.com/botanix-labs/botanix/issues/302

### DIFF
--- a/crates/consensus/authority/src/lib.rs
+++ b/crates/consensus/authority/src/lib.rs
@@ -28,9 +28,9 @@ use reth_interfaces::{
 };
 use reth_primitives::{
     constants::{EMPTY_RECEIPTS, EMPTY_TRANSACTIONS, ETHEREUM_BLOCK_GAS_LIMIT},
-    proofs, Address, Block, BlockBody, BlockHash, BlockHashOrNumber, BlockNumber, Bloom, Bytes,
-    ChainSpec, Header, ReceiptWithBloom, SealedBlock, SealedHeader, TransactionSigned, B256,
-    EMPTY_OMMER_ROOT_HASH, U256,
+    proofs, public_key_to_address, Address, Block, BlockBody, BlockHash, BlockHashOrNumber,
+    BlockNumber, Bloom, Bytes, ChainSpec, Header, ReceiptWithBloom, SealedBlock, SealedHeader,
+    TransactionSigned, B256, EMPTY_OMMER_ROOT_HASH, U256,
 };
 use reth_provider::{BlockExecutor, BundleStateWithReceipts, StateProviderFactory};
 use reth_revm::{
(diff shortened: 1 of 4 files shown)


arminsabouri (approved):
This is awesome! Good work

crates/consensus/authority/src/task.rs
```diff
@@ -132,6 +136,7 @@ where
                 Err(error) => match error {
                     TryRecvError::Empty => {
                         debug!(target: "consensus::authority", "No new blocks from peers");
+                        continue
```
BatiGencho, line 139: @0xBEEFCAF3 without the continue, the first leg is incompatible.

crates/consensus/authority/src/task.rs
```diff
@@ -316,7 +312,10 @@ where
             ) {
                 Ok(ret) => ret,
                 Err(err) => {
-                    warn!(target: "consensus::authority", ?err, "failed to execute block")
+                    error!(target: "consensus::authority", ?err, "failed to execute block");
```
BatiGencho, line 315: @0xBEEFCAF3 Again, the leg is incompatible. I believe if the build fails the txs ought to be requeued for re-execution, right ?

arminsabouri: If there is a tx in the block template that is invalid and causes the whole template to fail that should get re-queued. But yes ideally those that are valid are re-queued

arminsabouri (approved):

crates/consensus/authority/src/task.rs
BatiGencho: great, so will keep it then

scottmillner (approved):
🔥

crates/consensus/authority/src/lib.rs
```diff
@@ -239,6 +239,8 @@ impl StorageInner {
         transactions: &Vec<TransactionSigned>,
         chain_spec: &Arc<ChainSpec>,
         vote: &Option<(secp256k1::PublicKey, Vote)>,
+        sk: &secp256k1::SecretKey,
+        secp: &secp256k1::Secp256k1<secp256k1::All>,
```
stevenroose, line 243: Just 2 small pieces of feedback: if you only need the public key in your function, just take the public key as argument instead of the secret key. This would also avoid having to take the secp context argument.

And about the secp context, it's really just a nit, but to calculate public key from secret key, you only need the signing context so in a library API (here it's probably not really useful in the end), you could take `secp256k1::Secp256k1<impl secp256k1::Signing>` as an argument. (Cfg the `from_secret_key` function signature.)

crates/interfaces/src/consensus.rs
```diff
@@ -189,6 +189,11 @@ pub enum ConsensusError {
     /// Error when the authority is not in turn.
     #[error("Authority not in turn")]
     AuthorityNotInTurn,
+
+    /// Error when the assigned block beneficiary is not within the authorized authorities.
+    #[error("Block beneficiary is not within the authorities list")]
+    BlockBeneficiaryIsNotAuthority,
```
stevenroose, line 195: Be careful to use the beneficiary to deduce the creator of the block :p It would be fun if you could get someone slashed by putting his address in the fee beneficiary :)

PR #76, opened 2024-01-15
feature: poa task split into separate subtasks

BatiGencho:
This PR simply splits up the poa task into a block fetcher and block builder subfiles for convenience.


BatiGencho:
Closing as a different restructure has been done.

PR #81, opened 2024-01-24
fixes to poa events and added events notifier

BatiGencho:
This PR aims to fix some of the poa-related events and event-streamers which were incorrectly imported from node and additionally introduces a notification client which currently supports sending messages to a slack webhook that would appear in our slack in a "testnet_notifications" channel e.g. The underlying client could also be changed easily.

### DIFF
--- a/bin/reth/src/args/rpc_server_args.rs
+++ b/bin/reth/src/args/rpc_server_args.rs
@@ -194,6 +194,12 @@ pub struct RpcServerArgs {
     /// The metrics will be served at the given interface and port.
     #[arg(long, value_name = "BITCOIN_BLOCK_SOURCE", value_parser = parse_url, help_heading = "Btc_block_source")]
     pub btc_block_source: Url,
+
+    /// Notifications Webhook Url
+    ///
+    /// Notifications webhook to send various reports to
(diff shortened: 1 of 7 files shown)


bin/reth/src/args/rpc_server_args.rs
```diff
@@ -194,6 +194,12 @@ pub struct RpcServerArgs {
     /// The metrics will be served at the given interface and port.
     #[arg(long, value_name = "BITCOIN_BLOCK_SOURCE", value_parser = parse_url, help_heading = "Btc_block_source")]
     pub btc_block_source: Url,
+
+    /// Notifications Webhook Url
+    ///
+    /// Notifications webhook to send various reports to
+    #[arg(long, value_name = "NOTIFICATIONS_WEBHOOK_URL", value_parser = parse_url, help_heading = "Notifications Webhook url", required = false)]
+    pub notifications_webhook_url: Option<Url>,
```
arminsabouri: Nit: consider changing this to be specific for slack
`slack_notifications_webhook_url `

arminsabouri (commented):
Nice. Awesome integration. Had one small NIT

scottmillner (approved):
cool stuff! this could definitely come in handy
🔥 💪

bin/reth/src/args/rpc_server_args.rs
BatiGencho: fixed

arminsabouri (approved):

PR #82, opened 2024-01-24
Scott/314 feature block produce should get payload to build from engine api

scottmillner:
Included in pr:

- CL uses new beacon messages to use payload builder to get block to produce
- Cl no longer use tx pool
- Cl no longer notifies blockchain tree after producing new block (beacon engine takes care of this bc of FCU)

### DIFF
--- a/crates/consensus/authority/src/block_builder.rs
+++ b/crates/consensus/authority/src/block_builder.rs
@@ -1,14 +1,13 @@
 use crate::{engine_util, task::BlockProductionTask};
+use reth_beacon_consensus::BeaconEngineMessage;
 use reth_eth_wire::NewBlock;
 use reth_interfaces::blockchain_tree::{
     BlockValidationKind::SkipStateRootValidation, BlockchainTreeEngine,
 };
-use reth_payload_builder::{PayloadBuilderAttributes, PayloadId};
 use reth_primitives::{public_key_to_address, Block, SealedBlockWithSenders, B256};
-use reth_provider::{BlockReaderIdExt, CanonChainTracker, Chain, StateProviderFactory};
+use reth_provider::{BlockReaderIdExt, CanonChainTracker, StateProviderFactory};
 use reth_rpc_types::engine::PayloadAttributes;
 use ruint::Uint;
-use std::sync::Arc;
 use tracing::{error, info};
 
 impl<Client> BlockProductionTask<Client>
@@ -40,64 +39,86 @@ where
         let authority_pub_key = secp256k1::PublicKey::from_secret_key(&self.secp, &self.sk);
         let suggested_fee_recipient = public_key_to_address(authority_pub_key);
 
-        let attr_result = PayloadBuilderAttributes::try_new(
-            best_hash,
-            PayloadAttributes {
-                timestamp: 0u64,
-                prev_randao: B256::ZERO,
-                suggested_fee_recipient,
-                withdrawals: None,
-                parent_beacon_block_root: None,
-            },
-        );
-        if let Err(err) = attr_result {
-            error!("Failed to create payload attributes, err: {:?}", err);
+        let payload_attributes = PayloadAttributes {
+            timestamp: 0_u64,
+            prev_randao: B256::ZERO, // only relevant for PoS
+            suggested_fee_recipient,
+            withdrawals: None,              // only relevant for PoS
+            parent_beacon_block_root: None, // only relevant for PoS
+        };
+
+        // start a new payload job and get the id
+        let (tx, rx) = tokio::sync::oneshot::channel();
+        let result = self.to_engine.send(BeaconEngineMessage::StartNewPayload {
+            payload_attributes,
+            parent: best_hash,
+            tx,
+        });
+
+        let payload_id = match result {
+            Ok(_) => {
+                let recv = rx.await;
+                match recv {
+                    Ok(payload_id) => match payload_id {
+                        Ok(payload_id) => payload_id,
+                        Err(e) => {
+                            error!(target: "consensus::authority", ?e, "Failed to start new payload");
+                            return
+                        }
+                    },
+                    Err(e) => {
+                        error!(target: "consensus::authority", ?e, "Failed to receive payload id");
+                        return
+                    }
+                }
+            }
+            Err(e) => {
+                error!(target: "consensus::authority", ?e, "Failed to send start new payload request");
+                return
+            }
+        };
+
+        // get payload by id
+        let (tx, rx) = tokio::sync::oneshot::channel();
+        let result = self.to_engine.send(BeaconEngineMessage::BestPayload { tx, payload_id });
+
+        let best_transactions = match result {
+            Ok(_) => {
+                let recv = rx.await;
+                match recv {
+                    Ok(payload) => {
+                        if let Some(payload) = payload {
+                            payload.block().clone().body
+                        } else {
+                            info!(target: "consensus::authority", "No best payload received");
+                            return
+                        }
+                    }
+                    Err(e) => {
+                        error!(target: "consensus::authority", ?e, "Failed to receive best payload");
+                        return
+                    }
+                }
+            }
+            Err(e) => {
+                error!(target: "consensus::authority", ?e, "Failed to send best payload request");
+                return
+            }
+        };
+
+        if best_transactions.is_empty() {
+            info!(target: "consensus::authority", "No transactions in best payload");
             return
         }
 
-        let attr = attr_result.expect("valid payload attributes");
-        let mut id: PayloadId = attr.id;
-        let recv = self.payload_store.send_new_payload(attr);
-
-        match recv.await {
-            Ok(res) => match res {
-                Ok(payload_id) => {
-                    info!("Payload builder sent new payload, {:?}", payload_id);
-                    id = payload_id;
-                }
-                Err(err) => {
-                    error!("Payload builder failed to send new payload, err: {:?}", err);
-                }
-            },
-            Err(err) => {
-                error!("Payload builder failed to send new payload, err: {:?}", err);
-            }
-        }
-
-        let mut best_txs = vec![];
-        match self.payload_store.best_payload(id).await {
-            Some(binding) => {
-                let payload = binding.unwrap();
-                best_txs = payload.block().clone().body;
-            }
-            None => {
-                println!("No payload found");
-                // Handle the case when no payload is found if needed
-            }
-        }
-
-        let (transactions, senders): (Vec<_>, Vec<_>) = best_txs
+        let (transactions, senders): (Vec<_>, Vec<_>) = best_transactions
             .into_iter()
             .map(|tx| {
-                let recovered = tx.clone().try_into_ecrecovered().unwrap();
+                let recovered = tx.clone().try_into_ecrecovered().expect("valid tx");
                 let signer = recovered.signer();
                 (tx, signer)
             })
             .unzip();
-
-        if transactions.is_empty() {
-            return
-        }
 
         let recent_bitcoin_block_header = *self.bitcoin_block_header.read().await;
         let authority_signers = storage.authorities.clone();
--- a/crates/consensus/authority/src/block_builder.rs
+++ b/crates/consensus/authority/src/block_builder.rs
@@ -3,14 +3,15 @@ use reth_eth_wire::NewBlock;
 use reth_interfaces::blockchain_tree::{
     BlockValidationKind::SkipStateRootValidation, BlockchainTreeEngine,
 };
-use reth_primitives::{Block, IntoRecoveredTransaction, SealedBlockWithSenders};
+use reth_payload_builder::{PayloadBuilderAttributes, PayloadId};
+use reth_primitives::{public_key_to_address, Block, SealedBlockWithSenders, B256};
 use reth_provider::{BlockReaderIdExt, CanonChainTracker, Chain, StateProviderFactory};
-use reth_transaction_pool::TransactionPool;
+use reth_rpc_types::engine::PayloadAttributes;
 use ruint::Uint;
-use std::{sync::Arc, task::Poll};
+use std::sync::Arc;
 use tracing::{error, info};
 
-impl<Client, Pool: TransactionPool> BlockProductionTask<Client, Pool>
+impl<Client> BlockProductionTask<Client>
 where
     Client: BlockReaderIdExt
         + StateProviderFactory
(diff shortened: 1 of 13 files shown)


crates/consensus/authority/src/block_builder.rs
```diff
@@ -1,72 +1,126 @@
...
 
         if !is_inturn {
             info!(target: "consensus::authority", "Not in turn, skipping");
             return
         }
 
-        // Check if we have transactions to insert
-        if self.queued.is_empty() || !is_inturn {
-            info!(target: "consensus::authority", "Txs list is empty, skipping");
-            // nothing to insert
```
scottmillner, line 50: do we need to do anything with pipeline events?

arminsabouri:
Looks like lint fmt and is failing. The clippy jobs we fail but this one should pass

crates/consensus/authority/src/block_builder.rs
```diff
@@ -1,72 +1,126 @@
...
+                        error!(target: "consensus::authority", ?e, "Failed to receive best payload");
+                        return
+                    }
+                }
+            }
+            Err(e) => {
+                error!(target: "consensus::authority", ?e, "Failed to send best payload request");
+                return
+            }
+        };
```
arminsabouri: Can this whole piece be abstracted out to two different engine API util methods in crates/authority

crates/consensus/beacon/src/engine/mod.rs
```diff
@@ -1876,6 +1876,30 @@ where
                         }
                         BeaconEngineMessage::EventListener(tx) => {
                             this.listeners.push_listener(tx);
+                        },
+                        BeaconEngineMessage::StartNewPayload { payload_attributes, tx, parent } => {
+                            let payload_builder = this.payload_builder.clone();
+                            tokio::spawn(async move {
```
arminsabouri, line 1882: Is it fine to use tokio here? not the task manager?

crates/consensus/beacon/src/engine/mod.rs
```diff
@@ -1876,6 +1876,30 @@ where
...
+                                let attributes_result = PayloadBuilderAttributes::try_new(parent, payload_attributes);
+                                let attributes = attributes_result.expect("valid payload attributes");
+                                let recv = payload_builder.send_new_payload(attributes);
+                                let result = recv.await.expect("payload builder channel to be open");
+                                let _ = tx.send(result);
+                            });
+                        },
+                        BeaconEngineMessage::BestPayload { tx, payload_id} => {
+                            info!(target: "consensus::engine", "Resolving payload {}", payload_id);
+                            let payload_builder = this.payload_builder.clone();
```
arminsabouri, line 1903: Do we need this clone?

crates/consensus/beacon/src/engine/mod.rs
```diff
@@ -1876,6 +1876,30 @@ where
                         }
                         BeaconEngineMessage::EventListener(tx) => {
                             this.listeners.push_listener(tx);
+                        },
+                        BeaconEngineMessage::StartNewPayload { payload_attributes, tx, parent } => {
+                            let payload_builder = this.payload_builder.clone();
+                            tokio::spawn(async move {
+                                let attributes_result = PayloadBuilderAttributes::try_new(parent, payload_attributes);
+                                let attributes = attributes_result.expect("valid payload attributes");
```
arminsabouri: If this is None should we return with some error?

crates/consensus/beacon/src/engine/mod.rs
scottmillner: my thinking was these tasks aren't critical so don't need to do anything if they fail
and tokio runtime automatically cleans up the task when they go out of scope

the task executor is passed to `BeaconConsensusEngine::new` and is set on the `sync` field but it's not public.  Could make it public and use it but not sure what we gain from that?

crates/consensus/authority/src/epoch_manager.rs
```diff
@@ -60,7 +49,8 @@ where
             .flatten();
 
         if latest_header.is_none() {
-            return (Poll::Pending, false)
+            info!("No latest header found");
```
BatiGencho: Should not this be rather a warning ?

crates/consensus/authority/src/epoch_manager.rs
```diff
@@ -69,26 +59,20 @@ where
...
             let latest_signer = utils::recovery_authority(&latest_header).unwrap();
             let current_ts = utils::unix_timestamp();
-            if let Err(_) = utils::validate_current_signer_against_last(
+            if utils::validate_current_signer_against_last(
                 (latest_signer, latest_header.timestamp / 60),
                 (signer_pk, current_ts / 60),
-            ) {
-                return (Poll::Pending, false)
+            )
+            .is_err()
```
BatiGencho: Can we log what the error is here  before returning Polling?

crates/consensus/beacon/src/engine/mod.rs
BatiGencho: I believe we need to use tokio as reth is using it in addition to hyper. I would say yes.

crates/consensus/beacon/src/engine/mod.rs
BatiGencho: I think we need as the payload builder is being moved into the spawned task below in order to have a 'static lifetime.

crates/consensus/beacon/src/engine/mod.rs
```diff
@@ -1876,6 +1876,30 @@ where
...
+                                let _ = tx.send(result);
+                            });
+                        },
+                        BeaconEngineMessage::BestPayload { tx, payload_id} => {
+                            info!(target: "consensus::engine", "Resolving payload {}", payload_id);
+                            let payload_builder = this.payload_builder.clone();
+                            tokio::spawn(async move {
+                                let best_payload = match payload_builder.best_payload(payload_id).await {
+                                    Some(payload) => match payload {
+                                        Ok(payload) => Some(payload),
```
BatiGencho: Can we make this a bit nicer by having .ok(), ok_or() , ok_or_else() .... ?

crates/consensus/authority/src/epoch_manager.rs
```diff
@@ -69,26 +59,20 @@ where
...
+            {
+                return false
             }
         }
 
         drop(storage);
         let is_inturn = utils::is_inturn(authority_len, signer_index as u64);
         info!("Epoch manager inturn: {}", is_inturn);
 
-        // drain the pool
```
BatiGencho, line 84: why do we not want to remove that here ?

crates/consensus/authority/src/block_builder.rs
```diff
@@ -1,72 +1,126 @@
...
+
+        // get payload by id
+        let (tx, rx) = tokio::sync::oneshot::channel();
+        let result = self.to_engine.send(BeaconEngineMessage::BestPayload { tx, payload_id });
+
+        let best_transactions = match result {
+            Ok(_) => {
+                let recv = rx.await;
+                match recv {
```
BatiGencho: can we do here like:

match rx.await {
 Ok(payload) => payload.and_then(|p| p.block().body) .... ?

crates/consensus/authority/src/block_builder.rs
```diff
@@ -1,72 +1,126 @@
...
+
+        // start a new payload job and get the id
+        let (tx, rx) = tokio::sync::oneshot::channel();
+        let result = self.to_engine.send(BeaconEngineMessage::StartNewPayload {
+            payload_attributes,
+            parent: best_hash,
+            tx,
+        });
+
+        let payload_id = match result {
```
BatiGencho: In fact reg. my comment below, can we create a func that would do what is currently being done and returns the value. A bit like an abstraction since both blocks are quite repetitive ?

crates/consensus/beacon/src/engine/mod.rs
scottmillner: yes bc it's getting move into the task

crates/consensus/beacon/src/engine/mod.rs
scottmillner: ~~my thinking was if we're passing invalid payload attributes then the node can't produce blocks and should panic, thoughts?~~

end up logging the error and returning

crates/consensus/authority/src/epoch_manager.rs
scottmillner: we're removing tx pool from all of CL and letting the Beacon Engine handle the tx pool

the txs will get removed from the pool as long as we're send FCU messages correctly

crates/consensus/beacon/src/engine/mod.rs
```diff
@@ -1877,6 +1877,41 @@ where
                         BeaconEngineMessage::EventListener(tx) => {
                             this.listeners.push_listener(tx);
                         }
+                        BeaconEngineMessage::StartNewPayload { payload_attributes, tx, parent } => {
+                            let payload_builder = this.payload_builder.clone();
+                            tokio::spawn(async move {
+                                let attributes_result =
+                                    PayloadBuilderAttributes::try_new(parent, payload_attributes);
+                                let attributes =
+                                    attributes_result.expect("valid payload attributes");
```
BatiGencho: if this fails we will never respond and panic. Maybe try to somehow handle and exit the task on Err ?

crates/consensus/beacon/src/engine/mod.rs
```diff
@@ -1877,6 +1877,41 @@ where
...
+                                let recv = payload_builder.send_new_payload(attributes);
+                                let result =
+                                    recv.await.expect("payload builder channel to be open");
```
BatiGencho: maybe same here : do .ok() and if Some await on receiver if not log error and exit the task ?

crates/consensus/authority/src/block_builder.rs
scottmillner: done

crates/consensus/authority/src/epoch_manager.rs
scottmillner: done

crates/consensus/authority/src/epoch_manager.rs
scottmillner: done

crates/consensus/beacon/src/engine/mod.rs
scottmillner: done

crates/consensus/authority/src/block_builder.rs
scottmillner: done

crates/consensus/authority/src/block_builder.rs
scottmillner: done

crates/consensus/beacon/src/engine/mod.rs
scottmillner: done

crates/consensus/beacon/src/engine/mod.rs
scottmillner: done

BatiGencho (approved):

crates/consensus/authority/src/block_builder.rs
arminsabouri: I don't think so. Not even sure why its passed in

crates/consensus/authority/src/block_builder.rs
```diff
@@ -1,72 +1,83 @@
...
+            withdrawals: None,              // only relevant for PoS
+            parent_beacon_block_root: None, // only relevant for PoS
+        };
+
+        // start new payload
+        let payload_id =
+            engine_util::start_new_payload(self.to_engine.clone(), payload_attributes, best_hash)
+                .await;
+
+        if payload_id.is_none() {
```
arminsabouri: Can we log a warning here? 
Is there anything worth erroring?

crates/consensus/authority/src/block_builder.rs
```diff
@@ -1,72 +1,83 @@
...
+        // get payload by id
+        let best_transactions = engine_util::get_best_transactions_from_payload(
+            self.to_engine.clone(),
+            payload_id.expect("payload id exists"),
+        )
+        .await;
 
-        let (transactions, senders): (Vec<_>, Vec<_>) = transactions
+        if best_transactions.is_none() {
+            return
```
arminsabouri, line 65: same thing here

crates/consensus/authority/src/engine_util.rs
```diff
@@ -121,10 +123,78 @@ pub(crate) async fn send_fork_choice_update_payload(
...
+    let (tx, rx) = tokio::sync::oneshot::channel();
+    let result =
+        to_engine.send(BeaconEngineMessage::StartNewPayload { payload_attributes, parent, tx });
+
+    match result {
+        Ok(_) => match rx.await {
+            Ok(payload_id) => payload_id.ok(),
+            Err(e) => {
+                error!(target: "consensus::authority", ?e, "Receiver error, channel closed");
+                None
```
arminsabouri: Should these be errors?

crates/consensus/authority/src/sync.rs
```diff
@@ -56,67 +56,68 @@ impl SyncController {
...
-            assert_eq!(state.head_block_hash, blockhash);
-        }
-
-        // Cancel the spawned task
-        handle.abort();
-
-        // TODO (scott) update test to check engine response when function is updated
-    }
-}
+// Todo (scott) refactor test
```
arminsabouri, line 59: Can we create an issue for this so we dont forget

crates/consensus/authority/src/block_builder.rs
scottmillner: logging now

crates/consensus/authority/src/block_builder.rs
scottmillner: logging now

crates/consensus/authority/src/engine_util.rs
scottmillner: yes, good point

refactored to return Result with error instead Option with None

crates/consensus/authority/src/sync.rs
scottmillner: done: https://github.com/botanix-labs/botanix/issues/319

scottmillner:
final pr here: https://github.com/botanix-labs/Macbeth/pull/85

easier to start fresh branch of latest to avoid merge conflicts

PR #83, opened 2024-01-25
Frost signing scaffolding

arminsabouri:
The idea with this struct is to have a stateful data structure that is aware of the DKG key status and aware of the current signing session. 
What I want to accomplish next is to have only the secrets persist somewhere and for the DKG struct to be initialized from secrets in the DB. There might be somewhere better to save the secrets. Still an open question 

Another open question is how to generate the frost identifier.


bin/btc-server/Cargo.toml
```diff
@@ -3,13 +3,15 @@ name = "btc-server"
...
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
 reth-btc-wallet = { path = "../../crates/btc-wallet" }
+client = {path = "./client"}
 
 anyhow = "1.0.71"
 bitcoin = { version = "0.30.2", features = [ "base64", "rand", "serde" ] }
+# TODO update this to the zcash frost library when tr is merged in
+frost-secp256k1-tr = { git = "https://github.com/mimoo/frost", branch = "mimoo/fix5",  features = ["serde", "serialization"] }
```
BatiGencho, line 14: In your example previously, I see you are targeting the main branch, here you are using a given branch, is there a particular reason for that ?

bin/btc-server/client-example/src/main.rs
```diff
@@ -1,13 +0,0 @@
...
-
-#[tokio::main]
-async fn main() -> Result<(), Box<dyn std::error::Error>> {
-    let mut client = BtcServerClient::connect("http://localhost:8080").await.unwrap();
-    let request = Request::new(Empty {}); // Use the generated Empty struct
-
-    let response = client.get_public_key(request).await.unwrap();
-    println!("Public Key: {}", response.get_ref().publickey);
-
-    Ok(())
```
BatiGencho, line 12: I personally feel we should keep this file and maybe extend it a bit with frost unary req just to show how its been used and how we can spin up a client ??

bin/btc-server/src/main.rs
```diff
@@ -131,7 +186,7 @@ impl App {
                 bdk_utxos,
                 bdk::FeeRate::from_sat_per_vb(fee_rate.to_sat_per_vb_ceil() as f32),
                 output.value,
-                self.change_script.as_script(), // drain_script
+                self.change_script.as_ref().expect("change script").as_script(), // drain_script
```
BatiGencho, line 199: should we not map the expect to a prop error ?

bin/btc-server/src/main.rs
```diff
@@ -141,9 +196,10 @@ impl App {
             .filter_map(|s| if s.is_some() { s } else { None })
             .collect::<Vec<_>>();
         let change = match selection.excess {
-            bdk::wallet::coin_selection::Excess::Change { amount, .. } => {
-                Some(TxOut { script_pubkey: self.change_script.clone(), value: amount })
-            }
+            bdk::wallet::coin_selection::Excess::Change { amount, .. } => Some(TxOut {
+                script_pubkey: self.change_script.as_ref().expect("change script").clone(),
```
BatiGencho, line 211: same here

bin/btc-server/proto/btc_server.proto
```diff
@@ -36,7 +44,13 @@ message MakeTxResponse {
 
 message GetPublicKeyResponse {
     // hex encoded
-    string publickey = 1; 
+    string publickey = 1;
 }
 
 message Empty {}
```
BatiGencho, line 50: When it comes to empty, I believe we could elegantly make use of : https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/empty.proto

which is the official google protobuf impl and there is also a lot of cool stuff they have added we could make use of. 

Here is one of my repos that I created in the past: https://github.com/BatiGencho/orderbook-rs/blob/main/build.rs

We could do something similar as above in a pre-compile build.rs file and include google's protobufs too - I believe htis is the standard way that its being done. Just a suggestion

bin/btc-server/Cargo.toml
arminsabouri: Yes. Taproot signings is not merged in to the main branch yet. 
https://github.com/ZcashFoundation/frost/pull/584

bin/btc-server/client-example/src/main.rs
arminsabouri: The reason I deleted it is the integration tests largely do the same thing. Lmk if you agree. If no happy to bring them back

bin/btc-server/Cargo.toml
BatiGencho: all right, ok, did not know that. all good

bin/btc-server/client-example/src/main.rs
BatiGencho: I realized this ... I guess its fine then,saw that are doing the same in the int. tests, whoever wants to have a look at could just go there and acquaint himself with the code, all good.

bin/btc-server/src/main.rs
arminsabouri: My thought was before the server actually starts its fine to panic. What do you think

bin/btc-server/src/main.rs
arminsabouri: Here we dont want to panic. Good call. Will fix

bin/btc-server/proto/btc_server.proto
arminsabouri: Whats the benefit of using google's empty?

BatiGencho: None, its just a more accepted industry standard I would say. googles protos also come with a lot of other handy impls, but we dont really need them tbh as of this stage.

bin/btc-server/src/main.rs
BatiGencho: I get your point, in that case all good. This is the main file. It might not be a bad idea to actually use anyhow::Context

e.g. restult.contect(|| "change script".to_string() ) ? .

or eyre. Anywya, expect is fine.

bin/btc-server/proto/btc_server.proto
arminsabouri: Ok we can add them down the line

bin/btc-server/src/main.rs
arminsabouri: Fixed. Added a is_none check and return a new error variant

BatiGencho: nice

BatiGencho (approved):

PR #86, opened 2024-01-30
feature: block signer should never sign a fork deeper than 1 block

BatiGencho:
Ticket description: https://github.com/botanix-labs/botanix/issues/307


crates/blockchain-tree/src/blockchain_tree.rs
```diff
@@ -581,13 +587,18 @@ impl<DB: Database, EF: ExecutorFactory> BlockchainTree<DB, EF> {
         hashes
     }
 
+    /// A mapper for the method above
```
arminsabouri:
```suggestion
    /// A wapper for the method above
```

crates/blockchain-tree/src/blockchain_tree.rs
```diff
@@ -581,13 +587,18 @@ impl<DB: Database, EF: ExecutorFactory> BlockchainTree<DB, EF> {
...
+    }
+
     /// Get the block at which the given chain forks off the current canonical chain.
     ///
     /// This is used to figure out what kind of state provider the executor should use to execute
     /// the block on
     ///
     /// Returns `None` if the chain is unknown.
-    fn canonical_fork(&self, chain_id: BlockChainId) -> Option<ForkBlock> {
+    pub fn canonical_fork(&self, chain_id: BlockChainId) -> Option<ForkBlock> {
```
arminsabouri: Can this still have private visibility?

crates/consensus/authority/src/fork_selector.rs
```diff
@@ -0,0 +1,59 @@
...
+        let all_chain_ids = self.client.chain_ids();
+
+        // get all block hashes from a sidechain that are not part of the canonical chain.
+        for chain_id in all_chain_ids {
+            let chain_id_blocks_hashes = self.client.all_chain_hashes(chain_id);
+            if chain_id_blocks_hashes.keys().len() > MAX_FORK_DEPTH {
+                continue
+            }
+
+            let mut total_difficulty = U256::from(0);
```
arminsabouri: `total_difficulty` should always be 0 right? I think we have setting it to a default value and there is no consenus  check. 

If so, maybe there is a better thing to sort off to pick the best fork? timestamp? Ordering of pk in the federation?

crates/interfaces/src/consensus.rs
```diff
@@ -2,11 +2,23 @@ use reth_primitives::{
...
 pub use reth_rpc_types::engine::ForkchoiceState;
 
+use crate::blockchain_tree::BlockchainTreeEngine;
+
+/// Selection criteria for block forks
+pub trait BlockForkSelectionCriteria<Client>
+where
+    Client: BlockchainTreeEngine,
+{
+    /// the selection function
```
arminsabouri: Maybe we can expand on this description

```
this function selects which blockchain fork is correct according to its consensus rules
```

crates/blockchain-tree/src/config.rs
```diff
@@ -61,17 +61,36 @@ impl BlockchainTreeConfig {
...
+
     /// Return number of additional canonical block hashes that we need to retain
     /// in order to have enough information for EVM execution.
     pub fn num_of_additional_canonical_block_hashes(&self) -> u64 {
         self.num_of_additional_canonical_block_hashes
     }
 
+    /// Set number of additional canonical block hashes that we need to retain
+    /// in order to have enough information for EVM execution.
+    pub fn set_num_of_additional_canonical_block_hashes(
```
arminsabouri: Does this get used?

crates/consensus/authority/src/fork_selector.rs
```diff
@@ -0,0 +1,59 @@
...
+                if let Some(sealed_block) = sealed_block {
+                    total_difficulty += sealed_block.difficulty;
+                }
+            }
+            forks.insert(total_difficulty, chain_id);
+        }
+
+        // since keys are ordered, get the maximum key value
+        if let Some((_total_difficulty, chain_id)) = forks.last_key_value() {
+            let best_fork_blocknumhash = self.client.canonical_fork(*chain_id);
```
arminsabouri, line 42: Just a question?
Do we need to call FCU? Or call the canon manager or does this suffice?

arminsabouri (commented):
This is a lot of progress!!!! 
I think we are on the right track
Just had a few questions

crates/blockchain-tree/src/config.rs
scottmillner: we don't need this method right?

as in that value will always end up being 256 so BLOCKHASH opcode works right?

based on the comments in blockchain tree `config`:

`/// The total number of block hashes retained in-memory will be
/// max(additional_canonical_block_hashes, max_reorg_depth), and for Ethereum that would
/// be 256. It covers both number of blocks required for reorg, and number of blocks
/// required for `BLOCKHASH` EVM opcode.`

crates/consensus/authority/src/fork_selector.rs
scottmillner: how about the block with the earliest timestamp? that enforces round robin in my mind...as in, i was in turn before you but bc of latency (or whatever) other blocks after me are competing with my valid block

that make sense?

scottmillner:
> This is a lot of progress!!!! I think we are on the right track Just had a few questions

agreed! nice start!! 🔥 💪

crates/blockchain-tree/src/blockchain_tree.rs
BatiGencho: done, thx

crates/blockchain-tree/src/blockchain_tree.rs
BatiGencho: sure, fixed

crates/consensus/authority/src/fork_selector.rs
BatiGencho: I am now sorting after the ts as suggested, thanks for the advice scott

crates/interfaces/src/consensus.rs
BatiGencho: changed, thanks

crates/blockchain-tree/src/config.rs
BatiGencho: removed, had it added just for completeness, but it was fully redundant

crates/consensus/authority/src/fork_selector.rs
BatiGencho: Thats the part I am still not 100% sure. The FCU uses the same tree as here, but it might be worth exploring that a bit further

arminsabouri:
#191 superceeds this. Closing

PR #87, opened 2024-01-30
Scott/319 refactor poa tests

scottmillner:


crates/consensus/authority/src/sync.rs
```diff
@@ -56,68 +56,69 @@ impl SyncController {
...
+        tokio::time::sleep(Duration::from_millis(100)).await;
+
+        // Assert that the message with peer tip was sent
+        if let BeaconEngineMessage::ForkchoiceUpdated { state, .. } = engine_rx.try_recv().unwrap()
+        {
+            assert_eq!(state.head_block_hash, blockhash);
+        }
+
+        // Cancel the spawned task
+        handle.abort();
```
BatiGencho: I dont think you need to cancel it here as its a unit test and once it ends (either panicks or reaches end of scope) the spawned thread should be closed by the context itself.

crates/consensus/authority/tests/it/poa_testnet.rs
```diff
@@ -309,12 +325,18 @@ pub async fn test_poa_testnet() {
...
+            payload_client.send_invalid(RECEIVER_ADDRESS).await;
+        }
+
         // block verfication
         let block_receipts = x.notification.block_receipts();
         println!("Block receipts? {:?}", block_receipts);
         assert_eq!(block_receipts.len(), 1);
         let block_payload = block_receipts.first().cloned().unwrap();
-        assert_eq!(block_payload.1, false);
+        assert!(!block_payload.1);
```
BatiGencho, line 339: Would not this cause the int. test to panic in the first round by sending an invalid tx ? or am I missing something :) The idea was for it to run N times. If that is a stupid question, just ignore it :)

scottmillner: no it doesn't panic bc i handle the error and print it in the `send_invalid` method

and the payload is still received

i ran it to completion and everything worked

crates/consensus/authority/src/sync.rs
scottmillner: removed `handle.abort()`
thanks!

crates/consensus/authority/tests/it/poa_testnet.rs
BatiGencho: all good, ok

BatiGencho (approved):

PR #88, opened 2024-01-31
added frost messaging

BatiGencho:


crates/net/network-api/src/lib.rs
```diff
@@ -67,6 +67,15 @@ pub trait PeersInfo: Send + Sync {
     fn local_node_record(&self) -> NodeRecord;
 }
 
+/// Provides general purpose information about Peers in the network.
+#[async_trait]
+pub trait FrostMessaging: Send + Sync {
+    /// Returns how many peers the network is currently connected to.
+    ///
+    /// Note: this should only include established connections and _not_ ongoing attempts.
```
scottmillner, line 76: nit: spelling typo `send_forst_msg` > `send_frost_msg`

crates/net/network/src/manager.rs
```diff
@@ -520,6 +520,10 @@ where
             PeerMessage::Other(other) => {
                 debug!(target: "net", message_id=%other.id, "Ignoring unsupported message");
             }
```
scottmillner: why do we need `PeerMessage::Frost` if we have `NetworkHandleMessage::FrostMessage`?

just for logging?

bin/reth/src/poa/mod.rs
```diff
@@ -555,6 +555,13 @@ impl<Ext: RethCliExt> PoaNodeCommand<Ext> {
             }),
         );
```
scottmillner, line 562: do we need a new task?

epoch block producer will always be the first node to send a frost message to initiate DKG and you have added the `NetworkHandleMessage::FrostMessage` which gets handled in `on_handle_message`

so nodes will be aware of frost messages and then need to do the next frost step based on what the message is so not sure if we need a new task since we have `on_handle_message` which is already in the NetworkManager loop?

maybe im misunderstanding the flow?

arminsabouri: >epoch block producer will always be the first node to send a frost message to initiate DKG 

Dkg must happen well before any signing can occur. Tbh im not sure when probably as soon as they realize they dont have the pubkey package.

arminsabouri: Generally it might make sense to have a frost task that only handles messages from peers. Keep the block producer task only focused on block production. Maybe they can communicate via shared message bus?

crates/consensus/authority/src/frost.rs
```diff
@@ -0,0 +1,113 @@
+use crate::{epoch_manager::EpochManager, Storage};
```
arminsabouri: Nit: can the name of this file be frost_task.rs?

crates/consensus/authority/src/frost.rs
```diff
@@ -0,0 +1,113 @@
...
+use secp256k1::{All, Secp256k1};
+use tracing::info;
+use std::sync::Arc;
+
+use tokio::sync::{mpsc::UnboundedSender, RwLock};
+use tokio_stream::wrappers::UnboundedReceiverStream;
+
+use client::BtcServerClient;
+
+pub struct FrostTask<Client> {
```
arminsabouri: How many of these fields do we actually need? I imagine just the client and network handle?

crates/consensus/authority/src/frost.rs
```diff
@@ -0,0 +1,113 @@
...
+            network_handle,
+            task_executor,
+        }
+    }
+
+    pub async fn start_task(&mut self) -> () {
+        let test_msg = b"Hello Frost!";
+        loop {
+            info!("Sending Frost Message to all peers");
+            let _ = self.network_handle.send_forst_msg(test_msg.to_vec()).await;
```
arminsabouri: lil typo

```suggestion
            let _ = self.network_handle.send_frost_msg(test_msg.to_vec()).await;
```

arminsabouri (commented):
This is great! I think the next steps are message handling logic and message structure

crates/consensus/authority/src/frost.rs
BatiGencho: True, I just took them all as it was a first commit. In fact only the latter are needed

crates/consensus/authority/src/frost.rs
BatiGencho: fixed thanks

crates/consensus/authority/src/frost.rs
BatiGencho: sure, changed

bin/reth/src/poa/mod.rs
BatiGencho: I will keep it for now, I wanted to integrate communication over channels

crates/net/network-api/src/lib.rs
BatiGencho: fixed, thanks

crates/net/network/src/manager.rs
BatiGencho: I had the understanding that both message types are allowed - direct and over the network handle. I am refactoring this bit now and testing it so hold tight.

PR #89, opened 2024-02-01
Add util function to return if bloom filter possibly contains a pegout 

scottmillner:


crates/consensus/authority/src/utils.rs
```diff
@@ -154,3 +154,34 @@ async fn process_botanix_log(
     }
     Ok(())
 }
+
+const MINTING_CONTRACT_ADDRESS: [u8; 20] = hex!("0Ea320990B44236A0cEd0ecC0Fd2b2df33071e78");
+const MINTING_CONTRACT_BURN_TOPIC: [u8; 32] =
+    hex!("922344dc04648c0ce028ecdf9b2c9eed9a6794dbb47b777b54b0cfe069f128aa");
```
arminsabouri: I think we also reference these in the revm crate? Can we create some kind of consensus constants for this?

crates/consensus/authority/src/utils.rs
```diff
@@ -154,3 +154,34 @@ async fn process_botanix_log(
...
+pub(crate) fn bloom_contains_pegout(bloom: Bloom) -> bool {
+    bloom.contains_input(BloomInput::Raw(&MINTING_CONTRACT_ADDRESS)) &&
+        bloom.contains_input(BloomInput::Raw(&MINTING_CONTRACT_BURN_TOPIC))
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
```
arminsabouri, line 179: Would you mind adding a test case that uses a header that contains a pegout

crates/consensus/authority/src/utils.rs
```diff
@@ -154,3 +154,34 @@ async fn process_botanix_log(
     }
     Ok(())
 }
+
+const MINTING_CONTRACT_ADDRESS: [u8; 20] = hex!("0Ea320990B44236A0cEd0ecC0Fd2b2df33071e78");
```
arminsabouri:
```suggestion
const MINTING_CONTRACT_ADDRESS: [u8; 20] = hex!("0ea320990b44236a0ced0ecc0fd2b2df33071e78");
```

Super nit: consistent letter casing in the hex

arminsabouri (commented):
This is awesome 👍 
 I just had a few comments nothing major

arminsabouri:
One more thing. Do you think we can set up the same thing for pegins? Dont want to scope creep. Only if its a easy lift

crates/consensus/authority/src/utils.rs
scottmillner: do you mind if add that in the pr that will have the `epoch_pegouts` method which loops through block headers and checks the bloom filter

for these tests, i just wanna test the underlying util function which the above method will use

crates/consensus/authority/src/utils.rs
scottmillner: ~~im using the checksum address here which we also use in `mint_validation`~~

im going to use the ref in mint_validation which revm uses

crates/consensus/authority/src/utils.rs
scottmillner: we have this in `mint_validation`:
```
lazy_static::lazy_static! {
    pub static ref MINT_TOPIC: B256 = keccak256("Mint(address,uint256,uint32,bytes)");
    pub static ref BURN_TOPIC: B256 = keccak256("Burn(address,uint256,bytes,bytes)");
    pub static ref MINT_CONTRACT_ADDRESS: Address = Address::from_str("0x0Ea320990B44236A0cEd0ecC0Fd2b2df33071e78").unwrap();
    static ref FROST_PUB_KEY: PublicKey = PublicKey::from_str("02d0a67d0b49551c6edfa7f00737b8139a28de6eb7102131c02704f3ad1cf579cd").unwrap();
}
```

do you want to reference those or move them to a concensus constants file?

im gonna reference them for now but can move them if need be

crates/consensus/authority/src/utils.rs
arminsabouri: Yup that sounds fine. Resolving

crates/consensus/authority/src/utils.rs
arminsabouri: Yeah that sounds good to me. We can move them in the future. I dont think it matter too much

arminsabouri (approved):

crates/consensus/authority/src/utils.rs
```diff
@@ -154,3 +158,58 @@ async fn process_botanix_log(
...
+
+pub(crate) fn bloom_contains_pegout(bloom: Bloom) -> bool {
+    bloom_contains_minting_contract_address(bloom) &&
+        bloom.contains_input(BloomInput::Raw(BURN_TOPIC.as_ref()))
+}
+
+pub(crate) fn bloom_contains_pegin(bloom: Bloom) -> bool {
+    bloom_contains_minting_contract_address(bloom) &&
+        bloom.contains_input(BloomInput::Raw(MINT_TOPIC.as_ref()))
+}
```
stevenroose, line 174: just a suggestion, you could define a trait for this as follows:
```rust
pub trait BloomExt: AsRef<Bloom> {
    fn contains_pegout(&self) -> bool {
        self.as_ref().contains_input(...)
    }
}
impl BloomExt for Bloom {}
```

that wa instead of calling
```rust
bloom_contains_pegout(bloom)
```
you can call
```rust
bloom.contains_pegout()
```

PR #91, opened 2024-02-06
feature: add function to return all pegouts in an epoch

scottmillner:
these changes get us to the point where the node can get a list of pegouts (as MakeTxRequest) which could then be passed to the btc_server

remaining steps for to be done in other prs

- add pegouts to cache and refactor to use cache first then `epoch_pegouts`
- refactor node to not send pegouts every block and only do at epoch end


crates/consensus/authority/src/epoch_manager.rs
```diff
@@ -28,6 +38,64 @@ where
...
+    /// # Returns
+    ///
+    /// A vector of `MakeTxRequest` representing the pegouts in the epoch
+    pub(crate) async fn epoch_pegouts(
+        &self,
+        end_block: u64,
+    ) -> Result<Vec<MakeTxRequest>, EpochManagerError> {
+        let start_block = (end_block - EPOCH_LENGTH) + 1;
+        let storage = self.storage.inner.read().await;
+        let mut pegouts: Vec<MakeTxRequest> = vec![];
```
scottmillner: this could probably be more elegant/efficient?

crates/consensus/authority/src/block_builder.rs
```diff
@@ -28,8 +30,27 @@ where
...
-            storage.get_best_block_and_hash().expect("best block exists");
+        let (best_block, best_hash) = storage.get_best_block_and_hash().expect("best block exists");
+
+        // if epoch end block, process pegouts
+        let current_block = best_block + 1;
+        let pegouts = if is_epoch_end(current_block) {
+            match self.epoch_manager.epoch_pegouts(current_block).await {
+                Ok(pegouts) => pegouts,
+                Err(e) => {
+                    panic!("Failed to get pegouts {}", e)
```
BatiGencho: Why do we panic here ? Would not this halt the entire loop and blocl builder ?

crates/consensus/authority/src/epoch_manager.rs
```diff
@@ -28,6 +38,64 @@ where
...
+                Ok(Some(block)) => {
+                    if bloom_contains_pegout(block.header.logs_bloom) {
+                        match storage
+                            .client
+                            .receipts_by_block(BlockHashOrNumber::Number(block.header.number))
+                        {
+                            Ok(Some(receipts)) => {
+                                for receipt in receipts {
+                                    if let Some(pegout) =
+                                        make_tx_request_for_pegout_in_receipt(receipt).await
```
BatiGencho: I would use here tokio::join! macro to put all the Futures and await them concurrently instead of one by one. This will speed it up. As a result you will get `Vec<Result<T>>`

crates/consensus/authority/src/utils.rs
```diff
@@ -22,6 +22,31 @@ pub(crate) enum ProcessBotanixLogError {
...
+/// Returns `Some(MakeTxRequest)` if a pegout is found in the receipt, otherwise returns `None`.
+pub(crate) async fn make_tx_request_for_pegout_in_receipt(
+    receipt: Receipt,
+) -> Option<MakeTxRequest> {
+    if !receipt.success {
+        info!(target: "consensus::authority", "Receipt status code is not success {:?}", receipt);
+        return None
+    }
+    for log in receipt.logs {
+        match make_tx_request_for_pegout(log).await {
```
BatiGencho: same batching as above here ?

BatiGencho: there is also OrderedFeatures if the order matters, I believe it does not so tokio::join! should be all good

crates/consensus/authority/src/utils.rs
```diff
@@ -75,6 +100,37 @@ pub(crate) async fn process_reciepts(
...
+///
+/// # Returns
+///
+/// Returns `Some(MakeTxRequest)` if a pegout is found in the log, otherwise returns `None`.
+async fn make_tx_request_for_pegout(log: Log) -> Option<MakeTxRequest> {
+    for topic in &log.topics {
+        match GenesisContractEvents::try_from(*topic) {
+            Ok(GenesisContractEvents::MintingEvent) => continue,
+            Ok(GenesisContractEvents::BurnEvent) => {
+                let fee_rate = 30u32;
```
BatiGencho: Why do we fix the fee_rate ?

crates/consensus/authority/src/epoch_manager.rs
BatiGencho: You could probably simplify the branching a bit :

Ok(Some(block)) if  bloom_contains_pegout(block.header.logs_bloom) { ... }
Ok(None) => { ... }
Err(e) => { ... }

crates/consensus/authority/src/block_builder.rs
scottmillner: i was thinking this is a critical error: if something goes wrong when processing pegouts, we want the node to halt then another node will come in turn and if that panics as well we know something is seriously wrong

also, if we don't panic and the block gets built, then all the pegouts from that epoch will never be sent since it's assumed all pegouts were handled at the end of the epoch

@0xBEEFCAF3 that reasoning seem correct?

crates/consensus/authority/src/utils.rs
scottmillner: we haven't added the logic yet to dynamically set the fee rate but it is something that we need to do at some point

BatiGencho: fair enough

crates/consensus/authority/src/block_builder.rs
BatiGencho: Thats fine, just wondering if we could then kill the node more gently than just panicking - the latter is like a force majeour event, cannot we somehow trigger a graceful termination here ?

crates/consensus/authority/src/epoch_manager.rs
scottmillner: done

crates/consensus/authority/src/epoch_manager.rs
scottmillner: done

crates/consensus/authority/src/utils.rs
scottmillner: done

BatiGencho (approved):

BatiGencho:
@scottmillner For this MR, I took the liberty and changed the target branch to be  `poa-consensus-backup` which is the one we are currently operating on. Whenever you get the chance, pls fix the 2 conflicts there are and it should be all good.

PR #94, opened 2024-02-12
fix(poa): payload builder 

BatiGencho:
MR context: https://github.com/botanix-labs/botanix/issues/333


crates/consensus/authority/src/engine_util.rs
```diff
@@ -148,37 +146,21 @@ pub(crate) enum StartNewPayloadError {
 /// * `payload_attributes` - The payload attributes.
 /// * `parent` - The parent block hash the payload will be built on.
 pub(crate) async fn start_new_payload<Engine: reth_node_api::EngineTypes>(
```
arminsabouri, line 148: Do the tests for this fn need to be remove/updated?

arminsabouri (approved):
Looks good i Just had the question

crates/primitives/src/constants/mod.rs
```diff
@@ -17,7 +17,7 @@ pub const RETH_CLIENT_VERSION: &str = concat!("reth/v", env!("CARGO_PKG_VERSION"
 pub const SELECTOR_LEN: usize = 4;
 
 /// Maximum extra data size in a block after genesis
-pub const MAXIMUM_EXTRA_DATA_SIZE: usize = 32;
+pub const MAXIMUM_EXTRA_DATA_SIZE: usize = 128;
```
BatiGencho, line 20: @0xBEEFCAF3 I was testing and due to this limit we were crashing. I increased it as I observed we are over 100 in our case, but this is purely empirical. Wdyt ?

arminsabouri (approved):

PR #96, opened 2024-02-12
cargo fmt linting

BatiGencho:
This MR adds linting via cargo fmt.


arminsabouri (approved):
Thanks for taking this on

PR #98, opened 2024-02-13
feature: added poa frost networking with dkg

BatiGencho:
First version of poa frost protocol integrated into reth with support for dkg generation.

There are two open points to resolve:

1. Who is triggering the dkg building process - the currently in turn fed member ? I believe there should be one fed member only while being in turn triggering the process which means we should mark that there is a running dkg job so we dont trigger a new one. This is the current implementation in the MR.
2. Do we need to clear the state machine after each successful dkg process (meaning rounds 1+2) ?
3. How should we currently make use of the generated public key (for the time being) - store it ?
4. Failure scenarios - what if ... (I have put TODOs in some places for discussions)

Current problem: the btc server takes authority index e.g. 1 and it converts this to frost_peer. The reth node uses the node_peer to convert that to frost_peer. Because of this discrepancy the frost message wont get transferred (I tested it with authority indexes and it was fine, but now we need to decide). We need to see which way we want to go!


bin/btc-server/proto/btc_server.proto
```diff
@@ -17,6 +17,10 @@ service BtcServer {
     rpc GetRound2DkgPackage(Empty) returns (DkgPayload) {}
 
     rpc NewRound2DkgPackage(DkgPayload) returns (Empty) {}
+
+    rpc IsRound2TransitionPossible(Empty) returns (TransitionResponse) {}
+;
+    rpc IsRound3TransitionPossible(Empty) returns (TransitionResponse) {}
```
arminsabouri: Do we need these? The get Pk endpoint will throw if a transition is not possible. Same thing for getting the round2 packages

BatiGencho: I thought of that too, but then I thought we need to parse the errors. I guess you are right, I will change it but we need to set those err strings to consts or something at least to prevent easily changing them.

arminsabouri: Agreed the less endpoints the better. I would prefer just error'ing from the existing endpoing

crates/consensus/authority/src/block_builder.rs
```diff
@@ -38,6 +39,41 @@ where
...
+                        let (sender, receiver) = tokio::sync::oneshot::channel::<DKGState>();
+                        frost_handle.send_command(FrostCommand::GetDkgState(sender));
+                        match receiver.await {
+                            Ok(dkg_state) => {
+                                info!(">>>>>>>>>>> Got a DKG state {:?}", dkg_state);
+                                match dkg_state {
+                                    DKGState::DkgFailed | DKGState::Initial => {
+                                        // start the dkg process / restart ?
+                                        info!(">>>>>>>>>>> Starting the DKG process...");
+                                        frost_handle.send_command(FrostCommand::NewDkgRoundStart);
```
arminsabouri: Should there be some check here that we already have done DKG? or is that handled by the calling function?

crates/net/network/src/frost/dkg.rs
```diff
@@ -0,0 +1,236 @@
...
+    Round2Waiting,
+    Round3,
+    DkgFailed,
+}
+
+#[derive(Debug, Clone)]
+pub struct DKGStateMachine {
+    state: DKGState,
+    personal_identifier: frost::Identifier,
+    public_key_package: Option<String>,
```
arminsabouri: what is the string here? If its a agg public key it should be secp Public key

crates/net/network/src/frost/dkg.rs
```diff
@@ -0,0 +1,236 @@
...
+
+    pub fn reset(self) -> Self {
+        Self {
+            state: DKGState::Initial,
+            personal_identifier: self.personal_identifier,
+            public_key_package: None,
+        }
+    }
+
+    pub fn get_publick_key_package(&self) -> Option<String> {
```
arminsabouri: Typo

```suggestion
    pub fn get_public_key_package(&self) -> Option<String> {
```

crates/net/network/src/frost/dkg.rs
```diff
@@ -0,0 +1,236 @@
...
+                            e.message() == "Missing round2 dkg package"
+                        {
+                            // TODO: re-trigger create round 2 package ?
+                            return;
+                        } else {
+                            return;
+                        }
+                    }
+                };
+                println!("round2_packages: {:?}", dkg2_package);
```
arminsabouri: Nit: remove this

crates/net/network/src/frost/dkg.rs
```diff
@@ -0,0 +1,236 @@
...
+                self.public_key_package = Some(round3_package.publickey);
+            }
+
+            _ => error!("Invalid dkg transition"),
+        }
+    }
+}
+
+pub fn peer_id_to_identifier(peer_id: &PeerId) -> frost::Identifier {
+    frost::Identifier::derive(peer_id.as_slice()).expect("can derive identifier")
```
arminsabouri: Frost id's should be dervied from the the signers index into the federation list

crates/net/network/src/frost/messages.rs
```diff
@@ -0,0 +1,200 @@
+#![allow(unreachable_pub)]
+use reth_eth_wire::{capability::Capability, protocol::Protocol};
+use reth_primitives::{Buf, BufMut, BytesMut};
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct Request {
+    pub identifier: Vec<u8>,
```
arminsabouri, line 16: What are your thoughts on including a version as part of the request/response payload?
For future backward compatible upgradability

crates/net/network/src/frost/messages.rs
```diff
@@ -0,0 +1,200 @@
...
+        }
+    }
+
+    /// Creates a new `TestProtoMessage` with the given message ID and payload.
+    pub fn encoded(&self) -> BytesMut {
+        let mut buf = BytesMut::new();
+        buf.put_u8(self.message_type as u8);
+        match &self.message {
+            FrostProtoMessageKind::Round1Dkg(resource) => {
+                buf.put_u8(resource.identifier.len() as u8); // Assuming identifier is not too long
```
arminsabouri, line 130: Id's should be u16

crates/net/network/src/frost/mod.rs
```diff
@@ -0,0 +1,83 @@
...
+    DkgRound1,
+    DkgRound2,
+    DkgRound3,
+}
+
+#[derive(Debug)]
+pub enum FrostProtocolEvent {
+    ConnectionEstablished {
+        #[allow(dead_code)]
+        // the connection direction - we connected to them, or they to us
```
arminsabouri: Nit: triple slashes if this is documenting these fields or the linter will complain

arminsabouri (changes requested):

crates/net/network/src/frost/messages.rs
BatiGencho: This is a different identifier - its the protocol message identifier, not the frost one. So far we have a few hence a u8 will suffice

crates/net/network/src/frost/dkg.rs
BatiGencho: hex string was returned by the grpc api. I now changed to use secp pubkey, ty

crates/net/network/src/frost/dkg.rs
BatiGencho: done, ty

crates/net/network/src/frost/dkg.rs
BatiGencho: ty, removed

crates/net/network/src/frost/mod.rs
BatiGencho: Yeah, thanks. I fixed these, there is more cleanup to be done in general.

crates/net/network/src/frost/messages.rs
BatiGencho: added one

arminsabouri:
> Current problem: the btc server takes authority index e.g. 1 and it converts this to frost_peer. The reth node uses the node_peer to convert that to frost_peer. Because of this discrepancy the frost message wont get transferred (I tested it with authority indexes and it was fine, but now we need to decide). We need to see which way we want to go!

I vote for authority index. Must easier to communicate that to the federation members

crates/net/network/src/frost/messages.rs
arminsabouri: Got it!

arminsabouri:
>How should we currently make use of the generated public key (for the time being) - store it ?

We don't need to store this. Its only needed when we are retrieving gateway address via RPC. I would ensure that functionality still works. it  should be the `eth_gatewayAddress` RPC endpoint

crates/consensus/common/src/utils.rs
```diff
@@ -116,6 +120,8 @@ pub enum GetAuthoritiesError {
     FailedToRecoverAuthorityList(
         reth_botanix_lib::extra_data_header::ExtraDataHeaderDeserialzeError,
     ),
```
scottmillner, line 121: nit...spelling
`FailedToRetrieveEopchHeader` > `FailedToRetrieveEpochHeader`

crates/net/network/src/frost/manager.rs
```diff
@@ -0,0 +1,273 @@
...
+    /// Network access.
+    network: NetworkHandle,
+    /// Subscriptions to all network related events.
+    ///
+    /// From which we get all new incoming transaction related messages.
+    from_network: UnboundedReceiverStream<NetworkFrostEvent>,
+    /// Copy of the sender half, so new [`FrostManager`] can be created on demand.
+    command_tx: mpsc::UnboundedSender<FrostCommand>,
+    /// Receiver half of the command channel.
+    command_rx: UnboundedReceiverStream<FrostCommand>,
```
scottmillner, line 41: nit, i think some of this comments need to be updated for frost

crates/net/network/src/frost/manager.rs
```diff
@@ -0,0 +1,273 @@
...
+        &mut self,
+        identifier: Vec<u8>,
+        payload: Vec<u8>,
+    ) -> Result<client::Empty, Status> {
+        let req = client::DkgPayload { identifier, payload };
+        let add_round2_dkg_package_future =
+            self.btc_client.new_round2_dkg_package(tonic::Request::new(req));
+        block_on(add_round2_dkg_package_future).map(|res| res.into_inner())
+    }
+
```
scottmillner: this isn't getting used anywhere?

crates/net/network/src/frost/manager.rs
```diff
@@ -0,0 +1,273 @@
...
+                    error!("Network message channel closed.");
+                    return Poll::Ready(())
+                }
+                Poll::Ready(Some(cmd)) => this.on_command(cmd),
+            };
+        }
+        Poll::Pending
+    }
+}
+
```
scottmillner, line 196: nit, `PeersManager` > `FrostManager`

crates/net/network/src/frost/manager.rs
```diff
@@ -0,0 +1,273 @@
...
+    /// Command for manually add
+    NewDkgRoundStart,
+    /// Check if connection to all federated peers is established
+    CheckConnectedToAll(oneshot::Sender<bool>),
+    /// Get the readily generated public key
+    GetFrostPublicKey(oneshot::Sender<Option<secp256k1::PublicKey>>),
+    /// Get current state machine status
+    GetDkgState(oneshot::Sender<DKGState>),
+}
+
```
scottmillner, line 209: same, FrostManager

crates/net/network/src/frost/mod.rs
```diff
@@ -0,0 +1,90 @@
...
+        /// the tx sender we send to the other peer to enable it to communicate with us
+        to_connection: mpsc::UnboundedSender<Command>,
+    },
+    /// An emitted event once a peer sends a message to another peer
+    PeerMessage(Response),
+}
+
+/// All events related to frost events emitted by the network.
+#[derive(Debug)]
+pub enum NetworkFrostEvent {
```
scottmillner, line 77: nit, update comment for frost

scottmillner:
just wanted to say awesome work! this was a huge task to tackle!!

crates/net/network/src/frost/dkg.rs
BatiGencho: Yes, I have changed that. Getting the authority over to that point was  quite difficult.

crates/consensus/common/src/utils.rs
BatiGencho: thanks, changed

crates/net/network/src/frost/manager.rs
BatiGencho: fixed, thanks.

crates/net/network/src/frost/manager.rs
BatiGencho: Not yet since we dont have that part yet (see armins comment above), but soon we will be fetching the pub key using the handle and calling this method.  Its only needed when we are retrieving gateway address via RPC

crates/net/network/src/frost/mod.rs
BatiGencho: updated, thanks.

bin/btc-server/proto/btc_server.proto
BatiGencho: I forgot to commit that part, now it should be all there.

crates/net/network/src/frost/manager.rs
BatiGencho: done

crates/consensus/authority/src/block_builder.rs
BatiGencho: This is the part where I am not qutie sure of, do we do it every new poa epoch (whoever the current in turn fed member is) or not ? Current implementation is: whoever is in turn will perform it once in the poa epoch. The calls are so designed that he would be quering the dkg core until its either successful or it fails.

bin/btc-server/src/main.rs
```diff
@@ -708,6 +713,12 @@ mod test {
             c2.get_public_key(tonic::Request::new(client::Empty {})).await.unwrap().into_inner();
 
         assert_eq!(pk_1.publickey, pk_2.publickey);
+
+        // check public key decoding
+        let decoded_pub_key = hex::decode(pk_1.publickey.clone()).unwrap();
+        let pub_key = secp256k1::PublicKey::from_slice(&decoded_pub_key).unwrap();
+        assert_eq!(pk_1.publickey, pub_key.to_string());
```
stevenroose, line 742: hint: `PublicKey::from_str` exists, doing exactly this ;)

crates/consensus/authority/src/dkg.rs
```diff
@@ -0,0 +1,398 @@
...
+            Err(e) => match e.code() {
+                tonic::Code::AlreadyExists if e.message().contains("already have key package") => {
+                    return Err(Error::KeyPackageAlreadyExists)
+                }
+                tonic::Code::Internal
+                    if e.message().contains("Failed to serialize round 1 dkg") =>
+                {
+                    return Err(Error::Round1PackageSerialize)
+                }
+                tonic::Code::Internal if e.message().contains("Missing round1 dkg package") => {
```
stevenroose, line 187: Uff, doing string matching is a really bad idea...

tonic's Status has 2 things that might help here: "details", which is a freeform binary field, and "metadata" which is a key-value map similar to HTTP headers. I never used them before, but it seems like we can just use them. So we could encode a 32-bit integer error code into the details if we wanted to. Using traits on `Status`, it's possible to create a nice API for them. Can make an enum with codes.

But before doing all that work, I guess it makes sense to ask how valuable it is to have that information at the client side..

crates/consensus/authority/src/dkg.rs
```diff
@@ -0,0 +1,398 @@
...
+                            data: dkg1_package.payload.clone(),
+                        };
+                        let _ = sender.send(FrostPeerCommand::PeerMessage(resp)); // TODO: handle error
+                        println!(">>>>>>>>>>> Command::Round1Dkg send!");
+                    }
+                });
+                std::thread::sleep(std::time::Duration::from_secs(1));
+                // Once the round 1 package is sent we are waiting
+                self.state = DKGState::Round1Waiting;
+                self.next_state(DKGEvent::Round1, btc_server_client, frost_peers_connections).await;
```
stevenroose: Bit of a strange pattern to have a recursive function that sets a new state and then calls the next_state function again. Did you consider using a transformation method that takes a state and returns a new state, then you can have a loop that calls the method repeatedly until a certain success or error state is reached?

crates/consensus/authority/src/frost_task.rs
```diff
@@ -0,0 +1,187 @@
...
+}
+
+impl<Client> std::fmt::Debug for FrostTask<Client>
+where
+    Client: BlockReaderIdExt
+        + StateProviderFactory
+        + CanonChainTracker
+        + BlockchainTreeEngine
+        + Clone
+        + 'static,
```
stevenroose: nit: I don't think you need all these generic constraints here.

crates/consensus/authority/src/utils.rs
```diff
@@ -232,13 +232,13 @@ fn bloom_contains_minting_contract_address(bloom: Bloom) -> bool {
...
-        && bloom.contains_input(BloomInput::Raw(BURN_TOPIC.as_ref()))
+    bloom_contains_minting_contract_address(bloom) &&
+        bloom.contains_input(BloomInput::Raw(BURN_TOPIC.as_ref()))
 }
 
 pub(crate) fn bloom_contains_pegin(bloom: Bloom) -> bool {
-    bloom_contains_minting_contract_address(bloom)
-        && bloom.contains_input(BloomInput::Raw(MINT_TOPIC.as_ref()))
+    bloom_contains_minting_contract_address(bloom) &&
+        bloom.contains_input(BloomInput::Raw(MINT_TOPIC.as_ref()))
```
stevenroose: Are these intentional formatting changes?

crates/net/network/src/frost/manager.rs
```diff
@@ -0,0 +1,201 @@
...
+    /// Handles a command received from a detached [`FrostHandle`]
+    fn on_command(&mut self, cmd: FrostCommand) {
+        match cmd {
+            FrostCommand::CheckConnectedToAll(tx) => {
+                // reply to caller
+                let _ = tx.send(self.all_peers_connected()).ok();
+            }
+            FrostCommand::GetAllConnectedFrostPeers(tx) => {
+                // reply to caller
+                let _ = tx.send(self.frost_peers_connections.clone()).ok();
```
stevenroose: nit: There `.ok()`'s are not having any need. If you want to ignore the error, you can just do `let _ = ` with the result without the `.ok()`.

crates/net/network/src/frost/manager.rs
```diff
@@ -0,0 +1,201 @@
...
+    ),
+}
+
+/// Config type for initiating a [`FrostManager`] instance.
+#[derive(Clone, Debug)]
+pub struct FrostConfig {
+    pub authority_index: usize,
+    pub total_authorities: usize,
+    pub min_signers: u16,
+    pub max_signers: u16,
```
stevenroose, line 220: Could these be documented? :pray:

crates/node-core/src/node_config.rs
```diff
@@ -619,16 +623,26 @@ impl NodeConfig {
...
+
+        let (handle, network, txpool, eth, frost) = network_builder.split_with_handle();
 
         task_executor.spawn_critical("p2p txpool", txpool);
         task_executor.spawn_critical("p2p eth request handler", eth);
+        let mut frost_handle: Option<FrostHandle> = None;
+        if let Some(frost) = frost {
+            frost_handle = Some(frost.handle());
+            task_executor.spawn_critical("p2p frost", frost);
+        }
```
stevenroose: nit: just a hint to avoid having a unnecessarily mutable variable down the line:
```rust
let frost_handle = frost.map(|frost| {
    let ret = front.handle();
    task_executor.spawn_critical("p2p frost", frost);
    ret
});
```

stevenroose (commented):
Definitely not knowledgable enough on FROST matters to review the FROST stuff, but left some general comments on Rusty-ness.

crates/consensus/authority/src/dkg.rs
BatiGencho: I agree with all you are saying, but the errors are loosely typed as strings. I guess for better handling we should introduce codes in the metadata or something. For now its irrelevant as I dont really distinguish between error A or B, just treat the error as a whole. I will put a todo to refactor this part in the next round, ty.

crates/consensus/authority/src/frost_task.rs
BatiGencho: agree, just copy-paste. Removed them

crates/consensus/authority/src/utils.rs
BatiGencho: yes, cargo fmt

crates/net/network/src/frost/manager.rs
BatiGencho: agreed, missed that when coding :)

crates/net/network/src/frost/manager.rs
BatiGencho: I have added docs for everything now.

crates/node-core/src/node_config.rs
BatiGencho: good refactor, ty. I have adopted it.

crates/consensus/authority/src/dkg.rs
BatiGencho: I have reworked this part anyway as it wasnt really suiting our new critical FrostTask handling the dkg state machine logic. You are indeed right.

bin/btc-server/src/main.rs
BatiGencho: Yes, did not see this method:) Ty.

crates/consensus/authority/Cargo.toml
```diff
@@ -45,8 +45,11 @@ reth-btc-wallet.workspace = true
 btc-server = { path = "../../../bin/btc-server" }
 client = {path = "../../../bin/btc-server/client"}
 
-## Ethereum
-ethers = "2.0.11"
+# hex
+hex = "0.4.3"
+
+# frost
+frost-secp256k1-tr = { git = "https://github.com/mimoo/frost", branch = "mimoo/fix5",  features = ["serde", "serialization"] }
```
arminsabouri: can you leave a reminder to change this once tr gets merged

crates/consensus/authority/src/dkg.rs
arminsabouri: We could do error code number that could help

crates/consensus/authority/src/dkg.rs
```diff
@@ -0,0 +1,494 @@
...
+                {
+                    return Err(Error::FailedToAddRound2Package)
+                }
+                _ => return Err(Error::InternalGrpc),
+            },
+        };
+        Ok(())
+    }
+
+    pub(crate) async fn start(
```
arminsabouri: How does DKG get initiated?

crates/consensus/authority/src/frost_task.rs
```diff
@@ -0,0 +1,158 @@
...
+        let (sender, receiver) = tokio::sync::oneshot::channel::<bool>();
+        self.frost_handle.send_command(FrostCommand::CheckConnectedToAll(sender));
+        match receiver.await {
+            Ok(is_connected) => {
+                if !is_connected {
+                    info!("Not yet connected to all frost peers. Waiting ....");
+                    return;
+                }
+                info!(">>>>>>>>>>> Connected to all frost peers {}", is_connected);
+                // start the dkg process / restart ?
```
arminsabouri, line 68: Maybe we can get_public_key and use that as an indication to start dkg or not?

crates/net/network/src/manager.rs
```diff
@@ -78,6 +84,10 @@ pub struct NetworkManager<C> {
     handle: NetworkHandle,
     /// Receiver half of the command channel set up between this type and the [`NetworkHandle`]
     from_handle_rx: UnboundedReceiverStream<NetworkHandleMessage>,
+    /// Receiver half of the command channel set up between this type and the [`NetworkHandle`]
+    from_frost_protol_events: UnboundedReceiverStream<FrostProtocolEvent>,
```
arminsabouri: typo here 
```suggestion
    from_frost_protocol_events: UnboundedReceiverStream<FrostProtocolEvent>,
```

arminsabouri (commented):
Just had a few questions

crates/consensus/authority/src/dkg.rs
scottmillner: looks like `poa_builder.launch()` spawns the `frost_task` which has a `loop` that checks if `dkg` is in the `running` state and if not, it starts `dkg`

wrote this comment more for me just to make sure i understand how it starts

scottmillner (approved):
@BatiGencho 
approving pr to signal that everything has been addressed that's within my knowledge

im not a frost expert so leave that focus to @0xBEEFCAF3 and certainly wouldn't merge until his approval

crates/consensus/authority/src/dkg.rs
BatiGencho: The DKG is started inside of the frost task which checks its has not started it yet and we are in turn. In other words once in turn the dkg is initiated until it completes.

crates/net/network/src/manager.rs
BatiGencho: thanks, fixed.

crates/consensus/authority/src/frost_task.rs
BatiGencho: I have added this condition now. Now we are onlyu starting the dkg process if:

- we are in turn 
- we get an error calling the get_public_key grpc
- we are in DKGState::Initial state

crates/consensus/authority/Cargo.toml
BatiGencho: added

crates/consensus/authority/src/frost_task.rs
arminsabouri: Perfect that should work fine. If DKG is already finished the btc_server signers will just reject any dkg requests

bin/reth/src/poa_builder.rs
```diff
@@ -288,6 +290,20 @@ impl<DB: Database + DatabaseMetrics + DatabaseMetadata + 'static> NodeBuilderWit
...
+        // create authority config
+        let (authority_index, total_authorities) = get_authority_signer_index(
+            blockchain_db.clone(),
+            Arc::clone(&self.config.chain),
+            secp256k1::Secp256k1::new(),
+            network_sk.clone(),
+        )
+        .expect("Failed to get authority index");
+
+        let frost_config = FrostConfig::new(authority_index, total_authorities, 2, 2);
```
arminsabouri, line 327: Can we leave a todo to make max/min singer configuarable

crates/consensus/authority/src/dkg.rs
```diff
@@ -0,0 +1,494 @@
...
+use reth_network::frost::{
+    manager::{FrostCommand, FrostHandle},
+    EventResponseType, FrostPeerCommand, Response,
+};
+use std::{collections::HashMap, str::FromStr};
+use tokio::sync::mpsc::UnboundedSender;
+use tracing::{error, info};
+
+#[derive(Debug, thiserror::Error)]
+pub(crate) enum Error {
```
arminsabouri, line 15: Just a heads up. Im refactoring the errors in BTC to have a bit more structure. No changes needed at the moment. but I thought I let you know

Root Error 
----DKGError
-------Round1Error
----------KeyPackageAlreadyExists
----SigningError
...
----CordinatorError
...

crates/consensus/authority/src/dkg.rs
```diff
@@ -0,0 +1,494 @@
...
+
+        info!(">>>>>>>>>>> Checking for transition to round 2");
+        let round1_payload = match self.get_round1_dkg_package().await {
+            Ok(round1_payload) => round1_payload,
+            Err(e) => {
+                error!("Error getting round 1 dkg package {:?}", e);
+                return Err(e);
+            } // TODO: do what ?
+        };
+        let round1_group_packages: HashMap<frost::Identifier, frost::keys::dkg::round2::Package> =
```
arminsabouri: Maybe we re-name this to round2_group_packages

crates/consensus/authority/src/dkg.rs
```diff
@@ -0,0 +1,494 @@
...
+            {
+                Ok(packages) => packages,
+                Err(e) => {
+                    error!("Error parsing round2 group packages {:?}", e);
+                    return Err(e);
+                } // TODO: do what ?
+            };
+
+        // Check if we are ready to progress to round 3
+        if round2_group_packages.len() >= (self.max_signers - 1) as usize {
```
arminsabouri, line 562: One other way to check this condition is to call get_publickey

crates/consensus/authority/src/dkg.rs
arminsabouri: Or is the type here wrong? Should it be `frost::keys::dkg::round1::Package`

crates/consensus/authority/src/dkg.rs
arminsabouri: And if you get a pk we can put the state machine in a success state

arminsabouri (commented):
Just a few more questions

bin/reth/src/poa_builder.rs
BatiGencho: added

crates/consensus/authority/src/dkg.rs
BatiGencho: already fixed that, thanks

crates/consensus/authority/src/dkg.rs
BatiGencho: thanks, noted

crates/consensus/authority/src/dkg.rs
BatiGencho: Should we rather tackle this in the next MR where we do the signing ? The current one is already massive, maybe I could do a new one and add it in ?

arminsabouri:
@BatiGencho I ran some manual tests. My set up was two federation nodes. I seems like we never get past round 1 on the node where the federation index is 1. 
The first fed memeber successfully stores the round 1 message. however the second one seems to acknowlege it but never stores it 

```
2024-03-05T20:34:30.202357Z  INFO >>>>>>>>>>> [FROST_TASK] Peer messaged received Response { response_type: DkgRound1, identifier: [52, 39, 160, 118, 176, 185, 222, 25, 54, 213, 2, 171, 233, 2, 75, 75, 154, 59, 199, 10, 16, 208, 24, 249, 238, 56, 171, 146, 37, 245, 114, 93], data: [0, 35, 15, 138, 179, 2, 2, 5, 55, 243, 226, 160, 169, 71, 151, 232, 171, 245, 3, 212, 46, 92, 76, 58, 212, 164, 250, 191, 212, 118, 6, 140, 72, 85, 72, 134, 193, 67, 63, 2, 247, 177, 226, 44, 140, 100, 234, 125, 225, 35, 121, 224, 134, 18, 191, 1, 128, 168, 163, 167, 60, 243, 171, 19, 219, 54, 243, 8, 132, 69, 238, 20, 65, 3, 168, 155, 251, 7, 80, 254, 78, 199, 72, 224, 188, 11, 243, 146, 64, 164, 53, 188, 185, 68, 127, 178, 179, 60, 211, 195, 215, 145, 12, 6, 193, 190, 121, 219, 94, 167, 31, 201, 229, 163, 65, 74, 98, 170, 162, 195, 28, 23, 55, 7, 125, 185, 190, 236, 208, 1, 231, 55, 145, 87, 232, 72, 113, 165] }
2024-03-05T20:34:30.202458Z  INFO >>>>>>>>>>> [PROCESS_ROUND1] identifiers Identifier("3427a076b0b9de1936d502abe9024b4b9a3bc70a10d018f9ee38ab9225f5725d") Identifier("3427a076b0b9de1936d502abe9024b4b9a3bc70a10d018f9ee38ab9225f5725d")
2024-03-05T20:34:30.202869Z  INFO >>>>>>>>>>> [FROST_TASK] Processed Round 1 package successfully
2024-03-05T20:34:30.453969Z  INFO Epoch manager inturn: true
2024-03-05T20:34:30.454004Z  INFO >>>>>>>>>>> [FROST_TASK] Peer messaged received Response { response_type: DkgRound2,
...
```


Im running the second btc server with these flags
`cargo run -- --network testnet --db "./db2" --identifier 2 --address 0.0.0.0:8081`

arminsabouri:
Problem was frost identifiers via the cli flags start indexing at 0
tACK [1718a35](https://github.com/botanix-labs/Macbeth/pull/98/commits/1718a3595cb1c707fe690bd41acaf39d9a3b3562)

arminsabouri (approved):

PR #100, opened 2024-02-16
docs: Add document on deterministic coin selection

stevenroose:


docs/botanix/btc-wallet.md
```diff
@@ -0,0 +1,135 @@
...
+## Parameters
+
+The algorithm will take some hard-coded parameters that all validators have in
+common as long as they run software that is aware of the same version of the
+protocol. These parameters can be changed by bumping the protocol version
+number and upgrading a majority of validators to the new version.
+
+The parameters we will use:
+
+- a maximum number of outputs per transaction
```
arminsabouri, line 54: Why would we put a limit on this?

docs/botanix/btc-wallet.md
```diff
@@ -0,0 +1,135 @@
...
+    break the tie
+- take at most a number of pegouts equal to one less than the maximum number of
+  outputs from our parameter
+  - leaving space for a change outputs
+- take all UTXOs owned by the validators and order them by their Bitcoin
+  confirmation height
+  - from most confirmations to less
+  - ex aequos can be resolved by ordering lexicographically on outpoint
+- take UTXOs one by one until either
+  - the total input value if sufficient to cover the total output value, plus
```
arminsabouri:
```suggestion
  - the total input value is sufficient to cover the total output value, plus
```

docs/botanix/btc-wallet.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Deprecating Multisigs
+
+As validators come and go, some multisigs might live longer than others. When a
+high enough number of users from a certain multisig leave the system, the
+multisig will have to be deprecated and the money tied up inside it will have
+to be moved into another multisig.
+
+For this to work, a list of "dying" mulsitigs can be added to the state of the
+Bitcoin wallet.
```
arminsabouri, line 113: This a tough problem. In a spiderchain you need to ensure that the funds get moved to a lively multisig set before they leave.

docs/botanix/btc-wallet.md
```diff
@@ -0,0 +1,135 @@
...
+    UTXOs will have a similar satisfaction weight, a limit on the number might
+    be simpler to reason about)
+
+
+## The Algorithm
+
+Given the input from the above section and the parameters mention, the algorithm
+works as follows:
+
+- order all pegout requests by the Botanix block height they were made in
```
arminsabouri, line 66: Does the order of the pegout requests have significance? I suppose its linked to the max outputs listed above?

docs/botanix/btc-wallet.md
```diff
@@ -0,0 +1,135 @@
...
+  - ordered from more confirmations to less
+  - in case of ex aequo, ordering lexicographically by txid can be used to
+    break the tie
+- take at most a number of pegouts equal to one less than the maximum number of
+  outputs from our parameter
+  - leaving space for a change outputs
```
arminsabouri, line 72: I noticed outputs its plural? Would this algo produce multiple change outputs?

arminsabouri (commented):
Looks great! Just a few questions

docs/botanix/btc-wallet.md
stevenroose: There needs to be some tx size limit. In some way. Oversized txs can have trouble getting confirmed. A malicious user, or a benign situation, could cause us to make a really big tx that then fails to confirm and we're stuck with it.

docs/botanix/btc-wallet.md
stevenroose: fixed

docs/botanix/btc-wallet.md
stevenroose: Yeah, so there must be some kind of "trigger" threshold well above the actual consensus threshold. Like 80-90%.

docs/botanix/btc-wallet.md
stevenroose: Not really, it just needs to be deterministic. But also, since you might not be able to finish all of them in one go, it's good to give the older ones priority.

docs/botanix/btc-wallet.md
stevenroose: Oh, I don't think I wrote that on purpose, but it's a good question. Liquid functionaries often make multiple change outputs because we only spend btc utxos that are finalized, so if we have few utxos and spend all, we can't do anything for 100 blocks. But generally, I see two reasons why there might be 2 change outputs:

- when a really big amount of money comes in, you might want to spread it over multiple multisigs, maybe? Maybe some kind of threshold value per multisig etc. But this is definitely not included in the current scope.
- it's really annoying to spend unconfirmed change, so a multisig would be good to maintain a certain number of confirmed utxos to spend with in next rounds; arguably, I didn't really think about that in the current scope

arminsabouri: Yeah I didn't even think of spending UTXOs that are not "finalized" 
What is the appropriate number of blocks before a UTXO is finalized?

stevenroose: Depends. I would guess reasonably it would be 2 or 3, I guess. It all depends on how solid your logic can handle reorgs or so. If you have good logic, spending unconfirmed _could_ work, but it'd definitely be a lot harder IMO.

arminsabouri (approved):

PR #102, opened 2024-02-16
add function to return all pegouts in an epoch

scottmillner:


scottmillner:
this was already approved but i accidently closed it

PR #103, opened 2024-02-22
deduct tx costs from pegin amount and refund the receiver address

scottmillner:
~~Only review first commit since second commit is `cargo fmt` and has a bunch of files impacted~~


contracts/src/Minting.sol
```diff
@@ -23,17 +23,32 @@ contract Minting {
...
+        uint256 gasStart = gasleft();
+
         // Check that the user bitcoin block height is increasing.
         require(
             bitcoinBlockHeight > peginBitcoinBlockHeight[destination],
             "user bitcoinBlockHeight needs to increase"
         );
         peginBitcoinBlockHeight[destination] = bitcoinBlockHeight;
 
-        payable(destination).transfer(amount);
```
scottmillner, line 35: if the tx reverts, the Mint event will still show in logs since it occurs before the revert
but our validation logic skips receipts that don't have a success status

since the Mint event gas cost will vary depending on the size of the proofs in the metadata, we need to run it before calculating gas costs so we can be accurate

contracts/src/Minting.sol
```diff
@@ -23,17 +23,32 @@ contract Minting {
...
+        // emit here so gas is included in gasUsed
         emit Mint(destination, amount, bitcoinBlockHeight, metadata);
+
+        // account for gas needed for the transfers, amount update, and require statement if true
+        uint256 txCost = (gasStart - gasleft() + 21000 + 21000 + 2003 + 3) * tx.gasprice;
```
BatiGencho: nit: could we maybe declare all of these gas values as CONSTANTS in the contract and give them a meaningful name e.g.:

const GAS_TRANSFER  = 21000;
const GAS_AMOUNT_UPDATE = ...

scottmillner: yep thanks, i was trying to save gas but forget constants don't consume gas
done

contracts/src/Minting.sol
```diff
@@ -3,6 +3,9 @@ pragma solidity ^0.8.13;
 
 contract Minting {
     uint public constant SATS_TO_WEI = 10**10;
+    uint constant GAS_SIMPLE_TRANSFER = 21000;
+    uint constant GAS_AMOUNT_UPDATE = 2003;
+    uint constant GAS_REVERT_TRUE = 3;
```
arminsabouri, line 8: What is `GAS_REVERT_TRUE `? Is there a more descriptive name for this?

arminsabouri (commented):
@scottmillner Do we also need update mint validation?

contracts/src/Minting.sol
scottmillner: revert statements have different gas costs based on if theyre true or false
if it's false, it doesn't matter since the tx will revert

im open to suggestions for the name

scottmillner:
> @scottmillner Do we also need update mint validation?



> @scottmillner Do we also need update mint validation?

yep good catch
@0xBEEFCAF3 updated mint validation logic to extract the final pegin amount from the new MintAmount event and that gets sent to btc server

contracts/src/Minting.sol
arminsabouri: Ah ok understood. Thanks

crates/botanix-lib/src/mint_validation.rs
```diff
@@ -88,11 +105,37 @@ pub fn parse_pegout_reth_log_topic(
...
+            // first topic is the event signature, second topic is the indexed amount
+            if log.topics().len() != 2 {
+                return Err(MintConsensusError::UnexpectedLog("wrong number of topics"));
+            }
+
+            let amount = topic_to_amount(log.topics()[1])?;
+            return Ok(amount);
+        }
+    }
+    Err(MintConsensusError::MintContractDidNotEmitMintTopic())
```
arminsabouri, line 120: Are there any situations where a mint amount event isn't (or maybe shouldnt) be emited. I'm thinking about the case where there is no refunder.

contracts/src/Minting.sol
```diff
@@ -23,17 +29,37 @@ contract Minting {
...
             "user bitcoinBlockHeight needs to increase"
         );
         peginBitcoinBlockHeight[destination] = bitcoinBlockHeight;
 
-        payable(destination).transfer(amount);
+        // emit here so gas is included in gasUsed
         emit Mint(destination, amount, bitcoinBlockHeight, metadata);
+
+        // account for gas needed for the transfers, amount update, and require statement if true
+        uint256 txCost = (gasStart - gasleft() + GAS_SIMPLE_TRANSFER + GAS_SIMPLE_TRANSFER + GAS_AMOUNT_UPDATE + GAS_REVERT_TRUE + GAS_MINT_AMOUNT_EVENT ) * tx.gasprice;
```
arminsabouri, line 48: should the code from here and below be wrapped in a if? Refund address should be optional

arminsabouri (commented):
Looks great! just had one or two questions

arminsabouri:
@scottmillner Looks like the fmt is also failing on this job

crates/botanix-lib/src/mint_validation.rs
scottmillner: the mint amount event is always emitted whether the tx fails or not

the `mint()` method is always expecting a refunder (solidity doesn't have optional arguments) so the tx will be rejected from the mempool bc not enough args if no refunder

scottmillner: if we want to cover all the fees, we can just pass the user's address as the refund address so they get the full pegin amount

contracts/src/Minting.sol
scottmillner: see comment above: solidity doesn't have optional arguments
so we'd just pass the user address so they get the refund (net effect is user gets full amount) if we wanna cover the costs

scottmillner: putting here for transparency

@armin and i had slack discussion and he brought up a good point:
- we could refactor Minting.sol to check if refund address is `0x` which would be the equivalent of making the argument optional
- then we could have an `if` check that would short circuit all the refund cost code
- this would reduce how much the broadcaster would pay if they pass `0x` since that's the broadcaster's way of covering all the pegin costs
- without this refactor, the broadcaster will pay more but since we're setting the base fee so low, the difference is probably minimal
- so this is a trade off of a simpler leaner `mint()` method and less `mint_validation` logic v cheaper tx costs when pegin is being subsidized by broadcaster

scottmillner:
> @scottmillner Looks like the fmt is also failing on this job

second commit is cargo fmt for entire repo and is huge

arminsabouri:
> > @scottmillner Looks like the fmt is also failing on this job
> 
> second commit is cargo fmt for entire repo and is huge

ugh. Fmt is still failing. We must have different fmt rules than the upstream team. We should solve this in a seperate PR. Do you mind dropping the second commit. Sorry for the back and forth

scottmillner:
> > > @scottmillner Looks like the fmt is also failing on this job
> > 
> > 
> > second commit is cargo fmt for entire repo and is huge
> 
> ugh. Fmt is still failing. We must have different fmt rules than the upstream team. We should solve this in a seperate PR. Do you mind dropping the second commit. Sorry for the back and forth

all good...second commit dropped

contracts/test/Minting.t.sol
```diff
@@ -26,7 +28,7 @@ contract MintingTest is Test {
     function testFundContract() public payable {
         // mock metadata
         bytes memory metadata = bytes("0x00000000");
-        minting.mint(destination, amount, bitcoinBlockHeight, metadata);
+        minting.mint(destination, amount, bitcoinBlockHeight, metadata, refundAddress);
```
arminsabouri, line 40: Is there a way to write a test here to ensure `MintAmount ` is being emmitted with the correct values?

crates/revm/src/processor.rs
```diff
@@ -262,16 +262,17 @@ where
     ) -> Result<(), BlockExecutionError> {
         for log in result.logs() {
             if log.topics().get(0) == Some(&MINT_TOPIC) && recent_block_header.is_some() {
-                let pegin_data = parse_pegin_topic(&log).map_err(|e| {
+                let pegin_data = parse_pegin_topic(&log, &result.logs()).map_err(|e| {
```
arminsabouri, line 257: Should this be `parse_pegin_mint_amount_topic`?

arminsabouri: Ah nvm I see that it gets called by `parse_pegin_topic`

contracts/test/Minting.t.sol
scottmillner: yes sir, it is done!

arminsabouri (approved):
LGTM! Maybe @BatiGencho can provide a second set of eyes

crates/botanix-lib/src/mint_validation.rs
```diff
@@ -88,11 +105,37 @@ pub fn parse_pegout_reth_log_topic(
...
+
+pub fn parse_pegin_topic(log: &revm::primitives::Log, logs: &Vec<revm::primitives::Log>) -> Result<PeginData, MintConsensusError> {
     if log.address != *MINT_CONTRACT_ADDRESS {
         return Err(MintConsensusError::MintContractDidNotEmitMintTopic());
     }
 
+    // get mint amount from MintAmount event
+    let mut amount: ethers::types::U256 = ethers::types::U256::zero();
+    for log in logs {
+        match parse_pegin_mint_amount_topic(log) {
```
BatiGencho: maybe we can do here:

if let Ok(..) {  ... } else { continue }

arminsabouri:
There is a sidecar dependency here. 
@scottmillner to set up a v1 branch with the changes we need

arminsabouri:
v1 branch merge here https://github.com/botanix-labs/Side-Car/pull/87

crates/botanix-lib/src/mint_validation.rs
scottmillner: ah yes, thanks!!
done!

PR #104, opened 2024-02-22
k8s-manifest addition and workflow

birozuru:
Added GKE manifests for the reth-node and btc-server. Also updated the existing workflow file deployment on pull request merged.

### DIFF
--- a/k8s/reth-node/configmap.yml
+++ b/k8s/reth-node/configmap.yml
@@ -10,7 +10,7 @@ data:
     set -e
 
     # Set the bucket name and file name
-    BUCKET_NAME="botaniz-testnet-snapshot"
+    BUCKET_NAME="botanix-testnet-snapshot"
     FILE_NAME="botanix-testnet.tgz"
 
     # Download the file from the bucket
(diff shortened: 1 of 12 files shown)


scottmillner (approved):

k8s/reth-node/configmap.yml
```diff
@@ -0,0 +1,36 @@
...
+  name: botanix-reth-init-config
+  namespace: sandbox
+data:
+  extract.sh: |
+    #!/bin/bash
+
+    set -e
+
+    # Set the bucket name and file name
+    BUCKET_NAME="botaniz-testnet-snapshot"
```
BatiGencho: is the name correct here , Mark ? botaniz ?

birozuru: Oversight should be botanix

BatiGencho (approved):

k8s/reth-node/configmap.yml
BatiGencho: fixed, thx mark

PR #105, opened 2024-02-26
added bitcoind client & syncing

BatiGencho:
This MR changes the mempool impl to bitcoind introducing a new client and integration for syncing state. There are also clippy fixes and other rusty improvements.

Tickets:

- https://github.com/botanix-labs/botanix/issues/342
- https://github.com/botanix-labs/botanix/issues/343

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -418,7 +418,9 @@ start-poa-server-1:
 	--authrpc.addr "127.0.0.1" \
 	--authrpc.port 8551 \
 	--btc-server "localhost:8080" \
-	--btc-block-source "https://mempool.space/signet/api" \
+	--bitcoind.url "https://bitcoind.botanixlabs.dev" \
+	--bitcoind.username "mempool" \
+	--bitcoind.password "mempool" \
 	--p2p-secret-key "${NODE_1_DIR}/discovery-secret" \
 	--port 30303
 
@@ -437,6 +439,8 @@ start-poa-server-2:
 	--authrpc.addr "127.0.0.1" \
 	--authrpc.port 8552 \
 	--btc-server "localhost:8080" \
-	--btc-block-source "https://mempool.space/signet/api" \
+	--bitcoind.url "https://bitcoind.botanixlabs.dev" \
+	--bitcoind.username "mempool" \
+	--bitcoind.password "mempool" \
 	--p2p-secret-key "${NODE_2_DIR}/discovery-secret" \
 	--port 30304
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
+-include .env
+
 # Heavily inspired by Lighthouse: https://github.com/sigp/lighthouse/blob/693886b94176faa4cb450f024696cb69cda2fe58/Makefile
 .DEFAULT_GOAL := help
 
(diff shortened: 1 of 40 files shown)


crates/consensus/authority/src/block_builder.rs
```diff
@@ -34,6 +34,19 @@ where
         // Check if we are in_turn
         let is_inturn = self.epoch_manager.poll().await;
 
+        match self.bitcoind_client.is_synced().await {
```
arminsabouri: This code looks correct but im not sure whether it belong here or in the main function. i.e is there any point to starting reth until the bitcoind client is sync'd? Maybe we should busy loop until this is the case in main. I'll leave this as a open question. Lmk what you think

arminsabouri (commented):
I just had one open question. Awesome work!

crates/consensus/authority/src/block_builder.rs
BatiGencho: I though of it too, actually wanted to leave a note. It definitely makes sense to await the sync in the start (poa_builder.rs) - I will move the code there, not an issue. Thanks for pointing this out to me.

arminsabouri: Great! after that happy to get this merged

scottmillner: awesome sauce! looks great! will ✅ once sync check is in main function
The Mighty Mutex strikes again!!

BatiGencho:
@0xBEEFCAF3 I have addressed your request. Do you mind going over it again.

Makefile
```diff
@@ -418,7 +418,9 @@ start-poa-server-1:
 	--authrpc.addr "127.0.0.1" \
 	--authrpc.port 8551 \
 	--btc-server "localhost:8080" \
-	--btc-block-source "https://mempool.space/signet/api" \
+	--bitcoind.url "https://bitcoind.botanixlabs.dev" \
+	--bitcoind.username "mempool" \
+	--bitcoind.password "mempool" \
```
arminsabouri: These are the actual creds for the signet rpc right? Can we read these from env var?

crates/btc-wallet/src/bitcoind.rs
```diff
@@ -0,0 +1,161 @@
...
+    TransactionBroadcastFailed(bitcoincore_rpc::Error),
+    #[error("Block index failed")]
+    BlockIndexStatusFailed(bitcoincore_rpc::Error),
+}
+
+#[derive(PartialEq, Eq, Debug, Clone, Serialize, Deserialize)]
+pub struct BitcoindConfig {
+    url: Url,
+    username: Option<String>,
+    password: Option<String>,
```
arminsabouri: Do these have to be optional? Seems like they are required via the expect() below

crates/node-core/src/args/bitcoind_args.rs
```diff
@@ -0,0 +1,41 @@
...
+
+    /// Btcd username
+    ///
+    /// The username of the bitcoind server.
+    #[arg(
+        long = "bitcoind.username",
+        name = "bitcoind.username",
+        value_name = "BITCOIND_USERNAME"
+    )]
+    pub username: Option<String>,
```
arminsabouri: Same thing here. Do the creds have to be optional?

arminsabouri (changes requested):
Everything looks good. Just want to make sure we don't expose any real credentials

crates/node-core/src/args/bitcoind_args.rs
BatiGencho: I wasnt sure if bitcoind always needs creds, but that seems to be the case. I have removed the Options now.

crates/btc-wallet/src/bitcoind.rs
BatiGencho: I wasnt sure if bitcoind always needs creds, but that seems to be the case. I have removed the Options now.

Makefile
BatiGencho: They were, I removed them. Was using them for testing. The vals are now being read from .env file.

arminsabouri (approved):

scottmillner (approved):

PR #106, opened 2024-02-27
Merge latest reth changes from master

BatiGencho:


crates/consensus/authority/src/builder.rs
```diff
@@ -65,7 +65,8 @@ pub enum AuthorityConsensusBuilderError {
 impl<Client, EvmConfig, Engine> AuthorityConsensusBuilder<Client, EvmConfig, Engine>
 where
     Engine: EngineTypes + 'static,
-    EvmConfig: ConfigureEvmEnv + Clone + Unpin + Send + Sync + 'static,
+    EvmConfig:
+        ConfigureEvmEnv + Clone + Unpin + Send + Sync + 'static + reth_node_api::ConfigureEvm,
```
arminsabouri, line 69: Do we need both `ConfigureEvmEnv` and `reth_node_api::ConfigureEvm`

arminsabouri (approved):
Looks good. Reviewed only the POA consensus things. Had one question

arminsabouri:
Also I think its fine to create a merge commit for this PR. Rebasing will be messy as we have learned.

crates/consensus/authority/src/builder.rs
BatiGencho: We do, as they added a new trait which has the remaining functionality in it.

scottmillner (approved):

PR #107, opened 2024-02-27
refactor: reduce base fee by factor of 1000

scottmillner:


arminsabouri (approved):
Thanks for adding this back in

PR #109, opened 2024-03-01
refactor: process pegouts at end of epoch

scottmillner:
Testing:
I have tested this and stepped through the code and verified it's working up to sending the pegout request to `btc_server` which then prints the following error: `"Invalid tx request, dkg has not been completed"`
I have no reason to believe the pegout request will not be successful once DKG is fully implemented

Overview:

- When node spins up, it stores all pegouts in the epoch in `InnerStorage`
- When node is building the last block in the epoch, it gets pegouts from the cache and sends them to `btc_server`
- The node clears all pegouts from the cache once above is done

Open questions:

- What do we do with failed pegouts?
- Current implementation loops through pegouts and just logs the ones that fail and continues processing all pegouts until finished
- The failed pegouts don't carry over into the next epoch and are lost funds for the user

### DIFF
--- a/crates/consensus/authority/src/block_builder.rs
+++ b/crates/consensus/authority/src/block_builder.rs
@@ -1,8 +1,9 @@
 use crate::{
     engine_util::{self, BestTransactionsError},
     task::BlockProductionTask,
-    utils::is_epoch_end,
+    utils::{is_epoch_end, send_pegouts},
 };
+use client::MakeTxRequest;
 use reth_consensus_common::utils;
 use reth_eth_wire::NewBlock;
 use reth_interfaces::blockchain_tree::{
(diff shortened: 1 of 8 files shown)


arminsabouri (commented):
Awesome work!!! Left some questions and comments.

crates/consensus/authority/src/block_builder.rs
```diff
@@ -221,9 +203,25 @@ where
...
+                info!(target: "consensus::authority", "Sending pegouts: {:?}", pegouts);
+                match send_pegouts(&self.bitcoin_block_source, &mut self.btc_server, pegouts).await
+                {
+                    Ok(_) => {}
+                    Err(e) => {
+                        error!(target: "consensus::authority", ?e, "Failed to send all pegouts");
+                    }
+                }
+
+                storage.pegouts.clear();
```
arminsabouri: The problem here is that regarless of error or not the cache gets cleared. If the signing session restarts, signers have no way of validating the pegouts.

crates/consensus/authority/src/utils.rs
```diff
@@ -192,31 +226,20 @@ async fn process_botanix_log(
...
-
-                let raw_tx = response.into_inner().tx;
-                info!(target: "consensus::authority", "broadcasting withdrawal tx");
-
-                bitcoin_block_source
-                    .broadcast_tx(&hex::encode(raw_tx))
-                    .await
-                    .map_err(|_| ProcessBotanixLogError::FailedToBroadcastPegout)?;
+                info!(target: "consensus::authority", "Adding pegout to cache");
+                storage.write().await.pegouts.push(request);
```
arminsabouri: What happens here in the case of a re-org? This is a downside of the cache I think.

crates/consensus/authority/src/block_builder.rs
scottmillner: per our conversation will remove pegouts from cache and just grab all at the end of epoch and process

crates/consensus/authority/src/utils.rs
scottmillner: removing pegouts from cache

scottmillner:
superseded by https://github.com/botanix-labs/Macbeth/pull/112

PR #110, opened 2024-03-04
[DON'T MERGE] refactor: enforce required pegin confirmation depth by network

scottmillner:
tested running sidecar related pr as well and behaving as expected

pr https://github.com/botanix-labs/Macbeth/pull/115 supersedes this one

### DIFF
--- a/crates/consensus/authority/src/block_builder.rs
+++ b/crates/consensus/authority/src/block_builder.rs
@@ -1,7 +1,7 @@
 use crate::{
     engine_util::{self, BestTransactionsError},
     task::BlockProductionTask,
-    utils::is_epoch_end,
+    utils::{get_recent_block_height_or_zero, is_epoch_end, is_testnet},
 };
 use reth_consensus_common::utils;
 use reth_eth_wire::NewBlock;
@@ -11,7 +11,9 @@ use reth_interfaces::blockchain_tree::{
 use reth_node_api::{ConfigureEvmEnv, EngineTypes};
 use reth_node_ethereum::EthEngineTypes;
 use reth_payload_builder::{EthBuiltPayload, EthPayloadBuilderAttributes};
-use reth_primitives::{public_key_to_address, Block, SealedBlockWithSenders, B256};
+use reth_primitives::{
+    public_key_to_address, Block, SealedBlockWithSenders, B256, BOTANIX_TESTNET,
+};
 use reth_provider::{BlockReaderIdExt, CanonChainTracker, StateProviderFactory};
 use reth_rpc_types::engine::PayloadAttributes;
 use ruint::Uint;
@@ -135,7 +137,15 @@ where
             })
             .unzip();
 
-        let recent_bitcoin_block_header = *self.bitcoin_block_header.read().await;
+        let recent_bitcoin_block_header: Option<(bitcoin::block::Header, u32)> =
+            *self.bitcoin_block_header.read().await;
+        let recent_bitcoin_block_height =
+            get_recent_block_height_or_zero(recent_bitcoin_block_header);
+        if recent_bitcoin_block_height == 0 {
+            error!(target: "consensus::authority", "Failed to get recent bitcoin block height");
+            drop(storage);
+            return;
+        }
         let authority_signers = storage.authorities.clone();
 
         // Build and execute current block template
@@ -159,10 +169,13 @@ where
         };
 
         // Process Botanix specific logs
+        let is_testnet = is_testnet(self.chain_spec.chain().id());
         match crate::utils::process_receipts(
             &self.bitcoind_client,
             &mut self.btc_server.clone(),
             &bundle_state,
+            recent_bitcoin_block_height,
+            is_testnet,
             false,
         )
         .await
--- a/crates/consensus/authority/src/block_fetcher.rs
+++ b/crates/consensus/authority/src/block_fetcher.rs
@@ -1,7 +1,10 @@
-use crate::engine_util;
+use crate::{
+    engine_util,
+    utils::{get_recent_block_height_or_zero, is_testnet},
+};
 
 use reth_interfaces::blockchain_tree::BlockchainTreeEngine;
-use reth_primitives::{SealedBlockWithSenders, TransactionSigned};
+use reth_primitives::{SealedBlockWithSenders, TransactionSigned, BOTANIX_TESTNET};
 use reth_provider::{BlockReaderIdExt, CanonChainTracker, Chain, StateProviderFactory};
 
 use crate::Storage;
@@ -111,6 +114,12 @@ where
             };
 
             let recent_bitcoin_block_header = *self.bitcoin_block_header.read().await;
+            let recent_bitcoin_block_height =
+                get_recent_block_height_or_zero(recent_bitcoin_block_header);
+            if recent_bitcoin_block_height == 0 {
+                error!(target: "consensus::authority", "Failed to get recent bitcoin block height");
+                continue;
+            }
             let mut storage = self.storage.write().await;
 
             match storage.execute_imported_block(
@@ -126,10 +135,13 @@ where
                         SealedBlockWithSenders::new(sealed_block.clone(), senders)
                             .expect("senders are valid");
                     // Process Botanix specific logs
+                    let is_testnet = is_testnet(self.chain_spec.chain().id());
                     match crate::utils::process_receipts(
                         &self.bitcoind_client,
                         &mut self.btc_server.clone(),
                         &bundle_state,
+                        recent_bitcoin_block_height,
+                        is_testnet,
                         false,
                     )
                     .await
--- a/crates/consensus/authority/src/utils.rs
+++ b/crates/consensus/authority/src/utils.rs
@@ -1,3 +1,4 @@
+use bitcoin::block::Header;
 use client::{BtcServerClient, MakeTxRequest, NotifyPeginRequest};
 use futures_util::{stream::FuturesUnordered, StreamExt};
 use reth_botanix_lib::mint_validation::{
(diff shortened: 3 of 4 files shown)


crates/consensus/authority/src/block_builder.rs
```diff
@@ -159,10 +167,13 @@ where
         };
 
         // Process Botanix specific logs
+        let is_signet = self.chain_spec.chain().id() == BOTANIX_TESTNET.chain().id();
```
arminsabouri: nit: Do we you mind if we switch the this and other variable names to `testnet` to remain consistent with the other variables? i.e  `is_testnet`

scottmillner: you got it! done!

crates/consensus/authority/src/block_builder.rs
```diff
@@ -136,6 +136,14 @@ where
             .unzip();
 
         let recent_bitcoin_block_header = *self.bitcoin_block_header.read().await;
+        let recent_bitcoin_block_height = match recent_bitcoin_block_header {
```
ozymandias-rex: given this is the same as https://github.com/botanix-labs/Macbeth/pull/110/files#diff-31fa1cff6a450309ffc5d47fb4caababd00c4d0f542a863e29d96e9c284c59eeR114, does it make sense to convert it to a helper in utils and import it into  the block builder and block fetcher

crates/primitives/src/constants/mod.rs
```diff
@@ -192,6 +192,12 @@ pub const STAKING_CONTRACT_ADDRESS: &str = "0xA107e8549eB97f420a1D3b93Bd202e0845
 /// Storage slot index of list of stakers
 pub const STAKER_BALANCE_MAPPING_STORAGE_SLOT_INDEX: u64 = 1u64;
 
+/// Signet required confirmation depth for pegins
+pub const SIGNET_PEGIN_CONFIRMATION_DEPTH: u32 = 1;
+
+/// Mainnet required confirmation depth for pegins
+pub const MAINNET_PEGIN_CONFIRMATION_DEPTH: u32 = 100;
```
ozymandias-rex, line 201: Does this translate to 100 confirmations for mainnet ? How do we arrive as it ? Is there a property of pegins that necessitates the high threshold ?

crates/consensus/authority/src/utils.rs
```diff
@@ -170,6 +189,16 @@ async fn process_botanix_log(
                 info!(target: "consensus::authority", "Parsing and sending minting event to btc_server");
                 let pegin_data = parse_pegin_reth_log_topic(log)
                     .expect("passed evm check should pass this parse attempt");
+                // enforce required confirmation depth by network
+                let confirmation_depth = if is_testnet {
```
ozymandias-rex: nit: using a `match` here might be more idiomatic

```rust
let confirmation_depth = match is_testnet {
    true => SIGNET_PEGIN_CONFIRMATION_DEPTH,
    false => MAINNET_PEGIN_CONFIRMATION_DEPTH,
};
```

crates/consensus/authority/src/utils.rs
```diff
@@ -170,6 +189,16 @@ async fn process_botanix_log(
...
+                    SIGNET_PEGIN_CONFIRMATION_DEPTH
+                } else {
+                    MAINNET_PEGIN_CONFIRMATION_DEPTH
+                };
+                if pegin_data.bitcoin_block_height > recent_bitcoin_block_height - confirmation_depth {
+                    info!(target: "consensus::authority", "pegin confirmation depth not met, skipping");
```
ozymandias-rex: Should this be a `warn!` ?

crates/consensus/authority/src/block_builder.rs
scottmillner: i certainly can
my general rule is if i have to do it 3 times i pull it out into a separate function lol
i shall change the rule to 2 times and convert it to a helper!

implemented

crates/primitives/src/constants/mod.rs
scottmillner: btc protocol doesn't allow reorgs after 100 blocks
 @0xBEEFCAF3 correct?

crates/consensus/authority/src/utils.rs
scottmillner: thanks, implemented!

crates/primitives/src/constants/mod.rs
ozymandias-rex: IIRC , the current code will skip peggins until they have reached 100 confirmations (600 mins). This is alot more than the time that exchanges wait for .

crates/consensus/authority/src/utils.rs
scottmillner: yes it should, implemented, thanks!

crates/primitives/src/constants/mod.rs
scottmillner: @ozymandias-rex 
We're trying to ensure that a reorg is never possible and 100 is the only depth where that is guaranteed.
Currently, a depth of 38 will guarantee that with the current hashrate (see [calculator](https://jlopp.github.io/bitcoin-confirmation-risk-calculator/) but this is a moving target). 100 is future proof unless a protocol change is made.

But agreed, this is a major constraint and may not be practical so something we need to discuss as a team.

arminsabouri: As it turns out there isnt a protocol limit. 100 is just what liquid uses to reduce the probability of a re-org to pretty much 0

crates/consensus/authority/src/utils.rs
```diff
@@ -250,11 +286,31 @@ pub(crate) fn is_epoch_end(current_block_number: u64) -> bool {
...
+///
+/// # Arguments
+///
+/// * `recent_bitcoin_block_header` - The recent bitcoin block header
+///
+/// # Returns
+///
+/// Returns the recent block height or 0 if None.
+pub(crate) fn get_recent_block_height_or_zero(recent_bitcoin_block_header: Option<(Header, u32)>) -> u32 {
+    match recent_bitcoin_block_header {
```
BatiGencho: nit: I believe we could rewrite this in a more rusty way by:
```
recent_bitcoin_block_header.map(|header| header.height).unwrap_or_else(|| {
   error!(target: "consensus::authority", "Failed to get recent bitcoin block height");
            0
})
```

crates/consensus/authority/src/utils.rs
```diff
@@ -170,6 +195,17 @@ async fn process_botanix_log(
                 info!(target: "consensus::authority", "Parsing and sending minting event to btc_server");
                 let pegin_data = parse_pegin_reth_log_topic(log)
                     .expect("passed evm check should pass this parse attempt");
+                // enforce required confirmation depth by network
+                let confirmation_depth = match is_testnet {
+                    true => SIGNET_PEGIN_CONFIRMATION_DEPTH,
+                    false => MAINNET_PEGIN_CONFIRMATION_DEPTH,
+                };
```
BatiGencho: I would personally move this into its own function e.g. 
```
fn get_confirmation_depth(is_testnet) {
 match is_testnet {
                    true => SIGNET_PEGIN_CONFIRMATION_DEPTH,
                    false => MAINNET_PEGIN_CONFIRMATION_DEPTH,
                }
}
```
as we could later also evtl. extend this with regtest network and others ? Maybe not :)

crates/consensus/authority/src/block_fetcher.rs
```diff
@@ -126,10 +131,13 @@ where
                         SealedBlockWithSenders::new(sealed_block.clone(), senders)
                             .expect("senders are valid");
                     // Process Botanix specific logs
+                    let is_testnet = self.chain_spec.chain().id() == BOTANIX_TESTNET.chain().id();
```
BatiGencho: I guess we could also pull this into its own function for later re-usability

e,g.
```
fn is_testnet() -> bool {
self.chain_spec.chain().id() == BOTANIX_TESTNET.chain().id()
}
```

crates/consensus/authority/src/block_builder.rs
```diff
@@ -159,10 +168,13 @@ where
         };
 
         // Process Botanix specific logs
+        let is_testnet = self.chain_spec.chain().id() == BOTANIX_TESTNET.chain().id();
```
BatiGencho: same here, see my comment below reg. reusability

crates/primitives/src/constants/mod.rs
```diff
@@ -192,6 +192,12 @@ pub const STAKING_CONTRACT_ADDRESS: &str = "0xA107e8549eB97f420a1D3b93Bd202e0845
 /// Storage slot index of list of stakers
 pub const STAKER_BALANCE_MAPPING_STORAGE_SLOT_INDEX: u64 = 1u64;
 
+/// Signet required confirmation depth for pegins
+pub const SIGNET_PEGIN_CONFIRMATION_DEPTH: u32 = 1;
```
BatiGencho, line 198: Hmm... is this something that should live among the reth primitives or can we not move it into some of our botanix poa specific rust files ? Ignore if comment is stupid pls

crates/consensus/authority/src/utils.rs
scottmillner: that's better for sure, thanks!
done

crates/consensus/authority/src/utils.rs
scottmillner: you got it!
done!

crates/primitives/src/constants/mod.rs
scottmillner: added a TODO
if we move these we should move all the botanix specific consts in this file(`STAKING_CONTRACT_ADDRESS` and `STAKER_BALANCE_MAPPING_STORAGE_SLOT_INDEX`)

maybe add a new file in this folder (constants)?

crates/consensus/authority/src/block_fetcher.rs
scottmillner: done!

crates/consensus/authority/src/block_builder.rs
scottmillner: done!

crates/primitives/src/constants/mod.rs
BatiGencho: Nah, its fine. Now that I think about it, leave them in there, we have already massively protruded the reth repo with our impls, why not to keep botanix and reth consts in the same place.

BatiGencho (approved):

scottmillner:
implemented in pr https://github.com/botanix-labs/Macbeth/pull/115

PR #113, opened 2024-03-06
refactored readme file with new instructions

BatiGencho:
Updated README.md


arminsabouri (approved):
Good stuff. We'll probably need to revist some of the bitcoind instrucitons when we open source but good to go for now

scottmillner (approved):

PR #114, opened 2024-03-06
Mark/feature/k8s

birozuru:
Updated workflow pipeline


birozuru:
Updated the workflow to use the new self-hosted runner. There are some weird commits from earlier this yeah causing conflicts. @0xBEEFCAF3 @scottmillner could you check through to confirm those aren't  file breaking commits.

arminsabouri (changes requested):
I think you probably branched off main instead of feature/poA-consensus.
There are some commits from main coming into this branch

PR #118, opened 2024-03-07
refactor(btc_server): move most psbt validation to `validate_psbt()`

arminsabouri:
This psbt needs to be validated at each stage. By both the signers and coordinator. Most of this validation is duplicated in each rpc endpoint. This PR refactors it into a single util methods

This might be overkill but I also realized that the signers may modify the original PSBT by adding additional inputs and/or outputs. And the coordinator as of now doesnt check for this. Could be good to check for this


arminsabouri:
Needs rebase. Drafting

scottmillner (approved):
🔥 💪 ⭐

BatiGencho (approved):

PR #120, opened 2024-03-09
Scott/355 feature get block filters from bitcoind and store in memory

scottmillner:
Tested and it's working. Mark updated our bitcoind args to pass `blockfilterindex=1` so filters are being created and can be consumed.

### DIFF
--- a/bin/reth/src/poa_builder.rs
+++ b/bin/reth/src/poa_builder.rs
@@ -12,7 +12,7 @@ use crate::{
 use eyre::Context;
 use fdlimit::raise_fd_limit;
 use futures::{future::Either, stream, stream_select, StreamExt};
-use reth_authority_consensus::{AuthorityConsensusBuilder, utils::{is_testnet, get_confirmation_depth}};
+use reth_authority_consensus::{AuthorityConsensusBuilder, task::BlockFilterData, utils::{is_testnet, get_confirmation_depth}};
 use reth_beacon_consensus::{
     hooks::{EngineHooks, PruneHook},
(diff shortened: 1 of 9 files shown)


bin/reth/src/poa_builder.rs
```diff
@@ -184,6 +186,67 @@ impl<DB: Database + DatabaseMetrics + DatabaseMetadata + 'static> NodeBuilderWit
             }
         }
```
scottmillner, line 188: decided to create a separate task instead of including it in the `bitcoin block header task` so there is some duplicated code

thought separate task is easier to reason about and debug

probably could be more efficient/concise

bin/reth/src/poa_builder.rs
```diff
@@ -184,6 +186,68 @@ impl<DB: Database + DatabaseMetrics + DatabaseMetadata + 'static> NodeBuilderWit
...
+
+            let mut current_filters: Vec<(json::GetBlockFilterResult, u64)> = match bitcoin_block_filters.read().await.clone() {
+                Some(filters) => filters,
+                None => Vec::new(),
+            };
+            loop {
+                let tip = match bitcoind_client.get_tip().await {
+                    Ok(tip) => tip,
+                    Err(_) => {
+                        error!(target: "reth::cli", "Failed to fetch the tip. Retrying...");
```
arminsabouri: Maybe update the target to be something specific to fetching block filters. Just b/c this error is also logged during the async bitcoin header task

arminsabouri (commented):
Conceptually this code looks good!

bin/reth/src/poa_builder.rs
scottmillner: done!

scottmillner:
closing bc don't need block filters

PR #122, opened 2024-03-11
Utxomerkle

ozymandias-rex:
This PR:

- adds an endpoint to retrieve all the utxos
- updates the Merkle Tree stored in the db on every pegin
- adds a remove_txo endpoint
closes #https://github.com/botanix-labs/botanix/issues/294


bin/btc-server/src/merkle.rs
```diff
@@ -0,0 +1,66 @@
+use crate::database::{Db, Utxo};
+use rs_merkle::{algorithms::Sha256 as MerkleSha256, MerkleTree};
+use serde_cbor;
+use sha2::{Digest, Sha256};
+
+pub fn hash_utxo(utxo: &Utxo) -> [u8; 32] {
+    let utxo_bytes = serde_cbor::to_vec(utxo).expect("Failed to serialize UTXO");
+    Sha256::digest(&utxo_bytes).into()
+}
+
```
scottmillner, line 10: i see  `construct_merkle_tree` and `construct_merkle_tree_from_utxos` aren't be used...do we plan to use these at some point?

one creates a tree of Utxos while the other creates a tree of hashed Utxos...why do we need both?

scottmillner:
@ozymandias-rex awesome sauce!  just had a question around two helper methods and left a comment

bin/btc-server/src/merkle.rs
ozymandias-rex: nice catch ! removed the unused ones

scottmillner (approved):

ozymandias-rex:
Notes 

- [ ] We probably want a type alias on the merkle tree.

bin/btc-server/src/database.rs
```diff
@@ -61,6 +68,9 @@ pub struct Db {
     /// round 1 signing commitments and round 2 partial signatures are commited inside the psbt
     /// Only relevant for the coordinator
     psbt: sled::Tree,
+
+    /// A tree of merkle trees
+    merkle_trees: sled::Tree,
```
arminsabouri: Nit: maybe a better name for this would be `utxo_merkel_tree`. As there should only be one tree

bin/btc-server/src/database.rs
```diff
@@ -362,6 +374,49 @@ impl Db {
...
+        Ok(())
+    }
+
+    /// Stores the Merkle root of UTXOs for a given transaction ID.
+    pub fn store_utxo_merkle_root(
+        &self,
+        txid: &bitcoin::Txid,
+        merkle_root: &[u8; 32],
+    ) -> Result<(), sled::Error> {
+        let txid_bytes = bitcoin::consensus::encode::serialize(txid);
```
arminsabouri: Im not sure if we should index the merkel tree by a particular txid. Since its repersenting all UTXOs (txid, vout pairs) that are currently spendable. Might make sense to construct this new model similar to how  `PUBKEY_PACKAGE` is constructed.

bin/btc-server/src/merkle.rs
```diff
@@ -0,0 +1,67 @@
...
+// pub fn construct_merkle_tree(utxos: &[Utxo]) -> MerkleTree<MerkleSha256> {
```
arminsabouri: Can we remove this commented out part?

bin/btc-server/src/database.rs
arminsabouri: There should ideally be an gRPC endpoint to get "the" UTXO merkel root. Not for a particular txid.
This root repersents all the UTXO that are spendable. The key here is that federation members have majority consensus on this [UTXO set].

arminsabouri (changes requested):
This looks great! Just a few things around the database changes

BatiGencho (approved):

BatiGencho:
Great job!

arminsabouri:
@0xBEEFCAF3 To re-review and @ozymandias-rex to rebase

arminsabouri (approved):
Looks good!

arminsabouri:
@ozymandias-rex not a big deal rn. But in the future can you rebase and merge the changes instead of creating the merge commit. And these could probably all be squshed into one. Makes rebasing easier and commits become atomic.

ozymandias-rex:
Apologies @0xBEEFCAF3  will rebase and merge next time

PR #123, opened 2024-03-11
frost signing compilation errors fixed & added grpc config, reflection, handling, term handlers etc.

BatiGencho:
This MR does:

- makes the frost signing branch compile (fmt, compilation errors and clippy all fixed)
- adds TODOs to the places where the code is unhandled @0xBEEFCAF3 is aware of that
- adds a toml file parser with a grpc section and config settings for extended behaviour
- adds reflection and healtcheck to grpc server
- adds gzip compression
- adds max encoding decoding sizes
- adds a missing grpc endpoint
- adds proper grpc termination handler (important as it is a separate mircoservice)


bin/btc-server/src/rpc/btc_server.bin
arminsabouri, line 1: What purpose is this file serving?

crates/consensus/authority/src/dkg.rs
```diff
@@ -226,8 +226,9 @@ impl DKGStateMachine {
     }
 
     async fn get_round1_dkg_packages(&mut self) -> Result<DkgPayload, Error> {
-        let round2_payload =
-            self.btc_client.get_round1_dkg_packages(tonic::Request::new(client::Empty {})).await;
+        // TODO: fix this code
```
arminsabouri: Should this get fixed before we merge this PR?

crates/consensus/authority/src/dkg.rs
```diff
@@ -244,11 +245,17 @@ impl DKGStateMachine {
...
     }
 
     pub(crate) async fn get_public_key(&mut self) -> Result<GetPublicKeyResponse, Error> {
-        let round3_payload =
-            self.btc_client.get_public_key(tonic::Request::new(client::Empty {})).await;
+        let round3_payload = self
+            .btc_client
+            .get_public_key(tonic::Request::new(GetPublicKeyRequest {
+                eth_address: "XXX".to_string(),
+            }))
```
arminsabouri, line 254: Just a heads up, im working on a PR that breaks out `get_public_key` and `get_gateway_address`. In the latter the key is tweaked and and an address is generated. In the former it just returns the aggregate key and requires no inputs

arminsabouri (commented):
Awesome work with the config and graceful shutdown. Just a few questions

bin/btc-server/src/rpc/btc_server.bin
BatiGencho: This file contains the reflection defintions. The idea is that a grpc client or an api does not need the exact protofiles (or even an sdk in case we have one) but can rather ping the server and it would return the latest proto definitions to the client ready for use. Its more like a discovery mechanism for gprc

crates/consensus/authority/src/dkg.rs
BatiGencho: Yeah, I will add this one endpoint that I need.

crates/consensus/authority/src/dkg.rs
BatiGencho: That is fine. Maybe we could merge this one first and then you could rebase and apply the fixes ?

crates/consensus/authority/src/dkg.rs
BatiGencho: added

crates/consensus/authority/src/dkg.rs
arminsabouri: Yeah that sounds good

crates/btc-wallet/src/address.rs
```diff
@@ -219,7 +222,7 @@ mod tests {
         )
         .unwrap();
 
-        let gateway = gateway_address(&secp, &key_pair.public_key(), &eth_addr, network).unwrap();
+        let gateway = gateway_address(&key_pair.public_key(), network).unwrap();
```
ozymandias-rex, line 225: Should we be handling / returning this error properly ?

ozymandias-rex (approved):

scottmillner (approved):

arminsabouri (approved):

PR #125, opened 2024-03-13
Arm/dkg mint validation

arminsabouri:
Saves aggregate public key after dkg in authority storage. And passes that to consensus validation using the new `BotanixConsensusPackage` 

Creating commits for each atomic change. So should be reviewable by just looking at each individual commit

### Summary
- Fixed async bitcoin worker tip retrieval
- Use dkg key when doing mint validation
- Store dkg key in PoA storage cache after retrieving it
- Dynamically created the pegin ScriptPubkey using tweaked pk during mint validation


arminsabouri:
Also @scottmillner This pr also removed executor.rs b/c I needed to change the `execute_transactions()` interface

README.md
```diff
@@ -12,24 +12,24 @@
 To run the stack locally, please go through the following steps to ensure you have all necessary prerequisites:
 
 1. Install `rust` (best way to install it is through the rustup toolchain: [rust](https://rustup.rs/) - depending on your OS). Default to nightly version. Minimum required version is `1.75`.
-2. Install `docker` on your OS - simply follow the instructions here: [docker](https://docs.docker.com/engine/install/)
-3. Install `foundry/forge` on your system - simply follow the instructions here: [foundry](https://book.getfoundry.sh/getting-started/installation)
-4. Install and set up `git` to use ssh.
-7. Install protobuf native dependencies: `apt update && apt upgrade -y && apt install -y protobuf-compiler libprotobuf-dev` (for ubuntu only)
-8. Install libclang dependeny: `sudo apt-get install libclang-dev` (for ubuntu only)
-9. Install gcp-cli: [google-cloud-cli](https://cloud.google.com/sdk/docs/install)
-9. Install [k9s](https://k9scli.io/topics/install/)
```
scottmillner, line 21: should these all be 1. ?

bin/reth/src/poa_builder.rs
```diff
@@ -191,41 +192,44 @@ impl<DB: Database + DatabaseMetrics + DatabaseMetadata + 'static> NodeBuilderWit
...
                             drop(header_write);
-                            error!(target: "reth::cli", "Failed to fetch the tip. Retrying...");
+                            error!(target: "reth::cli", "Failed to fetch the best block hash. Retrying...");
                             tokio::time::sleep(sleep_ms).await;
                             continue;
                         }
                     };
-                    if current_tip != tip {
+                   
+                    if current_block_hash != best_block_hash {
```
scottmillner, line 207: nit for info!: tip > block_hash

bin/reth/src/poa_builder.rs
```diff
@@ -191,41 +192,44 @@ impl<DB: Database + DatabaseMetrics + DatabaseMetadata + 'static> NodeBuilderWit
...
                                 drop(header_write);
-                                error!(target: "reth::cli", "Failed to fetch a block hash. Retrying...");
+                                error!(target: "reth::cli", "Failed to fetch a block header. Retrying...");
                                 tokio::time::sleep(sleep_ms).await;
                                 continue;
                             }
                         };
-                        let block_header = match bitcoind_client.get_block_header(block_hash).await {
+
+                        let tip = match bitcoind_client.get_tip().await {
```
scottmillner, line 219: nit: block_header > tip

crates/consensus/authority/src/frost_task.rs
```diff
@@ -87,14 +91,34 @@ where
             }
         };
 
+
```
scottmillner, line 94: is it possible to ever have the public key in storage before we call `start_task`?
would we ever need this code block?

crates/consensus/authority/src/block_fetcher.rs
```diff
@@ -122,10 +124,28 @@ where
             }
             let mut storage = self.storage.write().await;
 
+            if recent_bitcoin_block_header.is_none() {
+                warn!(target: "consensus::authority", "Do not have recent block header in memory, skipping block import");
+                continue;
+            }
+
```
scottmillner, line 131: lil confused and need to understand the flow better

why is the `aggregate_public_key` needed in non-epoch end blocks?

don't we only do dkg before frost signing when sealing the end of an epoch?

crates/consensus/authority/src/epoch_manager.rs
```diff
@@ -63,7 +64,7 @@ where
                         .receipts_by_block(BlockHashOrNumber::Number(block.header.number))
                     {
                         Ok(Some(receipts)) => {
```
scottmillner, line 65: this method still uses the `pegouts: Vec<MakeTxRequest>` which is before the `for in` block
the `pegouts: Vec<PegoutData>` in the match arm isn't acually returned from the function

crates/consensus/authority/src/utils.rs
```diff
@@ -140,19 +141,12 @@ pub(crate) async fn process_receipts(
 /// # Returns
 ///
```
scottmillner, line 144: nit in comments: MakeTxRequest > PegoutData

crates/consensus/authority/src/utils.rs
```diff
@@ -166,26 +160,28 @@ async fn make_tx_request_for_pegout(log: Log) -> Option<MakeTxRequest> {
 pub(crate) async fn send_pegouts(
     bitcoin_block_source: &BitcoindClient,
     btc_server: &mut BtcServerClient<tonic::transport::Channel>,
-    pegouts: Vec<MakeTxRequest>,
+    pegouts: Vec<PegoutData>,
```
scottmillner, line 168: minor refactor: can we have `send_pegouts` return the `psbt` from `btc_server.get_psbt()` call?

`block_builder` will need psbt returned so it can extract witness data and put into EDH when sealing block

README.md
arminsabouri: yea markdown will convert this a ordered list of numbers if they are all ones. I only made this change cause I noticed 9 was being used twice

bin/reth/src/poa_builder.rs
arminsabouri: Oops thanks for catching that

scottmillner (approved):
🔥 🔥 🔥 🔥 🔥 
that's 5 😮

PR #126, opened 2024-03-14
chore: test integration pipeline with hosted runner

ozymandias-rex:


arminsabouri (approved):

arminsabouri:
This is a great start. If we want to get these pipelines fully operational agains there is a few pieces missing
```
error: the following required arguments were not provided:
  --btc-server <BTC_SERVER>
  --bitcoind.url <BITCOIND_URL>
  --bitcoind.username <BITCOIND_USERNAME>
  --bitcoind.password <BITCOIND_PASSWORD>
```

We need to instantiate a btc server and bitcoind node. Makes the most sense to use a regtest container for bitcoind.

PR #127, opened 2024-03-14
Scott/358 feature add pegout witness data to edh

scottmillner:


crates/botanix-lib/src/extra_data_header.rs
```diff
@@ -308,6 +342,14 @@ mod tests {
         let pk = secp256k1::PublicKey::from_slice(&maybe_pk.as_slice()).expect("a public key");
         // Check the public key is the same as one provided
         assert_eq!(pk, public_key);
+        // Check the length of the witness data
+        assert_eq!(buf[74..78], vec![1u8, 0u8, 0u8, 0u8].as_slice().to_owned());
+        // Check the witness data
+        let binding = buf[78..].to_vec();
+        let mut maybe_witness = binding.as_slice();
```
ozymandias-rex, line 349: nit: does this have to be mutable ?

ozymandias-rex (approved):

crates/consensus/authority/src/lib.rs
```diff
@@ -344,12 +345,19 @@ where
             .unwrap();
         header.state_root = state_root;
 
+        let witness_data_final = if header.is_poa_epoch() {
+            witness_data.clone().map_or(Some(vec![]), |witness_data| Some(witness_data))
```
arminsabouri, line 350: Maybe leave a TODO here that if there are pegouts pending from this epoch and there are no witnesses then we need to fail consensus validation

arminsabouri (commented):
Approach ACK

crates/consensus/authority/src/lib.rs
scottmillner: todo added

crates/botanix-lib/src/extra_data_header.rs
scottmillner: won't compile without mut in both places

PR #128, opened 2024-03-14
refactor: finalize_signing returns psbt instead of transaction

scottmillner:


arminsabouri (approved):

arminsabouri:
@scottmillner one thing that totally missed by review is we forgot to update the .proto file. THe other files are acutally autogenerated based on the .proto. I'll make that change in #125

scottmillner:
> @scottmillner one thing that totally missed by review is we forgot to update the .proto file. THe other files are acutally autogenerated based on the .proto. I'll make that change in #125

thanks! totally missed that!!

PR #131, opened 2024-03-15
feature: fetch and store block tx ids in memory

scottmillner:


bin/reth/src/poa_builder.rs
```diff
@@ -184,6 +186,70 @@ impl<DB: Database + DatabaseMetrics + DatabaseMetadata + 'static> NodeBuilderWit
             }
         }
 
+        // fetch and store bitcoin block tx ids
+        let is_testnet = is_testnet(self.config.chain.chain.id());
+        let confirmation_depth = get_confirmation_depth(is_testnet);
+        let bitcoin_block_tx_ids: Arc<RwLock<Option<Vec<(Vec<bitcoin::Txid>, u64)>>>> =
```
arminsabouri: How do you feel about A Map<u64, Vec<bitcoin::Txid>>

scottmillner: generally on fridays im devoid of all feelings in regards to data structures 😆 
i like it! faster lookup

scottmillner: it is done!

bin/reth/src/poa_builder.rs
```diff
@@ -184,6 +186,70 @@ impl<DB: Database + DatabaseMetrics + DatabaseMetadata + 'static> NodeBuilderWit
             }
         }
 
+        // fetch and store bitcoin block tx ids
+        let is_testnet = is_testnet(self.config.chain.chain.id());
+        let confirmation_depth = get_confirmation_depth(is_testnet);
+        let bitcoin_block_tx_ids: Arc<RwLock<Option<HashMap<u64, Vec<bitcoin::Txid>>>>> =
```
arminsabouri: Sorry. I should have noticed this in my first pass. Any reason why this needs to be optional. It might depends on where this is going to endup. If in consensus critical code its easier from a interface perspective to add an optional.

bin/reth/src/poa_builder.rs
```diff
@@ -184,6 +186,70 @@ impl<DB: Database + DatabaseMetrics + DatabaseMetadata + 'static> NodeBuilderWit
...
+                let start = tip - confirmation_depth_u64;
+                for height in start..=tip {
+                    // don't fetch tx ids we already have
+                    if !current_tx_ids.is_empty() && current_tx_ids.contains_key(&height) {
+                        continue;
+                    }
+                    let block_hash = match bitcoind_client.get_block_hash(height).await {
+                        Ok(block_hash) => block_hash,
+                        Err(_) => {
+                            error!(target: "reth::cli", "Failed to fetch block hash. Retrying...");
```
arminsabouri: At this point we have this same log message in a few places. Do you mind adding a mention of getting Txids?

bin/reth/src/poa_builder.rs
```diff
@@ -184,6 +186,70 @@ impl<DB: Database + DatabaseMetrics + DatabaseMetadata + 'static> NodeBuilderWit
...
+                        }
+                    };
+
+                    match bitcoind_client.get_block_info(&block_hash).await {
+                        Ok(block_info) => {
+                            let tx_ids = block_info.tx;
+                            current_tx_ids.insert(height, tx_ids);
+                        }
+                        Err(_) => {
+                            error!(target: "reth::cli", "Failed to fetch block info. Retrying...");
```
arminsabouri: Ditto

crates/btc-wallet/src/bitcoind.rs
```diff
@@ -78,6 +78,17 @@ impl BitcoindClient {
...
     }
+    
+    pub async fn get_block_info(
+        &self,
+        block_hash: &bitcoin::BlockHash,
+    ) -> Result<GetBlockResult, BitcoindError> {
+        let block = self
+            .rpc
+            .get_block_info(block_hash)
+            .map_err(BitcoindError::BlockHeaderRetrievalFailed)?;
```
arminsabouri: NIT: custom error variant for this operation

arminsabouri (commented):
Just a few things I missed in my initial pass. Approach ACK

bin/reth/src/poa_builder.rs
scottmillner: good point, not used in consensus critical code: nodes just need it to track pegouts but utxos and witness data are the consesus pieces
removed Option, thanks!

bin/reth/src/poa_builder.rs
scottmillner: done!

bin/reth/src/poa_builder.rs
scottmillner: done!

crates/btc-wallet/src/bitcoind.rs
scottmillner: done!

bin/reth/src/poa_builder.rs
```diff
@@ -185,6 +187,67 @@ impl<DB: Database + DatabaseMetrics + DatabaseMetadata + 'static> NodeBuilderWit
...
+        let is_testnet = is_testnet(self.config.chain.chain.id());
+        let confirmation_depth = get_confirmation_depth(is_testnet);
+        let bitcoin_block_tx_ids: Arc<RwLock<HashMap<u64, Vec<bitcoin::Txid>>>> =
+        Arc::new(RwLock::new(HashMap::new()));
+        let bitcoin_block_tx_ids_clone = bitcoin_block_tx_ids.clone();
+        let bitcoind_config_clone = bitcoind_config.clone();
+
+        executor.spawn_critical("async bitcoin block tx ids task", Box::pin(async move {
+            let sleep_ms = tokio::time::Duration::from_millis(5000);
+            let bitcoind_client =  BitcoindClient::new(bitcoind_config_clone).expect("Unable to create bitcoind client");
```
arminsabouri, line 200: Out of scope for this PR. But in the future we'll probably want to just have one instance of the bitcoind client. No reason to re-create it. Follow up PR should be fine

bin/reth/src/poa_builder.rs
```diff
@@ -185,6 +187,67 @@ impl<DB: Database + DatabaseMetrics + DatabaseMetadata + 'static> NodeBuilderWit
...
+                    Ok(tip) => tip,
+                    Err(_) => {
+                        error!(target: "reth::cli", "Failed to fetch the tip. Retrying...");
+                        tokio::time::sleep(sleep_ms).await;
+                        continue;
+                    }
+                };
+
+                // prune tx ids older than confirmation_depth
+                let confirmation_depth_u64 = <u32 as Into<u64>>::into(confirmation_depth);
```
arminsabouri, line 214: Again out the scope of this PR. Im not sure when we decided heights are u64, im pretty sure they are u32. We'll clean this up in a follow up PR

arminsabouri (approved):

PR #132, opened 2024-03-18
added integration test suite and moved integration tests into it

BatiGencho:

PR #133, opened 2024-03-18
360 feature signers need to finalize psbt

arminsabouri:
In this last step singers take the tx witness data and shove it back in the original PSBT. If validation passes, then we know a correct signature was produced for the corresponding Script pubkey. And by extension that the multisig was done correct


bin/btc-server/src/signer.rs
```diff
@@ -63,6 +67,20 @@ pub enum SigningRound2Error {
     PsbtToSigningPackageConversionError(#[from] crate::util::PsbtToSigningPackageConversionError),
 }
 
+#[derive(Debug, Error)]
+pub enum SigningFinalizeError {
+    #[error("missing key package")]
+    MissingKeyPackage,
+    #[error("too many witness items")]
```
scottmillner, line 74: just curious...when/how would we have a `TooManyWitnessItems` error?

bin/btc-server/src/signer.rs
```diff
@@ -63,6 +67,20 @@ pub enum SigningRound2Error {
...
+    TooManyWitnessItems,
+    #[error("PSBT finalization failed : {0:?}")]
+    PsbtFinalizationFailed(Vec<PsbtError>),
+    #[error("Taproot Signature validation error: {0}")]
```
scottmillner, line 78: same question for `TaprootSignatureValidationError` since i don't see these errors used yet in the codebase?

bin/btc-server/tests/integration_tests.rs
```diff
@@ -315,13 +318,46 @@ async fn test_many_inputs_signing() {
...
 
-    let _finalized = c3
+    let finalized = c3
         .finalize_signing(tonic::Request::new(client::FinalizeSigningRequest {
             signing_session_id: signing_session_id.to_vec(),
         }))
         .await
         .unwrap();
 
+    let psbt = Psbt::deserialize(finalized.into_inner().psbt.as_slice()).expect("valid psbt");
```
scottmillner, line 328: just a thought...do you think it's helpful to use the method `utils::get_witness_data_from_psbt()` here and kill two birds with one stone? although that utils method does have unit tests

feel free to ignore this comment

scottmillner (approved):
💪 🚀 🔥 
awesome! just had a few comments but nothing that would stop merging the pr

bin/btc-server/src/signer.rs
arminsabouri: Good q. Since we are only spending from the default taproot key path there should only ever be one witness item per input. That being the signature

bin/btc-server/tests/integration_tests.rs
arminsabouri: Yeah that could work. If its a poa util that might be tough as we try not to create dependencies between btc_server and reth

bin/btc-server/src/signer.rs
scottmillner: thanks for the explanation!

arminsabouri:
Waiting for #132 to get merged in before rebase

arminsabouri:
rebasing this was too hard. recreated here: https://github.com/botanix-labs/Macbeth/pull/144

PR #135, opened 2024-03-19
Reworked btc server client and added jwt auth

BatiGencho:


bin/btc-server/src/jwt.rs
```diff
@@ -0,0 +1,207 @@
...
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct Claims {
+    /// The "iat" value MUST be a number containing a NumericDate value.
+    /// According to the RFC A NumericDate represents the number of seconds since
+    /// the UNIX_EPOCH.
+    /// - [`RFC-7519 - Spec`](https://www.rfc-editor.org/rfc/rfc7519#section-4.1.6)
+    /// - [`RFC-7519 - Notations`](https://www.rfc-editor.org/rfc/rfc7519#section-2)
+    pub iat: u64,
+    /// Expiration, if any
```
arminsabouri, line 197: ideally this secret rotates with some refresh mechanism. But thats out the scope of this PR. I'll create a new issue for this

bin/btc-server/src/main.rs
```diff
@@ -121,6 +130,8 @@ struct App {
         Arc<Mutex<Option<Vec<(frost::round1::SigningNonces, frost::round1::SigningCommitments)>>>>,
     /// configuration
     config: Config,
+    /// Jwt Auth Token
+    jwt_auth_token: Option<JwtSecret>,
```
arminsabouri: Should this optional? The only senario I can think of is testing

arminsabouri (commented):
Overall this approach is good to start with 👍 . In the future its likely we'll want to use a seperate JWT secret than the one used for networking traffic. 

I think some changes from #132 snuck themselves in here. We'll probably rebase this after that gets it and I'l re-review. Good work!

bin/btc-server/src/main.rs
BatiGencho: Yes, this was the idea. I can make it non-optional if you wish

arminsabouri:
After #132 merge lets rebase this and we can re-review. Over all the approach seems correct

bin/btc-server/src/main.rs
arminsabouri: I think that is fine for now

bin/btc-server/src/main.rs
```diff
@@ -239,6 +259,26 @@ impl App {
...
+        if let Some(server_token) = self.jwt_auth_token.as_ref() {
+            if let Some(request_token) = request.metadata().get(JWT_HEADER_KEY) {
+                let request_token = request_token.to_str().map_err(|e| {
+                    error!("Failed to get request token from request metadata: {}", e);
+                    badarg!("Failed to get request token from request metadata: {}", e)
+                })?;
+                let request_token = JwtSecret::from_str(request_token).map_err(|e| {
+                    error!("Failed to parse jwt token. Maybe illegal ?: {}", e);
+                    badarg!("Failed to parse jwt token. Maybe illegal ?: {}", e)
+                })?;
```
arminsabouri: I have trouble following this. Is JWTSecret an actual secret or just the token? I'm pretty sure its just a token.

bin/btc-server/src/jwt.rs
```diff
@@ -0,0 +1,207 @@
...
+    #[error(transparent)]
+    IOError(#[from] std::io::Error),
+}
+
+/// Length of the hex-encoded 256 bit secret key.
+/// A 256-bit encoded string in Rust has a length of 64 digits because each digit represents 4 bits
+/// of data. In hexadecimal representation, each digit can have 16 possible values (0-9 and A-F), so
+/// 4 bits can be represented using a single hex digit. Therefore, to represent a 256-bit string,
+/// we need 64 hexadecimal digits (256 bits ÷ 4 bits per digit = 64 digits).
+const JWT_SECRET_LEN: usize = 64;
```
arminsabouri, line 55: Why not just repersent this in bytes?

bin/btc-server/src/jwt.rs
```diff
@@ -0,0 +1,207 @@
...
+        Ok(secret)
+    }
+
+    /// Validates a JWT token along the following rules:
+    /// - The JWT signature is valid.
+    /// - The JWT is signed with the `HMAC + SHA256 (HS256)` algorithm.
+    /// - The JWT `iat` (issued-at) claim is a timestamp within +-60 seconds from the current time.
+    ///
+    /// See also: [JWT Claims - Engine API specs](https://github.com/ethereum/execution-apis/blob/main/src/engine/authentication.md#jwt-claims)
+    pub fn validate(&self, jwt: String) -> Result<(), JwtError> {
```
arminsabouri, line 125: This function makes sense but are we actually using it? I think validate_jwt() should be calling this

arminsabouri (commented):
Just had a few questions.

crates/consensus/authority/src/extended_client.rs
```diff
@@ -0,0 +1,100 @@
+///! Extended bitcoin server client with authentication
```
scottmillner, line 1: does it make more sense if this file lives in the `btc-server/client` folder?
any benefit of it being in the `consensus` crate?

bin/btc-server/src/jwt.rs
BatiGencho: agreeed, lets do this

crates/consensus/authority/src/extended_client.rs
BatiGencho: I tried putting the file inside the btc-sever, but it creates  a cyclic deps :(

bin/btc-server/src/jwt.rs
BatiGencho: This impl was taken from the jwt impl in reth. Due to cyclic deps, the btc server cannot impor the reth impl as the latter already imports btc server. I have not touched that at all.

bin/btc-server/src/main.rs
BatiGencho: its the token.

BatiGencho: The jwt is a token.

bin/btc-server/src/jwt.rs
BatiGencho: Yes, I forgot to add the actual validation. Will do it.

crates/consensus/authority/src/extended_client.rs
scottmillner: ah gotchya...do we need separate structs? can we just refactor `BtcServerClient`?

arminsabouri: yea i've had this issue in the past. Ideally btc-server has no reth dependencies

BatiGencho: We cannot do this right now as  we already have:

reth-btc-wallet = { path = "../../crates/btc-wallet" }
client = {path = "./client"}

We could come up with an alternative arrangement but we would need a separate ticket for that

bin/btc-server/src/jwt.rs
BatiGencho: added

bin/btc-server/src/main.rs
BatiGencho: Sorry, I had been passing around the jwt secret which was wrong :( I have fixed all of tha tnow in the MR.

bin/btc-server/src/main.rs
```diff
@@ -239,6 +259,25 @@ impl App {
...
+        if let Some(jwt_secret) = self.jwt_secret.as_ref() {
+            if let Some(jwt_request_token) = request.metadata().get(JWT_HEADER_KEY) {
+                let jwt_request_token = jwt_request_token
+                    .to_str()
+                    .map_err(|e| {
+                        error!("Failed to get request token from request metadata: {}", e);
+                        badarg!("Failed to get request token from request metadata: {}", e)
+                    })?
+                    .to_string();
+                if jwt_secret.validate(jwt_request_token).is_err() {
```
arminsabouri, line 273: Nice! this looks right to me now!

arminsabouri (approved):

PR #124, opened 2024-03-12
Frost signing & networking DRAFT

BatiGencho:

PR #136, opened 2024-03-20
frost signing reincarnated with latest poA changes

BatiGencho:
This PR supersedes https://github.com/botanix-labs/Macbeth/pull/124 which I have now closed.

NOTE: a few minor things are missing and currently being implemented


crates/net/network/src/frost/mod.rs
```diff
@@ -27,32 +28,67 @@ impl ProtocolState {
...
     DkgRound1,
     /// DKG round 2
     DkgRound2,
 }
 
+/// Event Response Variants indicating the type of response
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
+pub enum SigningEventResponseType {
+    /// Initiates a new signing session
+    InitiateSigningSession,
```
arminsabouri, line 82: What is exactly happening at this stage? Is this collecting round 1 nonces?

crates/net/network/src/frost/mod.rs
```diff
@@ -27,32 +28,67 @@ impl ProtocolState {
...
+    /// Signers will add their signing commitments to the psbt
+    SigningRound1SigningPackage,
+    /// Coordinating node will collect the PSBTs with the signing commitments
+    NewRound1SigningPackage,
+    /// Signers get round 2 signing package
+    SigningRound2SigningPackage,
```
arminsabouri: Maybe we can prefix `cordinator` or `signer` in the names here. To make it clear what entity cares about what event.

Something like SignerRound1 or CordinatorRound1

crates/consensus/authority/src/signing.rs
```diff
@@ -0,0 +1,660 @@
...
+
+/// A state machine for transitioning between different DKG states
+#[derive(Debug, Clone)]
+pub(crate) struct SigningStateMachine<Client> {
+    btc_client: BtcServerClient<tonic::transport::Channel>,
+    storage: Storage<Client>,
+    frost_handle: FrostHandle,
+    state: SigningState,
+    personal_frost_identifier: frost::Identifier,
+    public_key_package: Option<secp256k1::PublicKey>,
```
arminsabouri: Nit: can we call this aggregate_publickey or verifying_key. public key package referes to a specific type in the frost library

arminsabouri: Also this value should also be stored in storage. can we just use that one?

crates/consensus/authority/src/signing.rs
```diff
@@ -0,0 +1,660 @@
...
+                tonic::Code::Internal if e.message().contains("Failed to serialize psbt") => {
+                    return Err(Error::FailedToSerializePsbt)
+                }
+                _ => return Err(Error::InternalGrpc),
+            },
+        };
+        Ok(round1_payload)
+    }
+
+    async fn get_round2_dkg_package(&mut self) -> Result<DkgPayload, Error> {
```
arminsabouri: Are these here just to get the public key? If so we can just the `self.storage`. #125 added the public key into the poa storage cache

crates/consensus/authority/src/signing.rs
```diff
@@ -0,0 +1,660 @@
...
+                tonic::Code::Internal if e.message().contains("Missing round2 dkg package") => {
+                    return Err(Error::MissingRound2Package)
+                }
+                _ => return Err(Error::InternalGrpc),
+            },
+        };
+        Ok(round3_payload)
+    }
+
+    async fn add_round1_dkg_package(
```
arminsabouri: Copy pasta?

crates/node-core/src/args/frost_args.rs
```diff
@@ -0,0 +1,42 @@
...
+pub struct FrostArgs {
+    /// Coordinator
+    ///
+    /// Indicates if the node is a coordinator.
+    #[arg(
+        long = "frost.is_coordinator",
+        name = "frost.is_coordinator",
+        value_name = "IS_COORDINATOR"
+    )]
+    pub is_coordinator: bool,
```
arminsabouri: I'm not sure this should come from the cli. The current cordinator is the one `in_turn`

arminsabouri (commented):
Approach ACK. I had one concern around where `in_turn` is coming from. And some other questions.

crates/node-core/src/args/frost_args.rs
BatiGencho: Oh, was not sure here, I though we should go with a super trusted coordinator which we fix in the first release. Sorry, will fix that.

crates/consensus/authority/src/signing.rs
BatiGencho: hmmm, outdated. I believe I had this fixed during your review :)

crates/consensus/authority/src/signing.rs
BatiGencho: Again, I believe I removed this one during your review. It was a leftover. The code is no logner there and therefore outdated.

crates/consensus/authority/src/signing.rs
BatiGencho: I have removed it now as it can be retrieved directly from storage  +we dont need it here as the btc server takes care of that.

crates/net/network/src/frost/mod.rs
BatiGencho: This is the frost message that the block_builder.rs (send_pegouts) is sending to frost_task to initiate a signing session for a set of psbt.

crates/net/network/src/frost/mod.rs
BatiGencho: done, changed throught the code

scottmillner (approved):
🔥 🔥 🔥 🔥 🔥

arminsabouri (approved):

PR #140, opened 2024-03-25
refactor: process pegouts before building block header

scottmillner:


crates/consensus/authority/src/block_builder.rs
```diff
@@ -189,6 +184,90 @@ where
             }
         }
 
+        // If end of epoch, process pegouts
+        let mut witness_data: Option<Vec<Witness>> = None;
+        if is_epoch_end(block.header.number) {
```
arminsabouri: can we just use the function on the header `is_epoch?`

crates/consensus/authority/src/block_builder.rs
```diff
@@ -189,6 +184,90 @@ where
...
+                    return;
+                }
+            };
+
+            // add current block pegouts
+            pegouts.extend(current_block_pegouts);
+
+            // send pegouts
+            info!(target: "consensus::authority", "Sending pegouts: {:?}", pegouts);
+            if let Err(e) = send_pegouts(
```
arminsabouri, line 208: What does send_pegouts do? Does this initiate frost signing somewhere?

crates/consensus/authority/src/block_builder.rs
```diff
@@ -189,6 +184,90 @@ where
...
+                // Exponential backoff
+                delay *= 2;
+                tokio::time::sleep(delay).await;
+            }
+
+            // TODO(scott): check psbt matches expected session id
+
+            // extract witness data from psbt to be included in the header
+            if let Some(psbt) = psbt_message {
+                let psbt_result = Psbt::deserialize(&psbt.psbt).expect("valid psbt");
```
arminsabouri: Any reason why we shouldn't map_err here and error out?

arminsabouri (commented):
Just had some questions

crates/consensus/authority/src/block_builder.rs
arminsabouri: Calls intiate frost 
closing

crates/consensus/authority/src/block_builder.rs
```diff
@@ -189,6 +186,73 @@ where
...
+            info!(target: "consensus::authority", "Sending pegouts: {:?}", pegouts);
+            if let Err(e) = send_pegouts(
+                &self.bitcoind_client,
+                &mut self.btc_server,
+                &self.frost_handle,
+                pegouts,
+            )
+            .await
+            {
+                error!(target: "consensus::authority", ?e, "Failed to send pegouts");
```
arminsabouri, line 216: Should we error out here? or continue to produce a block?

arminsabouri (approved):
Looks good. I would also have @BatiGencho take a look before merging

crates/consensus/authority/src/block_builder.rs
scottmillner: ive added a `return;`
- yeah i think we should just start the loop again and try and process the pegouts again
- if they keep failing, then node will go out of turn and next member will try and seal the epoch

crates/consensus/authority/src/block_builder.rs
scottmillner: refactored: error out now

BatiGencho (approved):

BatiGencho:
> Looks good. I would also have @BatiGencho take a look before merging

Approved

PR #147, opened 2024-03-26
refactor: add btc-network cli arg

scottmillner:
add back btc-network arg so we can target regtest

### DIFF
--- a/bin/btc-server/src/server.rs
+++ b/bin/btc-server/src/server.rs
@@ -15,7 +15,7 @@ use util::{parse_eth_address, VerifyingKeyExt};
 use crate::database::Utxo;
 use crate::{rpc, util, App, SECP};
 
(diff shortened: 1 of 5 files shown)


bin/btc-server/src/server.rs
```diff
@@ -15,7 +15,7 @@ use util::{parse_eth_address, VerifyingKeyExt};
 use crate::database::Utxo;
 use crate::{rpc, util, App, SECP};
 
-const JWT_HEADER_KEY: &'static str = "jwt-auth";
+const JWT_HEADER_KEY: &'static str = "trace-proto-bin";
```
arminsabouri, line 18: Im having issues following what these should be. @BatiGencho Can you verify this is correct

BatiGencho: its the second one what we should be using, scott is right.

BatiGencho (approved):

PR #154, opened 2024-04-04
feature(edh): authority signatures should be a list of sigs

arminsabouri:


crates/botanix-lib/src/extra_data_header.rs
```diff
@@ -263,23 +275,27 @@ impl ExtraDataHeader {
...
+        for sig in self.authority_signatures.expect("is some") {
+            let recovered_pk =
+                sig.recover(&msg).map_err(|_| ValidateAuthoritySignatureError::InvalidSignature)?;
+            for signer in authority_signers {
+                if signer == &recovered_pk && sig.to_standard().verify(&msg, signer).is_ok() {
+                    signer_count += 1;
+                }
+            }
+        }
+        // TODO (armins) change this to be a quorum of signatures
```
scottmillner, line 295: i know this will changed to a quorum but why `>=` instead of `==`?
doesn't look like the `signer_count` can ever be greater than the `authority_signers.len()`?
maybe i just need another coffee lol

scottmillner:
is this in preparation of PBFT?
as in a quorum of members will attest (sign) each EDH whereas before only the block producer was signing?

For my understanding, would the flow be:
- block builder sends proposed block to federation members
- members sign the edh and send back to block producer
- block producer shove all sigs into vec (incuding their own), appends to EDH then seals block?

arminsabouri:
> * block builder sends proposed block to federation members

yes in preperation for P*BFT.
the flow is best described here: https://blog.web3labs.com/web3development/comparing-byzantine-fault-tolerance-consensus-algorithms

arminsabouri:
Commits moved to [arm/feature/pbft-networking](https://github.com/botanix-labs/Macbeth/tree/arm/feature/pbft-networking)

PR #156, opened 2024-04-05
WIP - 377 fixfrost authenticate all dkg frost networking messages

arminsabouri:


crates/net/network/src/manager.rs
```diff
@@ -330,12 +330,20 @@ where
             mpsc::unbounded_channel::<FrostProtocolEvent>();
         let authority_index =
             frost_config.as_ref().map(|f| f.authority_index as u16).unwrap_or_default();
+        let authorities = frost_config
+            .as_ref()
+            .map(|f| f.authorities.clone())
+            .unwrap_or_default()
+            .iter()
+            .map(|pk| PeerId::from_slice(&pk.serialize_uncompressed()[1..]))
```
BatiGencho: There is a method in the ecies crate that actually does that - maybe we want to reuse it here instead of deriving it ourselves.? 

I mean reth_ecies::util::pk2id

BatiGencho:
Looks great to me, left just one comment... Needs to be tested though

arminsabouri:
@BatiGencho ready for review. Integration test running green locally

BatiGencho (approved):

BatiGencho:
Great, ready to go!

PR #160, opened 2024-04-10
Replace user/pass with cookie file for bitcoind RPC 

stevenroose:
Hmm, somehow it added one extra commit that is not mine..


arminsabouri (approved):
tACK
testing suite passing

PR #162, opened 2024-04-11
refactor: use timeout when getting new payload

scottmillner:
Using a timeout when getting a new payload and only retrying once.  One retry is enough as there's a race condition where payload builder just needs a little time to finish the payload.  I've never seen the payload not finish on first retry after waiting a second but never had a large payload so can always adjust as needed.

### DIFF
--- a/crates/consensus/authority/src/block_builder.rs
+++ b/crates/consensus/authority/src/block_builder.rs
@@ -75,41 +75,41 @@ where
         }
 
         let payload_id = payload_id.expect("payload id exists");
-
-        // retry if best_transactions is empty bc it could be a race condition
-        let mut retries = 0;
-        let mut delay = tokio::time::Duration::from_secs(1);
-        let max_retries = 5;
-        let mut best_transactions: Result<EthBuiltPayload, BestTransactionsError> =
-            Err(BestTransactionsError::PayloadEmpty);
-        loop {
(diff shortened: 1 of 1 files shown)


crates/consensus/authority/src/block_builder.rs
```diff
@@ -74,42 +74,44 @@ where
...
+                    tokio::time::sleep(tokio::time::Duration::from_secs(1)).await;
+                    let transactions =
+                        engine_util::best_transactions_from_payload::<EthEngineTypes>(
+                            &self.payload_builder,
+                            payload_id,
+                        )
+                        .await;
+
+                    if transactions.is_ok() {
+                        best_transactions = transactions;
```
arminsabouri: can we just return best_transaction out the match statment instead of setting a mut?
let x = match {
   Ok(s) => s;
   ... 
}

scottmillner: yep, that's cleaner, thanks!
and done!

arminsabouri: Closing

arminsabouri (approved):
integration passing
 tACK [c69d33c](https://github.com/botanix-labs/Macbeth/pull/162/commits/c69d33ca5539efd585a3ac1dfb62ae0151473076)

PR #166, opened 2024-04-11
don't track .env file and add template.env

scottmillner:


arminsabouri (approved):
Thanks for doing this!!!

PR #167, opened 2024-04-12
Scott/378 feature block subsidy is split between block producer and botanix

scottmillner:
currently using a placeholder address for botanix...will replace with actual once we have it

Overview of how fees work now:

1. Transaction fees are sent to the beneficiary address (revm sets coinbase address to beneficiary) during tx execution.
2. After block execution and during post state updates, block rewards are sent to the beneficiary.
3. Since the beneficiary is set to the burn address of 0x0..., all fees and rewards are burned
4. At this point, Botanix is deflationary: tx fees paid by senders have been burned.
5. `processor.execute` has been refactored to determine the total block fees as it executes txs (`best_transactions_from_payload` does this and I used this as reference)
6. During post state checks/updates, `apply_post_execution_state_change` has been refactored to accept the `total_block_fees`
7. The above method calls `post_block_balance_increments` which has been refactored to split `total_block_fees` between botanix and the block builder and update/commit their new balances.
8. At this point, Botanix is back to it's exact peg of 21million since the previously burned fees have been added to botanix and the block builder accounts.
9. `test_block_builder` has been refactored to assert that block fees are split correctly.
10. Have tested manually as well.

Screenshot 2024-04-17 at 1 47 14 PM (3)

### DIFF
--- a/bin/test-suite/src/suite/consensus/frost/test_block_builder.rs
+++ b/bin/test-suite/src/suite/consensus/frost/test_block_builder.rs
@@ -1,4 +1,6 @@
 use reth::core::cli::runner::CliRunner;
+use reth::primitives::{constants::BOTANIX_FEES_RECIPIENT, public_key_to_address};
+use reth_botanix_lib::extra_data_header::ExtraDataHeader;
 use std::{collections::HashSet, time::Duration};
 
 use crate::{
@@ -50,6 +52,25 @@ pub async fn block_builder(
     // create a minting contract instance
     let botanix_eth_client = targeted_fed_member.create_botanix_eth_client().await;
 
+    // get fed member and botanix address balances
+    let edh = targeted_fed_member.edh.unwrap();
+    let targeted_fed_member_pub_key = *edh.authority_signers.unwrap().first().unwrap();
+    let targeted_fed_member_ethereum_address =
+        public_key_to_address(targeted_fed_member_pub_key).to_checksum(Some(3636));
+
+    let target_fed_member_balance_before = botanix_eth_client
+        .get_botanix_balance(targeted_fed_member_ethereum_address.as_str())
+        .await
+        .unwrap();
+    it_info_print!("Targeted fed member balance before: {}", target_fed_member_balance_before);
+
+    let botanix_block_reward_address_balance_before =
+        botanix_eth_client.get_botanix_balance(BOTANIX_FEES_RECIPIENT).await.unwrap();
+    it_info_print!(
+        "Botanix block fee recipient balance before: {}",
+        botanix_block_reward_address_balance_before
+    );
+
     // create a hashmap to store tx hashes
     let mut tx_hashes_set = HashSet::new();
 
@@ -94,6 +115,39 @@ pub async fn block_builder(
                     assert!(!block_payload.1);
                     assert_eq!(block_payload.0.tx_receipts.len(), 1);
                     assert!(block_payload.0.block.number > 0);
+
+                    // get fed member and botanix block reward address balances
+                    let target_fed_member_balance_after = botanix_eth_client
+                        .get_botanix_balance(targeted_fed_member_ethereum_address.as_str())
+                        .await
+                        .unwrap();
+                    it_info_print!(
+                        "Targeted fed member balance after: {}",
+                        target_fed_member_balance_after
+                    );
+
+                    it_info_print!("Botanix block fee recipient: {}", BOTANIX_FEES_RECIPIENT);
+
+                    let botanix_block_reward_address_balance_before_after = botanix_eth_client
+                        .get_botanix_balance(BOTANIX_FEES_RECIPIENT)
+                        .await
+                        .unwrap();
+                    it_info_print!(
+                        "Botanix block reward address balance after: {}",
+                        botanix_block_reward_address_balance_before_after
+                    );
+
+                    // verfigy 80/20 block reward split is correct
+                    let target_fed_member_reward =
+                        target_fed_member_balance_after - target_fed_member_balance_before;
+                    let botanix_block_reward = botanix_block_reward_address_balance_before_after
+                        - botanix_block_reward_address_balance_before;
+
+                    let total_block_reward = target_fed_member_reward + botanix_block_reward;
+
+                    assert_eq!(target_fed_member_reward, (total_block_reward * 4) / 5); // 80%
+                    assert_eq!(botanix_block_reward, total_block_reward / 5); // 20%
+
                     return Ok(());
                 }
             }
--- a/bin/test-suite/src/suite/consensus/frost/test_block_builder.rs
+++ b/bin/test-suite/src/suite/consensus/frost/test_block_builder.rs
@@ -137,7 +137,7 @@ pub async fn block_builder(
                         botanix_block_reward_address_balance_before_after
                     );
 
-                    // verfigy 80/20 block reward split is correct
+                    // verify 80/20 block reward split is correct
                     let target_fed_member_reward =
                         target_fed_member_balance_after - target_fed_member_balance_before;
                     let botanix_block_reward = botanix_block_reward_address_balance_before_after
--- a/crates/consensus/authority/src/lib.rs
+++ b/crates/consensus/authority/src/lib.rs
@@ -21,7 +21,7 @@
 //! be mined.
 use reth_botanix_lib::extra_data_header::ExtraDataHeader;
 use reth_consensus_common::{
-    utils::{get_authority_address_from_header, recovery_authority, unix_timestamp},
+    utils::{get_block_producer_address, unix_timestamp},
     validation::{self, validate_poa_header_standalone},
 };
 use reth_interfaces::{
@@ -436,7 +436,7 @@ where
         let mut executor = EVMProcessor::new_with_state(chain_spec.clone(), db, evm_config);
 
         // derive block builder address to receive block fees
-        let block_builder_pub_key = secp256k1::PublicKey::from_secret_key(secp, sk);
+        let block_builder_pub_key = secp256k1::PublicKey::from_secret_key_global(sk);
         let block_builder_address = public_key_to_address(block_builder_pub_key);
         let (bundle_state, gas_used) = self.execute(
             &block_with_senders,
@@ -554,7 +554,7 @@ where
             },
         )?;
 
-        let block_builder_address = get_authority_address_from_header(&sealed_block.header.clone());
+        let block_builder_address = get_block_producer_address(&sealed_block.header.clone());
         let (bundle_state, _gas_used) = self.execute(
             &block_with_senders,
             &mut executor,
--- a/crates/consensus/authority/src/lib.rs
+++ b/crates/consensus/authority/src/lib.rs
@@ -21,7 +21,7 @@
 //! be mined.
 use reth_botanix_lib::extra_data_header::ExtraDataHeader;
 use reth_consensus_common::{
-    utils::unix_timestamp,
+    utils::{get_authority_address_from_header, recovery_authority, unix_timestamp},
     validation::{self, validate_poa_header_standalone},
 };
 use reth_interfaces::{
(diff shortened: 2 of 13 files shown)


crates/revm/src/state_change.rs
```diff
@@ -36,8 +36,12 @@ pub fn post_block_balance_increments(
         }
 
         // Full block reward
-        *balance_increments.entry(beneficiary).or_default() +=
-            calc::block_reward(base_block_reward, ommers.len());
+        let base_block_reward = calc::block_reward(base_block_reward, ommers.len());
```
arminsabouri: is it possible to add `block_reward_split` as a trait ext instead of a utility?

BatiGencho (approved):

arminsabouri (changes requested):
we should also set the beneficiary during block template construction to a unspendable address. Not sure what that would be in eth land? 0x0?

crates/revm/src/state_change.rs
scottmillner: not quite following...which trait would this be extending?

`state_change.rs` is just functions related to state change and aren't part of a struct or trait
and `calc.rs` which the state change methods use is the same: just a file of methods not related to a trait or struct

maybe i should move the method to `calc.rs`?

scottmillner:
> we should also set the beneficiary during block template construction to a unspendable address. Not sure what that would be in eth land? 0x0?

why wouldn't we want to set it to the block producer like we currently do in `build_header_template`?

```
// derive beneficary address being the producuing block federation member address
let beneficiary_pub_key = secp256k1::PublicKey::from_secret_key(secp, sk);
let beneficiary_address = public_key_to_address(beneficiary_pub_key);
```

are you saying the beneficiary would always be `0x0` and `block_reward_split` would be responsible for determining the beneficary?

arminsabouri:
> > we should also set the beneficiary during block template construction to a unspendable address. Not sure what that would be in eth land? 0x0?
> 
> why wouldn't we want to set it to the block producer like we currently do in `build_header_template`?
> 
> ```
> // derive beneficary address being the producuing block federation member address
> let beneficiary_pub_key = secp256k1::PublicKey::from_secret_key(secp, sk);
> let beneficiary_address = public_key_to_address(beneficiary_pub_key);
> ```
> 
> are you saying the beneficiary would always be `0x0` and `block_reward_split` would be responsible for determining the beneficary?

Yeah essentially. I think we either have the block fee reward either gets distributed via benefeciary (normal path). Or gets split via this incrementing balance logic. But you can't have both?

scottmillner:
> > > we should also set the beneficiary during block template construction to a unspendable address. Not sure what that would be in eth land? 0x0?
> > 
> > 
> > why wouldn't we want to set it to the block producer like we currently do in `build_header_template`?
> > ```
> > // derive beneficary address being the producuing block federation member address
> > let beneficiary_pub_key = secp256k1::PublicKey::from_secret_key(secp, sk);
> > let beneficiary_address = public_key_to_address(beneficiary_pub_key);
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > are you saying the beneficiary would always be `0x0` and `block_reward_split` would be responsible for determining the beneficary?
> 
> Yeah essentially. I think we either have the block fee reward either gets distributed via benefeciary (normal path). Or gets split via this incrementing balance logic. But you can't have both?

scottmillner:
> > > > we should also set the beneficiary during block template construction to a unspendable address. Not sure what that would be in eth land? 0x0?
> > > 
> > > 
> > > why wouldn't we want to set it to the block producer like we currently do in `build_header_template`?
> > > ```
> > > // derive beneficary address being the producuing block federation member address
> > > let beneficiary_pub_key = secp256k1::PublicKey::from_secret_key(secp, sk);
> > > let beneficiary_address = public_key_to_address(beneficiary_pub_key);
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > are you saying the beneficiary would always be `0x0` and `block_reward_split` would be responsible for determining the beneficary?
> > 
> > 
> > Yeah essentially. I think we either have the block fee reward either gets distributed via benefeciary (normal path). Or gets split via this incrementing balance logic. But you can't have both?

recap of offline discussion: beneficiary should be set to 0x0 so block reward (PoW 5eth + ommers reward) gets burned and incrementing balance logic handles all the rewards

scottmillner:
> we should also set the beneficiary during block template construction to a unspendable address. Not sure what that would be in eth land? 0x0?

done

crates/primitives/src/transaction/mod.rs
```diff
@@ -377,8 +377,9 @@ impl Transaction {
         let fee = max_fee_per_gas - base_fee;
 
         // Compare the fee with max_priority_fee_per_gas (or gas price for non-EIP1559 transactions)
```
scottmillner, line 379: this change was made before but must've got lost in a rebase...it makes sure we don't burn the base fee but include it in the block fees

bin/test-suite/src/suite/consensus/frost/test_block_builder.rs
```diff
@@ -94,6 +115,39 @@ pub async fn block_builder(
...
+                    // verfigy 80/20 block reward split is correct
+                    let target_fed_member_reward =
+                        target_fed_member_balance_after - target_fed_member_balance_before;
+                    let botanix_block_reward = botanix_block_reward_address_balance_before_after
+                        - botanix_block_reward_address_balance_before;
+
+                    let total_block_reward = target_fed_member_reward + botanix_block_reward;
+
+                    assert_eq!(target_fed_member_reward, (total_block_reward * 4) / 5); // 80%
+                    assert_eq!(botanix_block_reward, total_block_reward / 5); // 20%
```
arminsabouri, line 149: 🔥 🔥 🔥 🔥 🔥 🔥 
nice

crates/consensus/authority/src/lib.rs
```diff
@@ -433,11 +435,15 @@ where
 
         let mut executor = EVMProcessor::new_with_state(chain_spec.clone(), db, evm_config);
 
+        // derive block builder address to receive block fees
+        let block_builder_pub_key = secp256k1::PublicKey::from_secret_key(secp, sk);
```
arminsabouri: nit: im hoping we can drop `secp` as there is a global context now available in the library.
Something like `secp256k1::PublicKey::from_secret_key_global(sk)` would make that transition easier

bin/test-suite/src/suite/consensus/frost/test_block_builder.rs
```diff
@@ -94,6 +115,39 @@ pub async fn block_builder(
...
+                    let botanix_block_reward_address_balance_before_after = botanix_eth_client
+                        .get_botanix_balance(BOTANIX_FEES_RECIPIENT)
+                        .await
+                        .unwrap();
+                    it_info_print!(
+                        "Botanix block reward address balance after: {}",
+                        botanix_block_reward_address_balance_before_after
+                    );
+
+                    // verfigy 80/20 block reward split is correct
```
arminsabouri: Typo

```suggestion
                    // verify 80/20 block reward split is correct
```

crates/consensus/common/src/utils.rs
```diff
@@ -323,6 +322,21 @@ pub fn validate_inturn(
     Ok(())
 }
 
+// not in authority utils because of circular dependency
+/// Get the authority address from the header
+pub fn get_authority_address_from_header(header: &Header) -> Address {
+    let block_builder_public_key = recovery_authority(header).expect("recovered authority");
+    public_key_to_address(block_builder_public_key)
+}
```
arminsabouri: can we re-name this to `get_block_producer_address`. Only b/c in the multi signer context (PBFT) this is going to get confusing. And eventually we'll move this to the header ext as well

scottmillner: done

bin/test-suite/src/suite/consensus/frost/test_block_builder.rs
scottmillner: done

crates/consensus/authority/src/lib.rs
scottmillner: done

crates/primitives/src/constants/mod.rs
```diff
@@ -200,6 +200,10 @@ pub const SIGNET_PEGIN_CONFIRMATION_DEPTH: u32 = 1;
 /// Mainnet required confirmation depth for pegins
 pub const MAINNET_PEGIN_CONFIRMATION_DEPTH: u32 = 100;
 
+// TODO: replace with actual botanix address: using placeholder address for testing purposes
+/// The address that receives the botanix block fees.
+pub const BOTANIX_FEES_RECIPIENT: &str = "0x4e0f6e05C8ca4b3dc2B7b7Ad6249B149b1980394";
```
arminsabouri: Just curious is this an address of one of the two fed members?

crates/consensus/authority/src/lib.rs
```diff
@@ -296,22 +292,28 @@ where
...
+        block_builder_address: Option<Address>,
     ) -> Result<(BundleStateWithReceipts, u64), BlockExecutionError>
     where
         EvmConfig: ConfigureEvmEnv + Clone + 'static + reth_node_api::ConfigureEvm,
     {
         // set the first block to find the correct index in bundle state
         executor.set_first_block(block.number);
 
-        let (receipts, gas_used) =
+        let (receipts, gas_used, total_block_fees) =
```
arminsabouri, line 303: Whats the difference between gas_used and total_block_fees?

scottmillner: in this context, `gas_used` is the total gas needed for all the txs in the block

and `total_block_fees` = is the total fees paid for all the txs in the block

a `tx_fee` = `gas` * `gas_price`

`gas_price` = `base_fee` + `tip`

crates/primitives/src/constants/mod.rs
scottmillner: it's an address i use for testing...more than happy to share it's private key or switch to another account

just needed something for testing and confirm fees were being distributed correctly

crates/consensus/common/src/utils.rs
```diff
@@ -555,13 +578,13 @@ mod tests {
...
         header.beneficiary = beneficiary_address;
-        assert!(validate_poa_block_beneficiary(&header, &authority_signers).is_err());
+        // assert!(validate_poa_block_beneficiary(&header, &authority_signers).is_err());
 
         // test good signer
         let mut header = Header::default();
         let beneficiary_address = public_key_to_address(auth_signer2);
         header.beneficiary = beneficiary_address;
-        assert!(validate_poa_block_beneficiary(&header, &authority_signers).is_ok());
+        // assert!(validate_poa_block_beneficiary(&header, &authority_signers).is_ok());
```
arminsabouri: Can we just remove this test?

crates/primitives/src/constants/mod.rs
arminsabouri: Is it possible to one of the keys of the fed members?

crates/consensus/authority/src/lib.rs
arminsabouri: Understood. Thanks

crates/consensus/common/src/utils.rs
scottmillner: will do, was just leaving it for when we eventually set the beneficiary to block producer when we're not PoA and fully decentralized...but that's a wee bit off in the future lol

and done!

crates/primitives/src/constants/mod.rs
scottmillner: yes sir, and done!

crates/consensus/common/src/utils.rs
```diff
@@ -528,40 +542,11 @@ mod tests {
...
-        // test good signer
-        let mut header = Header::default();
-        let beneficiary_address = public_key_to_address(auth_signer2);
-        header.beneficiary = beneficiary_address;
-        assert!(validate_poa_block_beneficiary(&header, &authority_signers).is_ok());
+    fn should_validate_poa_block_beneficiary() {
+        // default beneficiary is the burn address
+        let header = Header::default();
+        let result = validate_poa_block_beneficiary(&header);
+        assert!(result.is_ok());
```
arminsabouri, line 549: Sorry I should have noticed this before 
can we have a test case for the opposite condition? Fails for anything that is not burn address. I know the code is really dead simple. but it would be good to have this fail if that code changes by accident

arminsabouri (approved):
Nice work!

crates/consensus/common/src/utils.rs
scottmillner: done!

PR #168, opened 2024-04-16
retry signing after current signing session fails

BatiGencho:


bin/test-suite/src/suite/consensus/frost/test_frost_e2e.rs
```diff
@@ -324,3 +326,248 @@ pub async fn frost_e2e(suite: &ConsensusIntegrationTestSuite) -> Result<(), supe
...
+    tokio::time::sleep(Duration::from_secs(3)).await;
+
+    // now disconnect the peers of fed member 0
+    test_fed_members.get(&0).cloned().unwrap().send_test_signal(TestSignal::DisconnectAll());
+
+    // wait for a new epoch to start
+    tokio::time::sleep(Duration::from_secs(10)).await;
+
+    // now reconnect the peers of fed member 0
+    test_fed_members.get(&0).cloned().unwrap().send_test_signal(TestSignal::DisconnectAll());
```
arminsabouri: The comment says reconnect but the signal is disconnect

bin/test-suite/src/suite/consensus/frost/test_frost_e2e.rs
```diff
@@ -324,3 +326,248 @@ pub async fn frost_e2e(suite: &ConsensusIntegrationTestSuite) -> Result<(), supe
...
+    let pegout_destination =
+        ethers::core::types::Bytes::from(btc_address.to_string().as_bytes().to_vec());
+    // use empty pegout data
+    let pegout_data = ethers::core::types::Bytes::new();
+    let pegout_amount = Amount::from_btc(0.5).unwrap();
+    let tx_receipt =
+        mint_contract.burn(pegout_destination, pegout_data, pegout_amount.to_wei()).await.unwrap();
+    it_info_print!("Pegout Tx Receipt: ", tx_receipt);
+
+    // ===================== FAILURE =====================
```
arminsabouri: Should this reconnect / disconnect be happening after L521?
Where we send the final eoa tx?

crates/consensus/authority/src/signing.rs
```diff
@@ -500,13 +610,52 @@ where
         psbt: Vec<u8>,
     ) -> Result<(), Error> {
         let session_id = parse_signing_session_id(&signing_session_id)?;
-        self.insert_signing_state(session_id, SigningState::Initial);
-        info!(">>>>>>>>>>> [START NEW SIGNING SESSION]");
+
+        // the coordinator is always expected to be us in this case, i.e. None
+        let coordinator = self.get_coordinator().await?;
+        if coordinator.is_some() {
+            panic!(">>>>>>>>>>> A non-coordinator is trying to (re)initiate a signing process!");
```
arminsabouri: Should we be panicing here? What If this is just a timing error?

crates/consensus/authority/src/utils.rs
```diff
@@ -495,12 +520,16 @@ pub(crate) enum GenerateSigningSesssionIdError {
 pub(crate) fn generate_signing_session_id(
     best_hash: &[u8],
     authority_pk: &secp256k1::PublicKey,
+    has_timestamp: bool,
 ) -> Result<SigningSessionId, GenerateSigningSesssionIdError> {
     let mut engine = sha256::HashEngine::default();
```
arminsabouri: I feel like this function ended up getting more complicated than it needs to for its intended usecase.
How do you feel about just hashing uuidv4()

crates/consensus/common/src/utils.rs
```diff
@@ -295,6 +295,23 @@ pub fn is_inturn(authorities_len: u64, signer_index: u64) -> bool {
     (timestamp / authorities_len) % authorities_len == signer_index
 }
 
+/// Returns the inturn interval for a signer index
+pub fn get_in_turn_interval(authorities_len: u64, signer_index: u64) -> (u64, u64, u64, u64) {
```
arminsabouri: Can we typedef `(u64, u64, u64, u64)` to somethign like `CoordinatorInterval`?

bin/test-suite/src/suite/consensus/frost/test_frost_e2e.rs
BatiGencho: correct, sorry. I have fixed it now.

bin/test-suite/src/suite/consensus/frost/test_frost_e2e.rs
BatiGencho: I have re-arranged this logic now.

crates/consensus/authority/src/signing.rs
BatiGencho: i made an error out of it. It should not happen though

crates/consensus/authority/src/utils.rs
BatiGencho: done. Much cleaner

crates/consensus/common/src/utils.rs
BatiGencho: done. Thx for the suggestion

crates/consensus/authority/src/signing.rs
```diff
@@ -545,49 +695,99 @@ where
...
+                .unwrap_or_default();
+
+            // abort this previous session as it is outdated
+            if unix_timestamp() >= prev_validity_to {
+                self.abort_signing().await?;
+                self.update_signing_state(session_id, SigningState::Failed);
+                return Ok(());
+            }
+
+            // session exists and is valid, return if we are not in round 2
```
scottmillner, line 745: nit: should say if we are not in round 1

it confused me for a second lol

scottmillner (dismissed):
as always 🔥 🔥 🔥 🔥 🔥 
awesome sauce!!

crates/net/network/src/frost/protocol.rs
```diff
@@ -163,6 +163,7 @@ impl Stream for FrostProtoConnection {
             return Poll::Ready(Some(initial_ping.encoded()))
         }
         let peer_message_forwarder = this.peer_message_forwarder.clone();
+        //loop {
```
arminsabouri: Nit: remove commented out code

crates/consensus/authority/src/signing.rs
```diff
@@ -144,43 +165,93 @@ where
...
+    /// Checks if a signing session exists or not yet
+    pub(crate) fn signing_session_exists(&mut self, session_id: [u8; 32]) -> bool {
+        self.signing_sessions.contains_key(&session_id)
+    }
+
+    /// Inserts a new signing session
+    pub(crate) fn insert_new_signing_session(
         &mut self,
         session_id: [u8; 32],
+        coordinator_id: u64,
```
arminsabouri: nit: is this an index into the fed list? if so it should be `coordiantor_idnex`

crates/consensus/authority/src/signing.rs
```diff
@@ -504,13 +614,53 @@ where
...
+        if coordinator.is_some() {
+            error!(">>>>>>>>>>> A non-coordinator is trying to (re)initiate a signing process!");
+            return Ok(());
+        }
+        let (start, end, time_passed, time_remaining) =
+            self.get_inturn_interval_for_coordinator(None);
+
+        // try to find the signing session in cache in case it was re-triggered by the same
+        // coordinator
+        match self.get_signing_session(session_id) {
```
arminsabouri, line 629: What actually causes this to be re-triggered. IIRC the block builder just waits 60s for the current signing session to pass?

arminsabouri (commented):
Awesome work. Just had a few questions

crates/consensus/authority/src/signing.rs
BatiGencho: check the logic here: https://github.com/botanix-labs/Macbeth/pull/168/files#diff-f459d49b88e4803ede826537533f5067794f1d39165d9791add7a6bb6b3ac777R1017-R1067

if the current coordinator is still in turn and it decides there is time, it would retry it.

crates/consensus/authority/src/signing.rs
BatiGencho: of course, fixed

crates/net/network/src/frost/protocol.rs
BatiGencho: sure, done

.github/workflows/unit.yml
```diff
@@ -20,52 +20,6 @@ concurrency:
...
-      - name: Install Protobuf Compiler
-        run: sudo apt-get update && sudo apt-get install -y protobuf-compiler
-      - uses: dtolnay/rust-toolchain@stable
-      - uses: taiki-e/install-action@nextest
-      - uses: Swatinem/rust-cache@v2
-        with:
-          cache-on-failure: true
-      - name: Run e2e tests
-        run: |
-          make start-test-suite
```
arminsabouri: Did we mean to delete this? Is this being run somewhere else?

bin/test-suite/src/suite/consensus/frost/test_frost_e2e_edge_cases.rs
```diff
@@ -0,0 +1,294 @@
...
+
+    // now disconnect the peers of fed member 0
+    test_fed_members.get(&0).cloned().unwrap().send_test_signal(TestSignal::DisconnectAll());
+
+    // wait for a new epoch to start
+    tokio::time::sleep(Duration::from_secs(10)).await;
+
+    // now reconnect the peers of fed member 0
+    test_fed_members.get(&0).cloned().unwrap().send_test_signal(TestSignal::ReconnectAll());
+    tokio::time::sleep(Duration::from_secs(3)).await;
```
arminsabouri, line 240: What are we testing here? That we can perform a signing round without a peer being connected?

arminsabouri: It would be helpful to have a comment above the test function describing what exact edge case we are testing

arminsabouri:
Don't mean to be a commit Nazi. But the commits here could probably be squashed into one or maybe two 
- signing edge cases
- Fixed int pipeline

crates/botanix-lib/src/extra_data_header.rs
```diff
@@ -368,11 +368,28 @@ mod tests {
         );
         let mut buf: Vec<u8> = vec![];
         header.encode_into_without_signature(&mut buf).unwrap();
-        println!("{:?}", buf);
-        // serialize the same header
-        let serialized =
-            ExtraDataHeader::deserialize(&mut buf.as_slice()).expect("Deserialization");
-        assert_eq!(serialized, header);
+        // Check version
```
arminsabouri, line 371: I simplified this test in a earlier commit b/c I don't think doing exact bit by bit matching really got us much.
I think testing if something deserializes and serializes to the same thing is suffecient.
Lmk if you disagree

.github/workflows/unit.yml
arminsabouri: I just saw this was moved to unit.yml
Closing

arminsabouri (approved):

bin/test-suite/src/suite/consensus/frost/test_frost_e2e_edge_cases.rs
BatiGencho: Sure, I will add more comments in the next commit!

crates/botanix-lib/src/extra_data_header.rs
BatiGencho: I totally agree, all good!

PR #176, opened 2024-04-25
fix: only apply botanix block producer fee split if one is provided

arminsabouri:
Ef state tests are failing b/c block_builder_address is expected to exist but outside PoA consensus it will not


scottmillner (approved):
thanks for catching that!

BatiGencho (approved):

PR #177, opened 2024-04-26
WIP: non-federation nodes

scottmillner:
Still a few things need to do

- create doc on how to run non fed node
- make sure all tests pass

Will open a new pr once ci pipeline branch is merged since its commits will be squashed
but good to get review started here

Will link to this pr once new one is opened.


crates/consensus/authority/src/builder.rs
```diff
@@ -136,10 +136,14 @@ where
             })?
             .expect("authority signer list in epoch block");
 
-        let signer_index = authorities.iter().position(|a| *a == sk.public_key(&secp));
+        // authority length represents a non federation node since it would be out of bounds
+        let mut signer_index = Some(authorities.len() + 1);
+        if btc_server.is_some() {
```
BatiGencho, line 141: is this what the non-fed nodes have different to fed nodes ?

bin/reth/src/poa_builder.rs
```diff
@@ -415,7 +430,7 @@ impl<DB: Database + DatabaseMetrics + DatabaseMetadata + 'static> NodeBuilderWit
                 head,
                 &self.data_dir,
                 Some(Box::new(block_import)),
-                Some(frost_config.clone()),
+                frost_config.clone(),
```
BatiGencho, line 433: So I guess you are distinguishing b/n fed and non-fed mode based on the Option type of frost config ? Cannot we also do that based on the config boolean ?

bin/test-suite/src/suite/consensus/rpc_node/rpc_node.rs
```diff
@@ -0,0 +1,229 @@
...
+                    .await;
+            }
+        }));
+
+        Ok(())
+    }
+}
+
+#[allow(clippy::cast_possible_truncation)]
+pub async fn create_rpc_node(
```
BatiGencho, line 199: just wondering if we cannot merge that create func with the already existing one and just bring into scope the boolean based on which we will have a fed and a non-fed mode ?

PR #178, opened 2024-04-26
botanix-lib: Fix peg contract test

stevenroose:
It was changed in the impl but not accordingly changed in the test. I just copied the impl to the test.


arminsabouri (approved):

arminsabouri:
Included this fix in #179 
thanks again for fixing this

PR #180, opened 2024-04-29
Scott/392 tracking non federation nodes

scottmillner:
ive added comments from previous pr: https://github.com/botanix-labs/Macbeth/pull/177


crates/consensus/authority/src/builder.rs
```diff
@@ -136,10 +136,14 @@ where
             })?
             .expect("authority signer list in epoch block");
 
-        let signer_index = authorities.iter().position(|a| *a == sk.public_key(&secp));
+        // authority length represents a non federation node since it would be out of bounds
+        let mut signer_index = Some(authorities.len() + 1);
+        if btc_server.is_some() {
```
scottmillner: is this what the non-fed nodes have different to fed nodes ? (@BatiGencho comment)

bin/reth/src/poa_builder.rs
```diff
@@ -415,7 +430,7 @@ impl<DB: Database + DatabaseMetrics + DatabaseMetadata + 'static> NodeBuilderWit
                 head,
                 &self.data_dir,
                 Some(Box::new(block_import)),
-                Some(frost_config.clone()),
+                frost_config.clone(),
```
scottmillner, line 439: So I guess you are distinguishing b/n fed and non-fed mode based on the Option type of frost config ? Cannot we also do that based on the config boolean ? (@BatiGencho comment)

bin/test-suite/src/suite/consensus/rpc_node/rpc_node.rs
```diff
@@ -0,0 +1,229 @@
...
+                    .await;
+            }
+        }));
+
+        Ok(())
+    }
+}
+
+#[allow(clippy::cast_possible_truncation)]
+pub async fn create_rpc_node(
```
scottmillner, line 202: just wondering if we cannot merge that create func with the already existing one and just bring into scope the boolean based on which we will have a fed and a non-fed mode ? (@BatiGencho comment)

crates/consensus/authority/src/builder.rs
scottmillner: @BatiGencho 
For that specific change (`authorities.len() + 1`):
- This was so I didn't have to make `signer_index` an `Option` since non-fed nodes aren't signers
- So i just chose to make it an index which is not in the authorities list (out-of-bounds)

Difference with non-fed nodes
- No frost flags
- No btc server flag
- Uses a `trusted-peers` flag (which is list of fed nodes): not needed in `test-suite`

bin/reth/src/poa_builder.rs
scottmillner: @BatiGencho 
I'm not using frost to distinguish between types of nodes.
Here I'm using the boolean to set the correct Frost config: None or Some(frost_config)

In other parts of the code where the config boolean isn't available, im using `btc_server` to distinguish between nodes.  If a node has None for `btc_server`, then it's a non-fed node and vice versa

bin/test-suite/src/suite/consensus/rpc_node/rpc_node.rs
scottmillner: @BatiGencho 
yeah i started doing it that way but then broke it out into a separate function
the main reason i did this:
- `poa_node` has `create_poa_federation_members` but not a separate function to create just one node.
- we only need to create one non-fed node
-  so i could have pulled out the logic in the for loop in `create_poa_federation_members` that creates several nodes into a separate function (`create_node` for example) then refactored that to create either a fed node or non-fed node based on the config boolean
- it ended up being cleaner to just create a `create_rpc_node` method...at least for me it seems cleaner

BatiGencho: ok, that is all fine, was just asking.

bin/reth/src/poa_builder.rs
BatiGencho: I am ok with all of that, but pls make sure you leave comments in the code and maybe in the func annotations so we know where the diffs are.

crates/consensus/authority/src/builder.rs
BatiGencho: Is not this going to clash at a given point if we have a growing federation - then there will be at a given point  `(authorities.len() + 1)` signers and even more ?

BatiGencho: Not sure if it makes sense to wrap the btc server in some sort of enum which is more clearly defined. e.g.:

enum BtcServerType {
  FedNode(BtcSever).
  NonFedNode(()),
}

... just suggesting, but it might not make much sense here ?

scottmillner: > Is not this going to clash at a given point if we have a growing federation - then there will be at a given point `(authorities.len() + 1)` signers and even more ?

`authorities.len()` are the signers so `authorities.len() + 1` will never cause signers to clash

and non-fed nodes aren'ts signers so their index doesn't matter as long as it's outside the range of signer indexes

scottmillner: > Not sure if it makes sense to wrap the btc server in some sort of enum which is more clearly defined. e.g.:
> 
> enum BtcServerType { FedNode(BtcSever). NonFedNode(()), }
> 
> ... just suggesting, but it might not make much sense here ?

- i pushed a commit that makes what's happening more clear by:
at the top of a method/file i use `let is_fed_node = btc_server.is_some();`
and then just use `is_fed_node` so it's clear how that was determined and that the code path is forking based on `is_fed_node`


if the group (@0xBEEFCAF3) prefers the enum, i'll refactor it to do that

bin/reth/src/poa_builder.rs
scottmillner: yes, good point...i should've been more clear with comments and variable naming

please see the [commit](https://github.com/botanix-labs/Macbeth/pull/180/commits/cda1bc581f5f10c699ba94b273dbc67e6c6ad29a) where i add additional comments and the variable `is_fed_node` which is used as a conditional to make things more clear

hopefully that's an improvement, thanks for the feedback!

crates/consensus/authority/src/builder.rs
BatiGencho: No, thats fine, buddy. The name makes good sense to me, My whole point was not getting lost with Some and None when looking at the code, but rather knowing whether we are in fed node mode or not by looking at a variable e.g. Good work!

BatiGencho: Reg, the authorities.len() + 1 its all fine, but then I wonder if we have 3 feds and 5 non-fed just providing rpc. Should we provision for that also ?

scottmillner: i realized a previous comment of mine was incorrect: i was thinking of the test-suite
here is a clearer comment

the non-fed nodes can all have the same signer index:
- non-fed nodes aren't signers and their index only signifies that they aren't a signer
- this is shown by having a signer index outside the range of signers

i originally was going to hardcode non-fed indexes to `-1` but can't since the index type is `usize`
could refactor the type to `i16` and then set index to `-1` if that's more clear?

Makefile
```diff
@@ -432,3 +434,25 @@ start-poa-server-2:
...
+	-vvv \
+	--authrpc.jwtsecret "${NON_FED_1_DIR}/jwt.hex" \
+	--authrpc.addr "127.0.0.1" \
+	--authrpc.port 8553 \
+	--btc-network "${BITCOIND_NETWORK}" \
+	--bitcoind.url "${BITCOIND_URL}" \
+	--bitcoind.username "${BITCOIND_USER}" \
+	--bitcoind.password "${BITCOIND_PWD}" \
+	--p2p-secret-key "${NODE_2_DIR}/discovery-secret" \
+	--port 30305
```
arminsabouri: missing eof.

bin/reth/src/commands/node/mod.rs
```diff
@@ -261,10 +269,6 @@ mod tests {
     #[test]
     fn parse_common_node_command_chain_args() {
         for chain in SUPPORTED_CHAINS {
-            // TODO remove this condition once we add botanix_testnet to the supported chains
-            if chain == &"botanix_testnet" {
-                continue;
-            }
```
arminsabouri: I added this b/c it was breaking some CLI arg unit tests for me. The reth_primitives parse from method calls some alloy crate to pull infomation about the chain. Since we are not listen on that crate yet this test and other ones like this fail.

bin/test-suite/src/suite/consensus/rpc_node/rpc_node.rs
```diff
@@ -0,0 +1,233 @@
...
+        command
+    }
+}
+
+impl RethNodeCommandConfig for NonFederationMemberTestConfig {
+    #[allow(clippy::unwrap_used)]
+    fn on_node_started<Reth: RethNodeComponents>(&mut self, components: &Reth) -> eyre::Result<()> {
+        it_info_print!("Engine started non federation task with index: ", self.index);
+
+        let _pool = components.pool();
```
arminsabouri: Nit: remove unused

bin/test-suite/src/suite/consensus/rpc_node/test_rpc_node.rs
```diff
@@ -0,0 +1,114 @@
...
+            .send_eoa(eoa_receiver, SEND_AMOUNT)
+            .await
+            .unwrap()
+            .unwrap();
+        it_info_print!("Eoa tx: {:?}", last_tx_hash);
+
+        // wait for next member to come inturn
+        tokio::time::sleep(Duration::from_secs(60)).await;
+    }
+    it_info_print!("All eoa transactions sent");
```
arminsabouri, line 79: Just for my own undertanding. Are these N different Eoa txs? Or the same?
If they are the same should we rely on the gossip?

bin/test-suite/src/suite/consensus/rpc_node/test_rpc_node.rs
```diff
@@ -0,0 +1,114 @@
...
+        &rpc_node.secret_key,
+        ethers::core::types::Address::random(),
+    )
+    .await;
+
+    let rpc_latest_block_header = rpc_botanix_client.get_latest_block_hash().await.unwrap();
+    it_info_print!("RPC node latest header hash", rpc_latest_block_header);
+
+    assert_eq!(rpc_latest_block_header, fed_latest_header_hash);
+
```
arminsabouri, line 118: Could we add another test case here than we can submit a tx to a rpc node and have it confirmed. Then also check all the block hashes

template.env
```diff
@@ -1,10 +1,15 @@
...
 BITCOIND_NETWORK=regtest
 BITCOIND_URL=http://localhost:18443
 BITCOIND_USER=foo
-BITCOIND_PWD=bar
\ No newline at end of file
+BITCOIND_PWD=bar
+
+# trusted peers (federation nodes)
+TRUSTED_PEER_1=enode://abc@127.0.0.1:30303
+TRUSTED_PEER_2=enode://def@127.0.0.1:30304
```
arminsabouri: missing eof

arminsabouri (commented):
Very Clean! And nice test!
No show stoppers. I just have a few nits and noticed that a unit test change was reverted

bin/test-suite/src/suite/consensus/rpc_node/test_rpc_node.rs
scottmillner: they are N different txs: they have the same amount and receiver but they're distinct txs

- my intent is to create N number of new blocks (N is the number of authority members) and a new tx in each block

- so the non-fed node is several blocks behind and needs to sync to the tip which we check (a bit more robust then just syncing one block behind)

bin/test-suite/src/suite/consensus/rpc_node/test_rpc_node.rs
scottmillner: yes sir and done!

Makefile
scottmillner: done

bin/reth/src/commands/node/mod.rs
scottmillner: ah my bad...done

bin/test-suite/src/suite/consensus/rpc_node/rpc_node.rs
scottmillner: done

template.env
scottmillner: done

bin/reth/src/poa_builder.rs
```diff
@@ -161,14 +161,19 @@ impl<DB: Database + DatabaseMetrics + DatabaseMetadata + 'static> NodeBuilderWit
...
-        let btc_server_client = BtcServerExtendedClient::new(
-            self.config.rpc.btc_server.clone(),
-            Some(jwt_secret.clone()),
-        )
-        .await
-        .expect("cannot create btc_server");
-        info!(target: "reth::cli", "Btc server connected");
+        // Connect to btc signining server if in federation mode
+        let mut btc_server_client = None;
+        if self.config.federation_mode {
```
stevenroose: You can remove the `mut` here by doing
```rust
let btc_server_client = if self.config.federation_mode {
    ....
    Some(...)
} else {
    None
};
```
It's generally nicer to not have variables be mutable when it's not necessary :)

bin/reth/src/poa_builder.rs
```diff
@@ -386,19 +391,29 @@ impl<DB: Database + DatabaseMetrics + DatabaseMetadata + 'static> NodeBuilderWit
...
-        let (authority_index, authorities) = get_authority_signer_index(
-            blockchain_db.clone(),
-            Arc::clone(&self.config.chain),
-            secp256k1::Secp256k1::new(),
-            network_sk,
-        )
-        .expect("Failed to get authority index");
+        // create frost config if in federation mode
+        let mut frost_config = None;
+        if self.config.federation_mode {
```
stevenroose: Same `mut` thing here.

crates/consensus/authority/src/block_fetcher.rs
```diff
@@ -141,21 +141,27 @@ where
                 continue;
             }
 
-            if storage.aggregate_public_key.is_none() {
-                warn!(target: "consensus::authority", "Do not have aggregate public key in memory, skipping block import");
-                continue;
+            let mut botanix_consensus_pkg = None;
+            if self.btc_server.is_some() {
```
stevenroose: Same `mut` thing here. Btw, do you use `if self.btc_server.is_some()` here on purpose? There is no more explicit flag available at this point then?

arminsabouri (dismissed):

arminsabouri:
Looks great! I think there are some conflicts and CI should pass then we are good merge

crates/consensus/authority/src/block_fetcher.rs
```diff
@@ -232,6 +252,8 @@ where
                                 .collect();
                             let res = self
                                 .btc_server
+                                .clone()
+                                .expect("btc_server exists")
```
stevenroose, line 257: If the `if header.is_poa_epoch()` wasn't there, you should fix this by doing `if let Some(ref btc) = self.btc_server {` but right now Rust doesn't support things like 
```rust
if let Some(ref btc) = self.btc_server && header.is_poa_epoch() {
```
So I guess it's a choice between doing what you do now, or having two ifs inside one another
```rust
if let Some(ref btc) = self.btc_server {
    if header.is_poa_epoch() {
```
I have no real preference tbh.

crates/consensus/authority/src/builder.rs
stevenroose: Another `mut` case.

crates/consensus/authority/src/builder.rs
```diff
@@ -248,42 +252,53 @@ where
             tokio::sync::mpsc::unbounded_channel::<FrostNotificationMessage>();
         let (frost_task_notifications2_tx, frost_task_notifications2_rx) =
             tokio::sync::mpsc::unbounded_channel::<FrostNotificationMessage>();
+        // create frost and block production tasks if btc_server is available:
+        // only federation nodes will have btc_server
+        let mut frost_task = None;
+        let mut block_production_task = None;
+        if btc_server.is_some() {
```
stevenroose: Two more `mut` cases ;) tbh I hadn't anticipated this pattern would repeat so often :sweat_smile:

crates/rpc/rpc/src/eth/botanix_config.rs
```diff
@@ -211,10 +211,18 @@ impl Botanix {
...
-                .unwrap();
+        // Non-federation nodes will not have btc_server set
+        if self.botanix_rpc_config.btc_server.clone().is_none() {
+            return Err(GatewayAddressRPCError::InvalidParam(tonic::Status::invalid_argument(
+                "btc_server is not set",
+            )));
+        }
+
+        let mut client = client::BtcServerClient::connect(
+            self.botanix_rpc_config.btc_server.clone().expect("Valid URL").clone(),
```
stevenroose: nit: This expect just expects it to be non-empty, right?

bin/test-suite/src/suite/consensus/frost/botanix_client.rs
```diff
@@ -137,4 +142,17 @@ impl BotanixEthClient {
...
+
+    pub async fn get_latest_block_hash(&self) -> Result<ethers::core::types::H256, Error> {
+        let block_hash = self
+            .client
+            .get_block(BlockNumber::Latest)
+            .await
+            .map_err(Error::SignerMiddleware)?
+            .expect("block exists")
+            .hash
+            .expect("block hash exists");
```
stevenroose, line 154: Long live Rust and one-word lines!

```rust
        Ok(self.client.get_block(BlockNumber::Latest).await
            .map_err(Error::SignerMiddleware)?.expect("block exists")
            .hash.expect("block hash exists"))
```
:see_no_evil:

bin/test-suite/src/suite/consensus/frost/poa_node.rs
```diff
@@ -328,6 +331,10 @@ impl RethNodeCommandConfig for FederationMemberTestConfig {
     }
 }
 
+pub fn testnet_custom_chain() -> Arc<ChainSpec> {
+    BOTANIX_TESTNET.clone()
```
stevenroose: nit: Idk what you're going to use this `Arc` for, but you can avoid some potentially unnecessary shared memory accesses etc by just doing
```rust
pub fn testnet_custom_chain() -> &Arc<ChainSpec> {
    &*BOTANIX_TESTNET
```
In that case if the caller wants an `Arc` they can move around, they can clone themselves.

BatiGencho (dismissed):

bin/reth/src/poa_builder.rs
scottmillner: ah yep, thanks man
ill remember that moving forward

ack the other comments for same pattern

crates/consensus/authority/src/block_fetcher.rs
scottmillner: yeah, don't have the flag at this point

PR #182, opened 2024-05-06
Arm/tests/pbft state machine

arminsabouri:
To make these tests possible I had to define a trait `ToFrostMan` -> to frost manager. This is so I could stub it out. Majority of the changes to other files come from is replacing the FrostHandle struct with a trait.
I might need to do that for storage at some point. RIP.


BatiGencho:
Renaming makes total sense! The rest is great too !

BatiGencho (approved):

PR #185, opened 2024-05-08
Fix/arm/binding auth signature

arminsabouri:
Alot of changes here was moving the header ext trait and extra data header utils to reth_pritimitives to avoid cyclical dependencies. 
By update the block hash method directly we can essentially remove block signatures from the block hash calculcation.
This has two affects:
1. block signers will all sign the same sighash regardless of who adds new signatures
2. signatures can be re-ordered without affecting the blockhash


crates/primitives/src/header_ext.rs
```diff
@@ -89,6 +89,7 @@ impl HeaderExt for Header {
     }
 
     /// Provides block hash without extra data header bytes
+    /// Use this method instead of `hash_slow` to get an error if the extra data header is invalid
     fn segregated_signature_block_hash(&self) -> Result<B256, ExtraDataHeaderDeserializeError> {
```
stevenroose, line 93: Isn't this method removing the signatures twice at this point? I guess it's a nit, because it doesn't matter, but it's weird.

stevenroose (approved):
Looks good. One nit and other than that just two places where you use "affects" (edit: one might be in the MR description) ;) But you can blame English because you also correctly use "affecting" :shrug:

crates/primitives/src/header_ext.rs
arminsabouri: yeah thats a good point. Hash slow doesnt err out if the edh is invalid. Thats the only reason I kept this.
I think its fine.... it is certainly weird

arminsabouri: Should be effect lol

arminsabouri: fixed my broken english

PR #188, opened 2024-05-15
added templatable genesis config

BatiGencho:


chain.template.toml
```diff
@@ -0,0 +1,21 @@
+name="botanix_testnet"
+
+# >>>>>>>>>>> federation members secret keys
```
arminsabouri:
```suggestion
# >>>>>>>>>>> federation members public keys
```

BatiGencho: fixed

crates/node-core/src/args/genesis_args.rs
```diff
@@ -0,0 +1,72 @@
...
+    /// The account code (if any)
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub(crate) code: Option<String>,
+}
+
+#[derive(Clone, Debug, Deserialize)]
+#[serde(deny_unknown_fields, rename_all = "kebab-case")]
+pub(crate) struct FedMemberPubKey {
+    /// The pub key of the member
+    pub(crate) key: String,
```
arminsabouri, line 37: Super Nit:

```suggestion
    pub(crate) public_key: String,
```

arminsabouri (approved):
awesome stuff

PR #189, opened 2024-05-16
added cleanup tasks for stale entities

BatiGencho:


crates/consensus/authority/src/pbft.rs
```diff
@@ -20,11 +21,17 @@ use reth_primitives::{
...
     time::Duration,
 };
-use tokio::sync::mpsc::{error::SendError, UnboundedSender};
+use tokio::sync::{
+    mpsc::{error::SendError, UnboundedSender},
+    RwLock,
+};
 use tracing::{debug, error, info, warn};
 
+const BLOCK_TIME_DURATION_SECS: u64 = 1 * 10;
```
arminsabouri: 10 seconds? I think we are using 60 second blocks right now.
In the near future this should all get defined in chainspec somewhere so we dont have block times flying around everywhere

crates/consensus/authority/src/pbft.rs
```diff
@@ -113,14 +120,63 @@ impl<F: ToFrostManager> PbftStateMachine<F> {
...
+        let pre_commitments: Arc<RwLock<BTreeMap<BlockHash, HashSet<PeerId>>>> =
+            Arc::new(RwLock::new(BTreeMap::new()));
+        let sealed_blocks_clone = Arc::clone(&sealed_blocks);
+        let precommitments_clone: Arc<
+            RwLock<
+                BTreeMap<
+                    reth_revm::primitives::FixedBytes<32>,
+                    HashSet<reth_revm::primitives::FixedBytes<64>>,
+                >,
+            >,
```
arminsabouri: Do you need to define the whole type here? Shouldnt it be infered?

crates/consensus/authority/src/pbft.rs
```diff
@@ -113,14 +120,63 @@ impl<F: ToFrostManager> PbftStateMachine<F> {
...
+        let sleep_duration = Duration::from_secs(2 * BLOCK_TIME_DURATION_SECS);
+        tokio::spawn(async move {
+            loop {
+                // find stale tx hashes
+                let guard = sealed_blocks_clone.read().await;
+                let stale_hashes = guard
+                    .values()
+                    .cloned()
+                    .filter_map(|sealed_block| {
+                        if sealed_block.timestamp < unix_timestamp() - sleep_duration.as_secs() {
```
arminsabouri: Have you considered using block height instead of timestamp?
i.e if block.height < best_height - 2 -> remove ?

crates/consensus/authority/src/signing.rs
```diff
@@ -18,10 +18,15 @@ use reth_network::frost::{
...
-use std::{collections::HashMap, time::Duration};
-use tokio::sync::mpsc::{error::SendError, UnboundedSender};
+use std::{collections::HashMap, sync::Arc, time::Duration};
+use tokio::sync::{
+    mpsc::{error::SendError, UnboundedSender},
+    RwLock,
+};
 use tracing::{error, info, warn};
 
+const BLOCK_TIME_DURATION_SECS: u64 = 1 * 10;
```
arminsabouri: Same comment as above?

crates/consensus/authority/src/pbft.rs
arminsabouri: In the forking logic branch i have added the client into this struct so getting the best block height should be easy

arminsabouri (commented):
I had a general comment about block height vs timestamp

crates/consensus/authority/src/pbft.rs
stevenroose: Exactly, like above. If you prefer to be more explicit, please don't use 7 lines for a type indication..

crates/consensus/authority/src/pbft.rs
scottmillner: i have a preference of height instead of timestamp as well...in general prefer not to use timestamps if we don't have to

crates/consensus/authority/src/pbft.rs
```diff
@@ -113,28 +124,88 @@ impl<F: ToFrostManager> PbftStateMachine<F> {
...
+                    .cloned()
+                    .filter_map(|sealed_block| {
+                        let tip = client_clone.canonical_tip();
+                        let best_block_height = client_clone
+                            .block_by_number(tip.number)
+                            .ok()
+                            .flatten()
+                            .map(|b| b.header.number)
+                            .unwrap_or_default();
+                        let best_block_height = (best_block_height - 2).max(0);
```
stevenroose: can this go negative? if the type is unsigned here, this might panic and you should use `best_block_height.saturating_sub(2)` to get the same result.

stevenroose: And frankly block heights should be unsigned..

stevenroose (changes requested):
I looked at the code, but I have no idea what "stale txs", "stale blocks" and "sealed blocks" are..

crates/consensus/authority/src/pbft.rs
```diff
@@ -87,54 +97,113 @@ impl PbftState {
...
+                                None
+                            }
+                        })
+                        .collect::<Vec<_>>();
+                    drop(guard);
+
+                    // remove stale sealed blocks
+                    let mut guard = sealed_blocks_clone.write().await;
+                    guard.retain(|_, sealed_block| {
+                        sealed_block.timestamp >= unix_timestamp() - sleep_duration.as_secs()
```
arminsabouri: Leaving my original comment here again. Can we use block heights here instead of timestamps?
i.e if best_block.height - sealed_block.height  > 2 -> clean up

crates/consensus/authority/src/pbft.rs
```diff
@@ -537,6 +646,349 @@ mod tests {
         }
     }
 
+    #[derive(Clone)]
+    struct ClientMock {}
+
+    impl ReceiptProviderIdExt for ClientMock {}
```
arminsabouri: This is going to conflict with 
#191 that uses mockEth client defined in the reth utils

stevenroose (approved):

arminsabouri (approved):

PR #190, opened 2024-05-16
added time-based test logic and pbft test cases mechanisms

BatiGencho:
- added time-based validity checks
- added the ability to await all btc servers before running tests
- added a pbft e2e test (currently commented out)
- added the ability to signal the exact moment when signing has finished via session ids
- added better output and display traits masking long hex blocks
- reworked the time logic to be based on secs than on minutes and thus be more precise
- increased grpc timeout, ping and healthcheck durations
- added 2 new btc server grpc endpoints
- fixes to test-suite
- e2e tests renaming
- fmt, cleanup

PR #191, opened 2024-05-16
Arm/feature/bft forks

arminsabouri:
This PR:
* Moved extra data header to reth primitives crate
* Moves validate sigs helpers to header ext trait
* Validate block should validate that a block signer will not sign for a fork that is deeper than 1 block!
* Configure blockchain tree engine to only allow for reorgs of 1 depth
* Started mocking out block chain tree ext for mock eth provider

### DIFF
--- a/bin/reth/src/poa_builder.rs
+++ b/bin/reth/src/poa_builder.rs
@@ -106,6 +106,9 @@ pub async fn launch_poa_from_config<E: RethCliExt>(
     }
 }
 
+// PBFT consensus allows for at least block to be reorged
+const POA_MAX_REORG_DEPTH: u64 = 1;
+
 /// A version of the [NodeConfig] that has an installed database. This is used to construct the
 /// [NodeHandle].
 ///
@@ -366,7 +369,8 @@ impl<DB: Database + DatabaseMetrics + DatabaseMetadata + 'static> NodeBuilderWit
         let evm_config = EthEvmConfig::default();
 
         // configure blockchain tree
-        let tree_config = BlockchainTreeConfig::default();
+        // Rest of these block chain tree configs are defaults
+        let tree_config = BlockchainTreeConfig::new(POA_MAX_REORG_DEPTH, 65 /* max_blocks_in_chain */, 256/* num_of_additional_canonical_block_hashes*/, 200 /* max_unconnected_blocks */);
         let tree = self.config.build_blockchain_tree(
             provider_factory.clone(),
             consensus.clone(),
--- a/bin/reth/src/poa_builder.rs
+++ b/bin/reth/src/poa_builder.rs
@@ -517,6 +517,7 @@ impl<DB: Database + DatabaseMetrics + DatabaseMetadata + 'static> NodeBuilderWit
             network_sk,
             None,
             network.clone(),
+            network_client.clone(),
             frost_handle.clone(),
             block_import_rx,
             executor.clone(),
--- a/bin/reth/src/poa_builder.rs
+++ b/bin/reth/src/poa_builder.rs
@@ -370,7 +370,12 @@ impl<DB: Database + DatabaseMetrics + DatabaseMetadata + 'static> NodeBuilderWit
 
         // configure blockchain tree
         // Rest of these block chain tree configs are defaults
-        let tree_config = BlockchainTreeConfig::new(POA_MAX_REORG_DEPTH, 65 /* max_blocks_in_chain */, 256/* num_of_additional_canonical_block_hashes*/, 200 /* max_unconnected_blocks */);
+        let tree_config = BlockchainTreeConfig::new(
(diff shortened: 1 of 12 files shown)


crates/consensus/authority/src/pbft.rs
```diff
@@ -200,6 +272,79 @@ impl<F: ToFrostManager> PbftStateMachine<F> {
...
+        // or a fork
+        if block_to_sign.parent_hash == best_block_hash {
+            return Ok(());
+        } else if best_block.header.number == 0 {
+            // Case where the best block is the genesis block
+            // This should never happen
+            return Ok(());
+        } else {
+            // Somehow we have the parent block but its not the current canon chain?
+            // This should not happen
```
scottmillner, line 386: can we have the parent block, it is in the canonical chain, but it isn't `best_block_hash` bc it's deeper than that?

crates/consensus/authority/src/pbft.rs
```diff
@@ -200,6 +272,79 @@ impl<F: ToFrostManager> PbftStateMachine<F> {
...
+        }
+
+        // Check if we are building on a block that is in the canonical chain
+        // or a fork
+        if block_to_sign.parent_hash == best_block_hash {
+            return Ok(());
+        } else if best_block.header.number == 0 {
+            // Case where the best block is the genesis block
+            // This should never happen
+            return Ok(());
```
arminsabouri, line 383: TODO maybe we should error out here and not sign

crates/consensus/authority/src/pbft.rs
arminsabouri: Its possible but a non-byzantine node would never propose a block like that.

bin/reth/src/poa_builder.rs
```diff
@@ -366,7 +369,13 @@ impl<DB: Database + DatabaseMetrics + DatabaseMetadata + 'static> NodeBuilderWit
...
 
         // configure blockchain tree
-        let tree_config = BlockchainTreeConfig::default();
+        // Rest of these block chain tree configs are defaults
+        let tree_config = BlockchainTreeConfig::new(
+            POA_MAX_REORG_DEPTH,
+            65,  /* max_blocks_in_chain */
+            256, /* num_of_additional_canonical_block_hashes */
+            200, /* max_unconnected_blocks */
+        );
```
BatiGencho, line 378: Just a comment though - I remember the last 3 params did not have any effect when I was dealing a few months ago with the chain forks. Just fyi

crates/primitives/src/header_ext.rs
```diff
@@ -130,6 +192,17 @@ impl HeaderExt for Header {
         Ok(())
     }
 
+    /// Validate timestamp
+    fn validate_timestamp(&self, current_timestamp: u64) -> Result<(), ValidateInturnError> {
+        // Time stamp should be less that or greater than by 2 seconds
```
BatiGencho, line 197: In my MR: https://github.com/botanix-labs/Macbeth/pull/190/files this has been checked in a different way. We have to merge these eventually.

crates/consensus/authority/src/block_fetcher.rs
```diff
@@ -266,6 +278,40 @@ where
                         }
                     }
 
+                    // Need to decide if we accepting a forked block or not
```
arminsabouri, line 281: Note: this is not tested yet

crates/consensus/authority/src/block_fetcher.rs
```diff
@@ -266,6 +278,40 @@ where
...
+                    // There is a garuntee a quorum of signers will not sign an invalid fork
+                    let tip = storage.client.best_block_number().expect("best block exists");
+                    let best_block = storage
+                        .client
+                        .block_by_number(tip)
+                        .expect("best block exists")
```
BatiGencho, line 287: There is quite a few expects in here. Should we not just say that if .is_err() { continue } for each of these in the validation logic ?

bin/reth/src/poa_builder.rs
arminsabouri: Yeah I agree. The configs passed to the BTE largely have no affect

crates/consensus/authority/src/block_fetcher.rs
arminsabouri: the three expects here are for getting the best_block / tip. The DB should always have a tip. Should we not panic if one can't be found?

BatiGencho: I agree with you. What if its the genesis ? (ignore the comment if its not relevant)

BatiGencho (approved):

arminsabouri:
>I agree with you. What if its the genesis ? (ignore the comment if its not relevant) 
if you only have block 0. The tip should then just be genesis.

PR #205, opened 2024-05-25
workflow cleanup in progress

birozuru:
Workflow cleanup in progress. This should fix on-going issues and provision stable pipeline for the future moving forward.


BatiGencho (dismissed):

arminsabouri:
Note: lets squash merge this PR

.github/workflows/integration.yml
```diff
@@ -63,49 +65,51 @@ jobs:
...
-          --bitcoind-user ${{env.BITCOIND_USER}} \
-          --bitcoind-pass ${{ env.BITCOIND_PASS }} \
-          --fall-back-fee-rate-sat-per-vbyte 5 &
-          pid=$! 
-          sleep 150
-          echo "Killing BTC Server"
-          kill -9 $pid
+  # This performs the same test as the e2e job, but only on btc_server
+  # test:
+  #   name: test / ethereum
```
scottmillner: the name of this test always confuses me

it only spins up btc-server but doesn't actually run any tests, right?

arminsabouri: Yeah good point. We can probably remove this commented out workflow

.github/workflows/integration.yml
```diff
@@ -63,49 +65,51 @@ jobs:
...
+  #         --bitcoind-user ${{env.BITCOIND_USER}} \
+  #         --bitcoind-pass ${{ env.BITCOIND_PASS }} \
+  #         --fall-back-fee-rate-sat-per-vbyte 5 &
+  #         pid=$! 
+  #         sleep 150
+  #         echo "Killing BTC Server"
+  #         kill -9 $pid
 
   sync:
     name: sync / 100k blocks
```
scottmillner, line 82: same here as above, im confused, what am i missing?

both the `test / ethereum` and `sync / 100k blocks` jobs just run this command which starts up btc-server and nothing else: `/home/runner/.cargo/bin/cargo run --bin btc-server`

.github/workflows/integration.yml
birozuru: yeah that's why I commented it out cause it only spins up the btc-server

PR #206, opened 2024-05-28
test: wait for btc server to be online and connectable before continu…

arminsabouri:
…ing with test suite

### DIFF
--- a/bin/test-suite/src/suite/consensus/mod.rs
+++ b/bin/test-suite/src/suite/consensus/mod.rs
@@ -43,7 +43,7 @@ fn kill_child_processes_at_port(index: u16) {
 pub struct ConsensusIntegrationTestSuite {
     pub timeout: Duration,
     pub global_context: Arc<GlobalContext>,
-    pub outcome: Outcome,
+    pub outcomes: Vec<Outcome>,
     pub local_context: LocalContext,
 }
 pub struct LocalContext {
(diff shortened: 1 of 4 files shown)


scottmillner (approved):

bin/test-suite/src/suite/consensus/mod.rs
```diff
@@ -144,8 +145,40 @@ impl Suite for ConsensusIntegrationTestSuite {
         // create new context
         self.local_context.btc_servers = Some(spawn_n_btc_servers(self.global_context.clone()));
 
-        // let servers come up
-        tokio::time::sleep(tokio::time::Duration::from_secs(15)).await;
+        // try to connect to each btc server before moving on
+        let mut tries = 5;
+        let mut successes = 0;
+        loop {
```
BatiGencho, line 151: something feels weird about this loop. Do we let 5 tries for all btc servers altogether or allow each of them to fail up to max. 5 times before we deem it failed ? I have the feeling now that the successes could be incremented from different loop iterations, meaning one where a few only succeeded and one where another bunch only succeeded, but not all at the same time which is what we want. Or am I not getting it correctly ?

arminsabouri: It allows for 5 total tries across all of the btc server instances.

arminsabouri: No the only way we break out of this loop is if the number of success is equal to the total number of instances

PR #211, opened 2024-05-29
refactor: check witness data is not empty

scottmillner:


crates/consensus/authority/src/lib.rs
```diff
@@ -373,8 +373,15 @@ where
             .unwrap();
         header.state_root = state_root;
 
-        // TODO(scott): if pegouts are pending but no witness data is provided, we should fail
-        // consensus validation
+        // fail if witness data is empty
+        // witness data will be None if no pegouts are being processed in this block
+        if let Some(witness_data) = witness_data {
+            if witness_data.is_empty() {
```
arminsabouri, line 379: Should this only fail if there are no pegouts? Do we need to check that here?

scottmillner: yes, only fail if no pegouts
i was thinking of this as an additional sanity check....block_builder calls this method via `build_and_validate_header` and should pass None if no pegouts but if somehow we passed it an empty array of pegouts, we shouldn't continue with the psbt...that's my thinking at least

arminsabouri (approved):

PR #213, opened 2024-05-31
 missing pegin/outs during block download

BatiGencho:


bin/btc-server/.gitignore
```diff
@@ -1 +1,2 @@
-db*/
\ No newline at end of file
+db*/
+Cargo.lock
```
arminsabouri: Since the output of reth and the btc server is a executable should we continue commiting the Cargo.lock?

arminsabouri: Also this seems like it should be its own commit or PR. Seems independent of the main purpose of this PR.

crates/consensus/authority/src/block_fetcher.rs
```diff
@@ -105,30 +115,46 @@ where
...
+
+            // check for missing blocks from the tip
+            let are_blocks_missing = !block.parent_hash.eq(&best_hash);
+            let mut blocks_to_sync: Vec<Block> = vec![];
+            if are_blocks_missing {
+                warn!(target: "consensus::authority", "Block fetcher is missing blocks. Catching up...");
+                let mut block_num_hash = block.parent_num_hash();
+                loop {
+                    if let Some(block) =
+                        BlockReader::block_by_hash(&storage.client, block_num_hash.hash)
```
arminsabouri: This is just pulling from the DB correct? What we probably want is to use the networkclient
The flow here should be 
1. Validate the announced block
2. Call FCU. FCU will fill the gap by sync'ing with a peer
3. Now we can use the DB client to pull the blocks that we are missing
4. Use the bloom filter utils to detect for pegins
5. Call BTC server to update utxo set

crates/consensus/authority/src/block_fetcher.rs
```diff
@@ -160,139 +186,159 @@ where
...
-                                &bundle_state,
-                                recent_bitcoin_block_height,
-                                is_testnet,
-                                self.btc_network,
-                            )
-                            .await
-                            {
-                                Ok(pegouts) => pegouts,
+            // execute the entirety of blocks
+            for block_to_sync in blocks_to_sync.iter() {
```
arminsabouri: Do we have to execute each block in the gap? IIRC We only should have to execute if there are pegins. and we should be able to detect that via the bloom filter

crates/consensus/authority/src/block_fetcher.rs
scottmillner: just acking i reviewed and the flow above make sense to me

will review again when updated

bin/btc-server/.gitignore
BatiGencho: removed

crates/consensus/authority/src/block_fetcher.rs
BatiGencho: fixed

crates/consensus/authority/src/block_fetcher.rs
BatiGencho: fixed

crates/consensus/authority/src/block_fetcher.rs
```diff
@@ -105,49 +123,75 @@ where
...
+
+            // check for missing blocks from the tip and use the network client to fetch them
+            let network_client = self.network_client.clone();
+            let are_blocks_missing = !block.parent_hash.eq(&best_hash);
+            let mut blocks_headers_to_sync: Vec<BlockNumHash> = vec![block.parent_num_hash()]; // add the incoming block first
+            if are_blocks_missing {
+                warn!(target: "consensus::authority", "Block fetcher is missing blocks. Catching up...");
+                let mut block_parent_num_hash = block.parent_num_hash();
+                loop {
+                    let block_header = network_client
```
arminsabouri: Per your comment above. FCU should fill the gap. So the missing blocks should be in canon order now in the DB. So you should be able to just read them via the client.

crates/consensus/authority/src/block_fetcher.rs
```diff
@@ -105,49 +123,75 @@ where
...
+                        .ok()
+                        .map(|val| val.1)
+                        .flatten();
+                    if let Some(block_header) = block_header {
+                        blocks_headers_to_sync
+                            .push(BlockNumHash::new(block_header.number, block_header.mix_hash)); // TODO: check if mix_hash is correct
+                        block_parent_num_hash = block_header.parent_num_hash();
+                        if block_parent_num_hash.number == best_block {
+                            break;
+                        }
```
arminsabouri: This is a another reason to just have the engine handle reading from p2p network:
A malicious peer could send your a series of a blocks that never link a canon chain and then we're stuck in this loop

BatiGencho: fixed. please review

crates/consensus/authority/src/block_fetcher.rs
BatiGencho: fixed. please review

crates/consensus/authority/src/block_fetcher.rs
```diff
@@ -105,49 +123,70 @@ where
...
+                continue;
+            }
+
+            // check for missing blocks from the tip and use the network client to fetch them
+            let network_client = self.network_client.clone();
+            let are_blocks_missing = !block.parent_hash.eq(&best_hash);
+            let mut blocks_headers_to_sync: Vec<BlockNumHash> = vec![];
+            if are_blocks_missing {
+                warn!(target: "consensus::authority", "Block fetcher is missing {} blocks. Catching up...", block.header.number - best_block);
+                for block_index in block.header.number..best_block {
```
scottmillner: ✅ but just had a question...

i might be thinking about this incorrectly but are descending ranges allowed in Rust?:
- isn't `block.header.number` > `best_block`?
- was thinking it would be `for block_index in best_block..block.header.number`

`best_block` is our tip and `!block.parent_hash.eq(&best_hash);` so the incoming block is either greater than our tip or part of an invalid sidechain if it's height is less than our tip?

scottmillner (dismissed):
🔥 🔥 🔥 💪

crates/consensus/authority/src/block_fetcher.rs
BatiGencho: Thank you so  much, I have fixed these issues now. Good catch!

scottmillner (dismissed):

scottmillner (dismissed):

crates/consensus/authority/src/utils.rs
```diff
@@ -460,46 +460,112 @@ where
...
+        Err(e) => {
+            error!("Error fetching block {}: {}", block, e);
+            return Err(EpochPegoutsError::FailedToFetchPegouts);
+        }
     }
 
     Ok(pegouts)
 }
 
+pub(crate) async fn has_block_pegins<Client>(
```
arminsabouri: Nit rename to `block_has_pegins`

crates/consensus/authority/src/utils.rs
```diff
@@ -460,46 +460,112 @@ where
...
+            return Err(EpochPegoutsError::FailedToFetchPegouts);
+        }
+        Err(e) => {
+            error!("Error fetching block {}: {}", block, e);
+            return Err(EpochPegoutsError::FailedToFetchPegouts);
+        }
+    }
+}
+
+pub(crate) async fn has_block_pegouts<Client>(
```
arminsabouri: Nit: rename to `block_has_pegouts `

crates/consensus/authority/src/block_fetcher.rs
arminsabouri: This still seems to be using the networkclient instead of db client

crates/consensus/authority/src/block_fetcher.rs
```diff
@@ -176,177 +231,204 @@ where
...
-                    if header.is_poa_epoch() && is_fed_node {
-                        // get the pegouts from during the epoch
-                        let past_pegouts = crate::utils::epoch_pegouts(best_block, &storage.client, self.btc_network,).await.map_err(|e| {
-                            error!(target: "consensus::authority", ?e, "Failed to get epoch pegouts");
-                            e
-                        }).unwrap();
-                        pegouts.extend(past_pegouts);
-                        let extra_data = ExtraDataHeader::deserialize(
-                            &mut header.extra_data.clone().to_vec().as_slice(),
+                match storage.execute_imported_block(
```
arminsabouri, line 301: Open question not really a request for change.
is there a `get_reciepts` ? or `get_logs` provider method that we can call instead of executing each block?
Maybe `pending_block_and_receipts()`

scottmillner: don't we want to execute each block to make sure it's a valid block?
when we are synced and just importing the single latest block, we still execute the block so shouldn't we do the same if there's a gap of blocks?

BatiGencho: I agree with @scottmillner

BatiGencho: apart from that `execute_imported_block` also does all the checks for block validity

crates/consensus/authority/src/utils.rs
BatiGencho: done

crates/consensus/authority/src/utils.rs
BatiGencho: done

crates/consensus/authority/src/block_fetcher.rs
arminsabouri: Doesnt the BTE already execute and run consenus checks?

crates/consensus/authority/src/block_fetcher.rs
BatiGencho: I was using the network client initially since the db client was not synced, but since we awre doing the sync first, I have now changed it it to storage.client.

crates/consensus/authority/src/block_fetcher.rs
arminsabouri: just following up on this thread

PR #217, opened 2024-06-02
Add trusted peers from chain.toml

scottmillner:


bin/reth/src/commands/poa/mod.rs
```diff
@@ -456,11 +461,35 @@ where {
...
+
+        // set trusted nodes with --trusted-peers flag or chain.toml
+        // --trusted-peers flag takes precedence
+        info!(target: "reth::cli", "Adding trusted nodes");
         if !node_config.network.trusted_peers.is_empty() {
-            info!(target: "reth::cli", "Adding trusted nodes");
             node_config.network.trusted_peers.iter().for_each(|peer| {
                 reth_config.peers.trusted_nodes.insert(*peer);
             });
+        } else {
```
arminsabouri: How do you feel about adding the fed pks regardless of the cli flags. 
that way users can add other trusted peers while also having their fed members automatically listed as peers

scottmillner: good point, done!

arminsabouri:
Screenshot 2024-06-03 at 3 11 30 PM

This may already be broken on the pbft feature branch.

bin/reth/src/commands/poa/mod.rs
```diff
@@ -456,12 +458,36 @@ where {
...
         if !node_config.network.trusted_peers.is_empty() {
-            info!(target: "reth::cli", "Adding trusted nodes");
             node_config.network.trusted_peers.iter().for_each(|peer| {
                 reth_config.peers.trusted_nodes.insert(*peer);
             });
         }
+
+        // add trusted nodes (federation members) with chain.toml
+        // assumes chain.toml is present at `bin/reth/chain.toml` based on Makefile command
+        let chain_path = PathBuf::from_str("chain.toml").unwrap();
```
BatiGencho: Lets catch the errors here and in case of one, printout an elegant message to the user.

bin/reth/src/commands/poa/mod.rs
```diff
@@ -456,12 +458,36 @@ where {
...
+            get_federation_pks_from_path(&chain_path).expect("federation keys to exist");
+        for fed_member in fed_members.iter() {
+            // don't add self
+            let self_peer_id = pk2id(&secret_key.public_key(SECP256K1));
+            let peer_id = pk2id(&fed_member.0);
+            if self_peer_id != peer_id {
+                let peer = NodeRecord::new(fed_member.1, peer_id);
+                reth_config.peers.trusted_nodes.insert(peer);
+            }
+        }
```
BatiGencho: For the sake of  brevity (as the mod.rs) is already huge, could me abstract away this loop in some for of a func and just call it directly ?

.gitignore
```diff
@@ -65,3 +65,6 @@ contracts/.env
 
 # Rust bug report
 rustc-ice-*
+
+# botanix federation chain toml
+chain.toml
```
BatiGencho, line 70: Do we really want to ignore changes to the chain.toml even for local testing ? What if we add or change stuff in it ?

bin/reth/src/commands/poa/mod.rs
BatiGencho:
```
match PathBuf::from_str("chain.toml") {
   Ok(chain_path) => chain_path,
  Err(err) => { .... eyre::error!...) }

}
```

BatiGencho:
@scottmillner @0xBEEFCAF3 Just wondering out of curiosity if we should not  pass the chain.toml as a path. I see Scotts hardcoding it above, but it might make sense to have it as an arg ?

arminsabouri:
> @scottmillner @0xBEEFCAF3 Just wondering out of curiosity if we should not pass the chain.toml as a path. I see Scotts hardcoding it above, but it might make sense to have it as an arg ?

Agreed. It would be nice if this was an arg

.gitignore
scottmillner: i was thinking we could just update `chain.template.toml` with any changes needed but im fine with tracking it's changes.... @0xBEEFCAF3 any preference?

bin/reth/src/commands/poa/mod.rs
scottmillner: done, thanks!

bin/reth/src/commands/poa/mod.rs
scottmillner: good idea, done, thanks!

scottmillner:
> @scottmillner @0xBEEFCAF3 Just wondering out of curiosity if we should not pass the chain.toml as a path. I see Scotts hardcoding it above, but it might make sense to have it as an arg ?

Are we saying we want a separate cli flag for the path? Reason I ask:

1. `--chain <path_to_chain.toml>` is consumed by `Cli::parse_args` so we don't have access to that path value in `PoaNodeCommand::execute()`.  But we would if we added the flag `--chain_toml_path` for example.  It is a bit redundant but probably the easiest.

arminsabouri:
> saying

Alternatively chain.toml could be in the same dir as data_dir

crates/consensus/authority/src/pbft.rs
```diff
@@ -365,7 +365,8 @@ where
         // if the suggested block is the canon tip there is no point to signing it again
         match self.client.is_canonical(block_hash) {
             Ok(false) => (),                                // continue
-            Err(ProviderError::BlockHashNotFound(_)) => (), // continue
+            Err(ProviderError::BlockHashNotFound(_)) => (), /* great block being proposed is not */
+            // canon
```
arminsabouri, line 369: Its wild that rust fmt decides this is the right formatting for this

arminsabouri: not a request for a change

arminsabouri (approved):

PR #223, opened 2024-06-05
Scott/439 test invalid pegin gets reverted

scottmillner:


crates/blockchain-tree/src/noop.rs
```diff
@@ -52,6 +52,15 @@ impl BlockchainTreeEngine for NoopBlockchainTree {
         ))
     }
 
+    fn insert_block_with_botanix_consensus_package(
+        &self,
+        block: SealedBlockWithSenders,
+        validation_kind: BlockValidationKind,
+        _botanix_consensus_pkg: Option<BotanixConsensusPackage>,
```
arminsabouri, line 59: Should this get passed into insert_block?

crates/consensus/beacon/src/engine/message.rs
```diff
@@ -151,6 +152,8 @@ pub enum BeaconEngineMessage<Engine: EngineTypes> {
         cancun_fields: Option<CancunPayloadFields>,
         /// The sender for returning payload status result.
         tx: oneshot::Sender<Result<PayloadStatus, BeaconOnNewPayloadError>>,
+        /// Optional BotaniX consensus package.
```
arminsabouri: Big X typo

crates/blockchain-tree/src/noop.rs
arminsabouri: Or should we be using `insert_block_with_botanix_consensus_package ` here ?

crates/interfaces/src/error.rs
```diff
@@ -60,21 +60,21 @@ impl From<FsPathError> for RethError {
 }
 
 // Some types are used a lot. Make sure they don't unintentionally get bigger.
-#[cfg(all(target_arch = "x86_64", target_pointer_width = "64"))]
-mod size_asserts {
-    use super::*;
+// #[cfg(all(target_arch = "x86_64", target_pointer_width = "64"))]
```
arminsabouri: Is this worth re-enabling? Maybe ticket out so we dont forget?

arminsabouri (commented):
Just had a few nits.
approach Ack! Nice work

crates/blockchain-tree/src/noop.rs
scottmillner: i added `insert_block_with_botanix_consensus_package` to `BlockchainTreeEngine` trait but in this implementation of the trait, it isn't needed. So i've made `insert_block_with_botanix_consensus_package` just a wrapper around `insert_block`...so basically this is the same as calling `insert_block`

This is the approach im taking anytime the method needs to be implemented only bc it's part of the trait but the functionality isn't actually needed by the struct...was a tradeoff off extending the trait

crates/consensus/beacon/src/engine/message.rs
scottmillner: thanks! it looks kinda cool tho with the big X 😆... and done!

crates/interfaces/src/error.rs
scottmillner: i had to disable this bc `static_assert_size!(BlockExecutionError, 56);` fails bc the error message ends up being too big bc i need to pass pegin data in the error

i can try and renable and increase the the allowed size just for `static_assert_size!(BlockExecutionError, 56);`

and done!

bin/test-suite/src/suite/consensus/invalid_transactions/test_invalid_pegin.rs
```diff
@@ -0,0 +1,220 @@
...
+    // status should be 0 (failure)
+    it_info_print!("Pegin Tx Receipt", tx_receipt);
+    assert!(tx_receipt.status.unwrap().is_zero());
+
+    // pegin address balance after pegin
+    let pegin_address_final_balance =
+        botanix_eth_client.get_botanix_balance(eth_pegin_address.as_str()).await.unwrap();
+    it_info_print!("Final pegin address balance", pegin_address_final_balance);
+
+    assert_eq!(pegin_address_initial_balance, pegin_address_final_balance);
```
arminsabouri, line 223: Good assertion! Is there anything else to assert about this account that would indicate a revert happened? Like nonce is different?

bin/test-suite/src/suite/consensus/invalid_transactions/test_invalid_pegout.rs
```diff
@@ -0,0 +1,60 @@
...
+    let pegout_amount = Amount::from_btc(0.5).unwrap();
+    let tx_receipt = botanix_eth_client
+        .burn(invalid_pegout_destination, pegout_data, pegout_amount.to_wei())
+        .await
+        .unwrap()
+        .unwrap();
+    it_info_print!("Pegout Tx Receipt: ", tx_receipt);
+
+    assert!(tx_receipt.status.unwrap().is_zero());
+
```
arminsabouri, line 58: Maybe we assert value is the same as before for this account?

scottmillner: here im using an invalid address so there will be no balance

bin/test-suite/src/suite/consensus/invalid_transactions/test_invalid_pegin.rs
scottmillner: good idea, done!

arminsabouri (approved):

PR #226, opened 2024-06-06
added a mechanism for node services to await full synhcronization

BatiGencho:


bin/reth/src/commands/poa/mod.rs
```diff
@@ -59,6 +59,8 @@ use tokio::{
 };
 use tracing::{debug, error, info};
 
+const LIVE_SYNC_TIMEOUT: u64 = 60;
```
arminsabouri: Eventually when we hit hundreds of thousands of blocks this may take hours or days even.
So should we have a timeout here? If there is a problem with syncing then we should be loud about it in the logs and hope that people will open up bug tickets

BatiGencho: Gotcha. I changed it now. What we could also add is better printout on how many blocks we need to sync with and time remaining estimate , but I prefer this is a separate ticket.

arminsabouri (dismissed):

arminsabouri (approved):

PR #231, opened 2024-06-07
fix: disable engine RPC if chainspec is botanix

BatiGencho:


bin/reth/src/commands/poa/mod.rs
```diff
@@ -813,17 +813,19 @@ where {
...
-                blockchain_db.clone(),
-                transaction_pool.clone(),
-                network_handle.clone(),
-                executor.clone(),
-                blockchain_db.clone(),
-                evm_config.clone(),
-            )
-            .await?;
+        if !(node_config.chain.as_ref().eq(BOTANIX_TESTNET.as_ref())) {
+            let _rpc_server_handles = node_config
```
arminsabouri: Would this not disable the whole rpc server? 
We want to just disable the engine api endpoints

BatiGencho: fixed as it should be!

crates/node/builder/src/rpc.rs
```diff
@@ -328,3 +336,93 @@ where
...
+            .start_rpc_server(
+                provider.clone(),
+                pool.clone(),
+                network.clone(),
+                executor.clone(),
+                provider.clone(),
+                evm_config.clone(),
+            )
+            .await?;
+        (rpc_server_handle, None)
```
arminsabouri, line 385: Am i reading this right? If chainspec != Botanix_testnet -> Do not start the auth server?

arminsabouri: Is the auth server synonymous with the engine api ?

crates/node/builder/src/rpc.rs
```diff
@@ -328,3 +336,93 @@ where
 
     Ok((handles, registry))
 }
+
+/// Launch the poa rpc servers.
+pub async fn launch_poa_rpc_servers<Provider, Pool, Network, Tasks, EvmConfig, EngineT>(
```
arminsabouri, line 341: Can we just modify the existing start rpc method. Any reason for a poa specific one?

arminsabouri (commented):
Approach ACK

crates/node/builder/src/rpc.rs
BatiGencho: yes, the auth server uses the engine api to expose a further set of eth_xxx rpc endpoints that we want to disable.

crates/node/builder/src/rpc.rs
BatiGencho: What do you mean ? There are two methods `start_rpc_server` and `start_auth_server`. There is no need to modify the  `start_rpc_server` since it will be started under all circumstances.  `start_auth_server`. is only needed for none-botanix networks. If you mean to modify the `launch_rpc_servers` that does not work as we dont have the node: Node any longer in the poa impls. The structure is a bit different here hence the need of  a new method.

arminsabouri (approved):

PR #232, opened 2024-06-10
Feature/arm/bitcoind toml

arminsabouri:
The last commit is not my favorite. I wanted to read config from a toml file and prioritize that. Then fall back to cli args if needed. Lmk if there is a cleaner way

### DIFF
--- a/bin/btc-server/src/main.rs
+++ b/bin/btc-server/src/main.rs
@@ -33,6 +33,7 @@ use frost_secp256k1_tr as frost;
 use futures_util::future::FutureExt;
 use rand::thread_rng;
 use rpc::FILE_DESCRIPTOR_SET;
+use serde::{Deserialize, Serialize};
 use shutdown::{stop_signal, StopHandle};
 use thiserror::Error;
 use tokio::sync::{oneshot, Mutex};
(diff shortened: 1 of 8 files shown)


bin/btc-server/src/main.rs
```diff
@@ -303,7 +336,60 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+            .expect("bitcoind_pass is required"),
+        fee_rate_diff_percentage: cli_config
+            .fee_rate_diff_percentage
+            .or(file_config.fee_rate_diff_percentage)
+            .expect("fee_rate_diff_percentage is required"),
+        fall_back_fee_rate_sat_per_vbyte: cli_config
+            .fall_back_fee_rate_sat_per_vbyte
+            .or(file_config.fall_back_fee_rate_sat_per_vbyte)
+            .expect("fall_back_fee_rate_sat_per_vbyte is required"),
+    };
```
BatiGencho: I feel this could be put into its own separate method or even better From, TryFrom impl maybe ?

arminsabouri: Good idea. A utility method may work best since it requies both cli config and file config

arminsabouri: Moved to config.rs in [860767f](https://github.com/botanix-labs/Macbeth/pull/232/commits/860767f94cfa4f6a48396b312c41edf222a2237b)

BatiGencho (dismissed):

scottmillner (dismissed):

arminsabouri:
Waiting for CI to pass. Will merge after

BatiGencho (approved):

PR #233, opened 2024-06-10
fix: add genesis authorities to storage and validate list when import…

scottmillner:
…ing block

### DIFF
--- a/crates/consensus/authority/src/lib.rs
+++ b/crates/consensus/authority/src/lib.rs
@@ -144,6 +144,7 @@ where
     fn try_new(
         client: Client,
         headers: &mut [SealedHeader],
+        genesis_authorities: Vec<secp256k1::PublicKey>,
         authorities: Vec<secp256k1::PublicKey>,
         signer_index: usize,
         pk: secp256k1::PublicKey,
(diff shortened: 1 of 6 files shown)


BatiGencho (approved):

crates/consensus/authority/src/lib.rs
```diff
@@ -183,9 +185,11 @@ where
 /// In-memory storage for the chain the authority seal engine is building.
 pub(crate) struct StorageInner<Client> {
     client: Client,
+    /// The authority list in the genesis block
+    pub(crate) genesis_authorities: Vec<secp256k1::PublicKey>,
     /// Keep track of the  signers
     pub(crate) authorities: Vec<secp256k1::PublicKey>,
-    /// keep track of my place among the singer
+    /// keep track of my place among the signer
```
arminsabouri, line 192: Nit: during storage creation we could parse the gensis block edh and pull the authorities out there. Then we don't need two authorities fields

arminsabouri (approved):

crates/consensus/authority/src/lib.rs
scottmillner: yeah i originally was going to do it that way:

- i get that for a static federation, the authorities in the genesis block should always be the authority list moving forward
- but in the future if we go dynafed, then the two lists will diverge
- and storing the genesis authorities list then probably is unnecessary
- but thought there may be some reason we'd want to have the original genesis authority list

probably overthinking it but that's why i went with two lists

PR #242, opened 2024-06-12
fix: handle updating state for all pegin/pegout reverts

scottmillner:
Overview:

This pr updates state for failed pegins/pegouts and covers all fail paths.

Since a pegin/pegout that gets to our validation code is considered successful by the EVM, the state has been updated with the result of the pegin/pegout: balance incremented/decremented.

So we need to update the balance to what it was prior to the pegin/pegout minus the tx cost (only for pegout bc we send pegins and get refunded in the contract method itself)

### DIFF
--- a/bin/test-suite/src/suite/consensus/invalid_transactions/test_invalid_pegout.rs
+++ b/bin/test-suite/src/suite/consensus/invalid_transactions/test_invalid_pegout.rs
@@ -1,5 +1,9 @@
 use bitcoin::Amount;
 use reth_botanix_lib::utils::AmountExt;
+use reth_btc_wallet::address::EthAddress;
+use reth_cli_runner::CliRunner;
+use reth_primitives::Address;
+use std::time::Duration;
 
 use crate::{it_info_print, suite::consensus::ConsensusIntegrationTestSuite};
 
@@ -19,9 +23,22 @@ pub async fn invalid_pegout(
     let invalid_pegout_destination = ethers::core::types::Bytes::from(
         "invalid_pegout_destination".to_string().as_bytes().to_vec(),
     );
+
+    let sender_address = botanix_eth_client.get_sender_address();
+    let sender_address_string = Address::from_slice(sender_address.as_slice()).to_string();
+    // sender address balance before pegout
+    let mut sender_address_initial_balance =
+        botanix_eth_client.get_botanix_balance(sender_address_string.as_str()).await.unwrap();
+    it_info_print!("Sender address initial balance: ", sender_address_initial_balance);
+
+    // nonce before pegout
+    let nonce_before = botanix_eth_client.get_nonce(sender_address.clone()).await.unwrap();
+    it_info_print!("Nonce before pegout: ", nonce_before);
+
     // use empty pegout data
     let pegout_data = ethers::core::types::Bytes::new();
     let pegout_amount = Amount::from_btc(0.5).unwrap();
+    it_info_print!("Pegout amount: ", pegout_amount.to_wei());
     let tx_receipt = botanix_eth_client
         .burn(invalid_pegout_destination, pegout_data, pegout_amount.to_wei())
         .await
@@ -30,6 +47,24 @@ pub async fn invalid_pegout(
     it_info_print!("Pegout Tx Receipt: ", tx_receipt);
 
     assert!(tx_receipt.status.unwrap().is_zero());
+
+    // sender address balance after pegout
+    let sender_address_final_balance =
+        botanix_eth_client.get_botanix_balance(sender_address_string.as_str()).await.unwrap();
+    it_info_print!("Sender address final balance: ", sender_address_final_balance);
+
+    // subtract tx costs from initial balance
+    let tx_cost = tx_receipt.gas_used.unwrap() * tx_receipt.effective_gas_price.unwrap();
+    it_info_print!("Tx cost: ", tx_cost);
+    sender_address_initial_balance -= tx_cost;
+
+    assert_eq!(sender_address_initial_balance, sender_address_final_balance);
+
+    // nonce after pegout
+    let nonce_after = botanix_eth_client.get_nonce(sender_address.clone()).await.unwrap();
+    it_info_print!("Nonce after pegout: ", nonce_after);
+
+    assert!(nonce_after > nonce_before);
 
     Ok(())
 }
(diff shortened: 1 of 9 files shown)


crates/botanix-lib/src/mint_validation.rs
```diff
@@ -156,6 +168,9 @@ pub fn parse_pegin_topic(
             let destination = topic_to_address(log.topics()[1])?;
             let data = &log.data;
 
+            let pegin_data_error =
+                PeginDataError { pegin: (destination, amount, String::from("Pegin")) };
```
scottmillner, line 172: Not using `MintContractErrorType::Pegin.to_string()` here bc importing `MintContractErrorType` causes a circular dependency.

bin/test-suite/src/suite/consensus/invalid_transactions/test_invalid_pegout.rs
```diff
@@ -56,5 +71,23 @@ pub async fn invalid_pegout(
...
+    sender_address_initial_balance -= tx_cost;
+
+    assert_eq!(sender_address_initial_balance, sender_address_final_balance);
+
+    // nonce after pegout
+    let nonce_after = botanix_eth_client.get_nonce(sender_address.clone()).await.unwrap();
+    it_info_print!("Nonce after pegout: ", nonce_after);
+
+    assert!(nonce_after > nonce_before);
+
```
arminsabouri, line 68: Any other validations to check for? 
Maybe:
- peging out bitcoin you don't have
- Invalid pegout amount
?

crates/revm/src/processor.rs
```diff
@@ -261,14 +292,48 @@ where
...
+        account.info.balance = account
+            .info
+            .balance
+            .checked_sub(U256::from_be_bytes(amount.into()))
+            .expect("No overflow for checked_sub");
+        // update state with new balance
+        state.insert(address, account);
+    }
+
+    fn increment_balance_by_address(address: Address, amount: EthersU256, state: &mut EvmState) {
```
arminsabouri, line 327: Missing fn docs

crates/revm/src/processor.rs
```diff
@@ -261,14 +292,48 @@ where
...
+                    }
+                    .into());
                 }
             }
         }
 
         Ok(())
     }
 
+    /// Updates state for an account
```
arminsabouri: Nit: can we include a comment for this pair of functions that this is mean to only be used for revert and some description around this.

crates/revm/src/processor.rs
```diff
@@ -322,28 +395,44 @@ where
...
+                                BlockExecutionError::Validation(
+                                    BlockValidationError::FailedToParseMintTopic {
+                                        event_data: Some((address, amount, event_type)),
+                                    },
+                                ) => Some((address, amount, event_type)),
+                                _ => None,
+                            };
+
+                            if let Some(event_data) = event_data {
+                                let (address, amount, event_type) = event_data;
```
arminsabouri, line 419: How is gas taken into acount here? is amount = pegout_amount - gas ?
im assuming thats what it is

arminsabouri (commented):
Awesome refactor. Love the new granular error codes. 
I just had a few questions

crates/revm/src/processor.rs
scottmillner: at this point in the code, the gas for both pegins and pegouts has been accounted for:
- Pegin: the minting contract `mint` method subtracts the tx cost from the initial pegin amount and that difference is what the destination address receives so no additional math is needed in the above code
- Pegout: the amount is the `value` field in the tx and the tx cost (gas) has already been decremented from the sender so no additional math is needed

So `pegout amount == tx.value`
and the tx cost (gas) is decremented from the sender (`tx.to`) according to normal evm rules.  So if the sender didn't have enough to pay for the tx and send the value amount, it'd be rejected from the mempool.

So tx costs should never be added back during a revert since you still gotta pay even if the tx fails.

BatiGencho (dismissed):
Awesome job!

bin/test-suite/src/suite/consensus/invalid_transactions/test_invalid_pegout.rs
scottmillner: based on our conversation:
- `Invalid pegout amount`: can't send an invalid amount since we send the pegout amount to the mint contract using the `value` field in the tx.  The EVM will reject any tx that tries to send more than it has
- `peging out bitcoin you don't have`: im able to create a test that does this but only bc we prefund accounts on testnet.  So the test pegs out more botanix btc then btc in the multisig.  This can never happen on mainnet since we won't be prefunding any accounts in the genesis block.  So I don't think this test makes much since.

crates/revm/src/processor.rs
scottmillner: done!

crates/revm/src/processor.rs
scottmillner: done!

crates/revm/src/processor.rs
```diff
@@ -227,7 +234,23 @@ where
...
+                                Some(pegin)
+                            }
+                            MintConsensusError::FailedtoDeserialiseMetadata(pegin) => Some(pegin),
+                            _ => None,
+                        };
+                        let mint_topic_pegin_data = if pegin_data_error.is_some() {
+                            Some(pegin_data_error.expect("pegin data error to exist").pegin)
+                        } else {
+                            None
+                        };
```
arminsabouri, line 250: for the future. I think if let some sematics tend to work better here

crates/revm/src/processor.rs
arminsabouri: Awesome thank you!

crates/revm/src/processor.rs
```diff
@@ -322,28 +398,44 @@ where
...
-                                updated_state.insert(address, pegin_account);
-
-                                ResultAndState { result: new_result, state: updated_state }
+                            // Update state for reverted pegins/pegouts:
+                            // balances have been updated since tx was successful according to EVM
+                            // and we are reverting according to botanix validation
+                            let event_data = match e {
+                                BlockExecutionError::Validation(
+                                    BlockValidationError::MintContractViolation {
+                                        event_data: (address, amount, event_type),
```
arminsabouri, line 407: For the future I think we could have created a new struct for this event data. And kept event_type as an enum

arminsabouri (approved):

PR #244, opened 2024-06-13
block building or fetching should not occur if sync is in progress

BatiGencho:


scottmillner (dismissed):

crates/consensus/authority/src/block_builder.rs
```diff
@@ -39,6 +39,13 @@ where
         ConfigureEvmEnv + Clone + Unpin + Send + Sync + 'static + reth_node_api::ConfigureEvm,
 {
     pub(crate) async fn try_build_block(&mut self) {
+        // ensure the node is not syncing
+        if is_active_sync_in_progress(&self.network_handle) {
+            warn!(target: "consensus::authority", "Node is still syncing, block builder task is awaiting fully synced status ...");
+            tokio::time::sleep(Duration::from_secs(2)).await;
```
scottmillner, line 45: just curious on why 2 seconds?

BatiGencho: its just a value to introduce some relaxation within the loop not to keep the runtime busy.

crates/consensus/authority/src/pbft_task.rs
```diff
@@ -53,6 +58,10 @@ pub struct PbftTask<Client, ToFrostMan: ToFrostManager, NetworkClient> {
     secret_key: secp256k1::SecretKey,
     /// config
     config: FrostConfig,
+    /// network client
+    network_client: NetworkClient,
```
arminsabouri, line 62: Do we need network client?

arminsabouri (dismissed):
Nicely done!

crates/consensus/authority/src/pbft_task.rs
BatiGencho: yes, the pbft state machine needs it.

scottmillner (approved):

arminsabouri (approved):

arminsabouri:
./test_runner passing locally. Merging

PR #247, opened 2024-06-14
Test/arm/misc test suite fixes

arminsabouri:
I would like to do a bit more clean up but open to review right now

### DIFF
--- a/bin/test-suite/src/suite/consensus/common/poa_node.rs
+++ b/bin/test-suite/src/suite/consensus/common/poa_node.rs
@@ -87,7 +87,7 @@ pub struct FederationMemberTestConfig {
     pub bitcoind_password: String,
     pub bitcoin_server_url: String,
     pub peers_list: Vec<FederationMemberTestConfig>,
-    pub sender: tokio::sync::mpsc::Sender<Notifications>,
+    pub sender: tokio::sync::broadcast::Sender<Notifications>,
     pub jwt_secret_path: PathBuf,
     pub frost_min_signers: u16,
     pub frost_max_signers: u16,
@@ -102,7 +102,7 @@ impl FederationMemberTestConfig {
         index: u16,
         secret_key: SecretKey,
         authorities: Vec<PublicKey>,
-        sender: tokio::sync::mpsc::Sender<Notifications>,
+        sender: tokio::sync::broadcast::Sender<Notifications>,
         bitcoind_url: Url,
         bitcoind_username: String,
         bitcoind_password: String,
@@ -222,6 +222,7 @@ impl FederationMemberTestConfig {
             "--chain",
             "botanix_testnet",
             "--federation-mode",
+            "--ipcdisable",
             "--datadir",
             datadir,
             "--debug.terminate",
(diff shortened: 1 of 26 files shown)


bin/test-suite/src/suite/consensus/common/poa_node.rs
```diff
@@ -87,7 +87,7 @@ pub struct FederationMemberTestConfig {
     pub bitcoind_password: String,
     pub bitcoin_server_url: String,
     pub peers_list: Vec<FederationMemberTestConfig>,
-    pub sender: tokio::sync::mpsc::Sender<Notifications>,
+    pub sender: tokio::sync::broadcast::Sender<Notifications>,
```
scottmillner, line 90: why do we need multiple consumers instead of single which `broadcast` allows?
or maybe this change was made for another reason?

crates/consensus/common/src/utils.rs
```diff
@@ -157,9 +157,14 @@ pub fn validate_poa_extra_data_header(
...
-            edh.authority_signers.as_ref().expect("authority signers to exist")
-        {
+        let gensis_hashset = genesis_authorities.iter().collect::<std::collections::HashSet<_>>();
+        let edh_hashset = edh
+            .authority_signers
+            .as_ref()
+            .expect("authority signers to exist")
+            .iter()
+            .collect::<std::collections::HashSet<_>>();
+        if gensis_hashset != edh_hashset {
```
scottmillner, line 167: these changes ensure a unique set of authorities but order isn't gauranteed
didn't we want to ensure order as well?
that wasn't causing this check to sometimes fail, right?

i realize this was my oversight so thanks for fixing it!

arminsabouri: This is a temporary fix and I created a ticket here: https://github.com/botanix-labs/botanix/issues/491
So the order does from the chain.toml does indeed matter. The reason is our frost id is based on your index of your key in chain.toml. So I originally sorted the keys in EDH but this broke the frost id assignment.

arminsabouri: Now that i am thinking about this. If genesis authorities is in the same order as chain.toml then I should be able to revert this change.

arminsabouri: I'll come back to this

bin/test-suite/src/suite/consensus/common/poa_node.rs
arminsabouri: Ah yes good point! I needed to change this b/c I needed to clone a `tokio::sync::mpsc::Receiver` so i can share it with the individual tests that need to consume it. Alternatively we could have done `Arc<Mutec<...>>` But under @BatiGencho recommendation we decided to go with a broadcast channel

arminsabouri: Although I wonder if this could be the cause of https://github.com/botanix-labs/botanix/issues/495

BatiGencho (dismissed):

scottmillner (dismissed):

scottmillner (approved):

scottmillner:
Screenshot 2024-06-17 at 3 42 28 PM

PR #252, opened 2024-06-18
Makefile: Support running integration tests directly + run all

stevenroose:


BatiGencho (dismissed):

Makefile
```diff
@@ -435,6 +435,37 @@ start-test-suite:
...
+run-test-frost-e2e-failed-signing-disconnect:
+	TEST_TO_RUN=frost_e2e_failed_signing_disconnect make start-test-suite
+run-test-invalid-pegin:
+	TEST_TO_RUN=invalid_pegin make start-test-suite
+run-test-invalid-pegout:
+	TEST_TO_RUN=invalid_pegout make start-test-suite
+run-test-mempool-gossip:
+	TEST_TO_RUN=test_mempool_gossip make start-test-suite
+
+run-test-all:
```
arminsabouri, line 457: Can you re-purpose this to use test_runner.sh

stevenroose: oooh didn't see that one.. I thought everything was using makefile.. Let me check then maybe my code isn't needed anymore. Lol at Liquid I had a file with exactly the same name, test_runner.sh.

stevenroose:
I will close this in favor of using test_runner.sh. @BatiGencho I don't really understand the bash you wrote there, could you make it so that passing an argument to test_runner.sh only runs the test with that name? So I can run 
```
$ test_runner.sh frost_e2e_stable
```

PR #254, opened 2024-06-18
Use /usr/bin/env bash instead of /bin/bash

stevenroose:
This should work for everyone, it's the POSIX standard :)


arminsabouri (approved):

arminsabouri:
Waiting for CI to pass before merging

PR #257, opened 2024-06-18
Scott/491 fix epoch block should check authorities are in the correct order

scottmillner:


crates/consensus/common/src/utils.rs
```diff
@@ -131,6 +131,32 @@ pub fn validate_poa_block_beneficiary(header: &Header) -> Result<(), ConsensusEr
     Ok(())
 }
 
+/// Check authorities in EDH match and are in the same order and the genesis authorities
+pub fn validate_extra_data_header_authorities(
```
BatiGencho, line 135: Why do we want to do this - wont it just suffice to check they are the same ?

crates/consensus/common/src/utils.rs
```diff
@@ -131,6 +131,32 @@ pub fn validate_poa_block_beneficiary(header: &Header) -> Result<(), ConsensusEr
...
+        // Attempt to deserialize the extra data header
+        let edh = header.deserialize_extra_data_header().map_err(|e| {
+            error!("Failed to deserialize extra data header: {:?}", e);
+            ConsensusError::ExtraDataInvalid
+        })?;
+
+        // Validate the list of authorities matches the authorities in the genesis block
+        // This check is only for a static federation
+        if genesis_authorities !=
+            edh.authority_signers.as_ref().expect("authority signers to exist")
```
BatiGencho: we can still compare things without the expect(...) :

`if let Some(authority_signers) = edh.authority_signers { do the actual comparison here... } `

crates/consensus/common/src/utils.rs
scottmillner: from the comments in the [GH issue](https://github.com/botanix-labs/botanix/issues/491)
`At first we attempted to fix this by sorting the keys in the edh. However order that is created in the chain.toml matters. As the frost identifier is a hash(index).`

crates/consensus/common/src/utils.rs
scottmillner: thanks! i refactored it as above and added else clause to return error if no authorities in the edh and added an additional test case for that
done!

crates/consensus/common/src/utils.rs
BatiGencho: got it , all clear.

crates/consensus/common/src/utils.rs
BatiGencho: perfect, thx!

BatiGencho (dismissed):

bin/test-suite/src/suite/consensus/common/poa_node.rs
```diff
@@ -214,8 +217,21 @@ impl FederationMemberTestConfig {
         };
         fed_member_pks.push(my_pk);
 
+        // NOTE: fed members have already created their EDH with the correct authorities
+        // but the order may not be the same as fed_member_pks since we added ourselves last
+        // so compare the EDH authorities list and build a new list in the correct order
+        let mut edh_authorities = vec![];
+        for authority in edh_authorities_list.to_vec().iter() {
```
arminsabouri, line 224: Since we have a `edh_authorities_list` can we just use that instead of creating the `fed_member_pks` vec?

arminsabouri (commented):
Had one small Nit

bin/test-suite/src/suite/consensus/common/poa_node.rs
scottmillner: the Vec elements are different types:
- `GenesisTomlConfig::new()` expects `Vec<FedMemberPubKey>` which is the pub key and the socket addr
- that's why i need to use `fed_member_pks` bc it's `Vec<FedMemberPubKey>`
- `edh_authorities_list` is `Vec<PublicKey>`

BatiGencho (approved):

arminsabouri (approved):

PR #258, opened 2024-06-20
fixes for cleanup and weird behaviour

BatiGencho:
This MR provides 3 fixes:
1. when the test suite is run locally, the very first test simply creates context, but there might already be running btc servers so often we fail unless we cleanup the instances. For this reason I have added a cleanup part before actually constructing new context. 
2. At the end of each destructor we need a relaxation of a few secs. as processes take a while to go away completely.
3. Receiver channels are often being dropped during context destruction and channels tend to loose receivers leading to not being able to send messages. In those cases we simply retry until a new receiver comes back up.

### DIFF
--- a/test_runner.sh
+++ b/test_runner.sh
@@ -4,7 +4,7 @@ set -e
 set -o pipefail
 
 # Define the list of strings
-tests_to_run=("dkg_flow" "many_inputs_signing" "utxo_commitment" "block_builder" "frost_e2e_stable" "frost_e2e_failed_signing_disconnect" "invalid_pegin" "invalid_pegout" "test_mempool_gossip")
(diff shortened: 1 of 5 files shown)


test_runner.sh
```diff
@@ -4,7 +4,7 @@ set -e
 set -o pipefail
 
 # Define the list of strings
-tests_to_run=("dkg_flow" "many_inputs_signing" "utxo_commitment" "block_builder" "frost_e2e_stable" "frost_e2e_failed_signing_disconnect" "invalid_pegin" "invalid_pegout" "test_mempool_gossip")
```
arminsabouri, line 7: What was the problem with the pegout test?

arminsabouri (commented):
approachACK. My only question pertains to the disabled test

PR #259, opened 2024-06-20
test: delay so btc server has time to spin up

scottmillner:


bin/test-suite/src/suite/consensus/invalid_transactions/test_invalid_pegout.rs
```diff
@@ -11,6 +10,9 @@ use crate::{it_info_print, suite::consensus::ConsensusIntegrationTestSuite};
 pub async fn invalid_pegout(
     suite: &ConsensusIntegrationTestSuite,
 ) -> Result<(), super::error::InvalidTransactionError> {
+    // wait for everything to spin up otherwise btc_server will hit `Unable to get public key`
+    tokio::time::sleep(Duration::from_secs(5)).await;
```
arminsabouri: If you take a look at create context we repeately poll the btc servers till they are all up. Would that suffice here?

scottmillner: not quite following?
`create_new_context` has already run successfully by the time the test run so all the btc servers are up and running

but when one of poa nodes tries to authenticate with one of the btc_servers, it can't retreive it's pk from it's db yet

the delay resolves this...not totally sure why

scottmillner: @0xBEEFCAF3 and I talked about this

will move that delay to `create_new_context` method
done!

arminsabouri (approved):

arminsabouri (approved):

arminsabouri:
e2e passed with last test run so merging

PR #261, opened 2024-06-20
ci: clean up should be a step of `test / e2e` workflow

arminsabouri:
I noticed that some e2e tests simply stopped execuing wihtout a warning or error. I believe the clean up step should be part of the main e2e step. Other wise the clean up task could run on a different runner than the e2e test.


birozuru (approved):
lgtm

PR #265, opened 2024-06-25
Test/arm/increase timeout

arminsabouri:


.github/workflows/integration.yml
```diff
@@ -65,10 +65,8 @@ jobs:
           set -o pipefail
           chmod +x test_runner.sh
           bash test_runner.sh
-          sleep 3000
```
arminsabouri, line 68: im not sure why this sleep was here? Maybe in the past the previous command ran in the bkg...

scottmillner (approved):

.github/workflows/integration.yml
BatiGencho: i dont think we need it

BatiGencho (approved):

PR #268, opened 2024-06-25
Feature/chore/remove client from storage

arminsabouri:
Big refactor to remove client from poa storage. 
Part of this refactor also uses more concise trait contraints where needed. Where as before we we're using this everwhere
```
Client: BlockReaderIdExt
        + StateProviderFactory
        + CanonChainTracker
        + BlockchainTreeEngine
        + Clone
        + 'static,

```

This ended up being an issue when we needed to mock storage to other crate structs unit tests. Now storage can be mocked by using the trait `AuthorityStorage`. 

As a part of https://github.com/botanix-labs/botanix/issues/494 

In the future we could also remove all the methods that create the EVM processor and execute things and have them as standalone utilities. No need to be a part of authority storage

### DIFF
--- a/crates/consensus/authority/src/block_builder.rs
+++ b/crates/consensus/authority/src/block_builder.rs
@@ -58,9 +58,13 @@ where
             return;
         }
 
-        let storage = self.storage.write().await;
-        let (best_block, best_hash) = storage.get_best_block_and_hash().expect("best block exists");
-        drop(storage);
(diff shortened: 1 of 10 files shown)


scottmillner:
looks good, will wait for tests to pass then ✅ 
🔥 💪 🎸

arminsabouri:
@scottmillner Looks like e2e. Do you mind doing another pass when you get a chance

scottmillner (approved):

PR #271, opened 2024-06-26
Arm/refactor/executor factory

arminsabouri:
4 @scottmillner but this time it compiles


arminsabouri:
After doing this refactor I am realizing that we can also use the EF in block builder / fetcher as well and ommit engine types and evm config

scottmillner (approved):
🔥

PR #273, opened 2024-06-27
authority: Fix one loop and optimize another

stevenroose:

PR #274, opened 2024-06-27
Multiplex some channels

stevenroose:
Evgeni told me not to do this MR because he has other MRs that are conflicting with this. https://github.com/botanix-labs/Macbeth/pull/273/files is probably sufficient for now, but it could be made nicer using the multiplexing. The one in frost_task could be done equivalently, but I guess we'll have to get back at this when these files are stabilized and there's no other pending work on them.


stevenroose:
I'll just close it, just keep in mind we can do this if we want to :)

PR #279, opened 2024-06-28
Scott/494 fix should run full conseusus validate before giving pre commitment

scottmillner:


crates/consensus/authority/src/lib.rs
```diff
@@ -140,35 +141,26 @@ impl Consensus for AuthorityConsensus {
...
-        }
+        validate_extra_data_header_authorities(header, genesis_authorities)?;
 
-        // Validate a quorum of authority signatures
-        let valid_sigs = header.check_authority_sig_add(authority_signers).map_err(|e| {
-            error!("Failed to validate authority signature: {:?}", e);
-            ConsensusError::InvalidAuthoritySignature
-        })?;
+        // Validate a quorum of authority signatures except during pbft
+        if !is_pbft {
```
scottmillner: went with a boolean instead of `validate_extra_data_header_single_signer` bc:
- using `validate_extra_data_header_single_signer` has no way of knowing if this is pbft or not so this would potentially skip validation when it should happen

crates/consensus/authority/src/lib.rs
```diff
@@ -211,9 +204,16 @@ impl Consensus for AuthorityConsensus {
...
+            if header.timestamp > current_timestamp {
+                ConsensusError::TimestampIsInFuture {
+                    timestamp: header.timestamp,
+                    present_timestamp: current_timestamp,
+                }
+            } else {
+                ConsensusError::TimestampIsInPast {
+                    timestamp: header.timestamp,
+                    present_timestamp: current_timestamp,
+                }
```
BatiGencho, line 211: wasnt this supposed to be a trait impl on the header ?

BatiGencho (dismissed):
Overall it looks fine to me, need to go through it once more, but so far great!

crates/consensus/authority/src/block_builder.rs
```diff
@@ -305,6 +305,7 @@ where
                 return;
             }
         };
+        drop(storage);
```
arminsabouri, line 308: Nice find. yet another reason why we should remove these exec utils from storage

crates/consensus/authority/src/block_fetcher.rs
```diff
@@ -189,43 +190,42 @@ where
...
-                                    error!(target: "consensus::authority", ?e, "Failed to process botanix log");
-                                    continue;
-                                }
-                            };
+                    let header = sealed_block.header.clone();
 
-                            pegouts
+                    // process pegins
+                    // must be done before getting utxo commitment
+                    let rpc = self.btc_server.as_mut().expect("btc_server exists");
```
arminsabouri:
```suggestion
                    let btc_signing_server_rpc = self.btc_server.as_mut().expect("btc_server exists");
```

More verbose is fine here

crates/consensus/authority/src/lib.rs
arminsabouri: in what condition would it skip validation?

crates/consensus/authority/src/pbft.rs
```diff
@@ -617,6 +708,9 @@ where
             return Ok(());
         }
 
+        // execute block and run poa consensus
+        self.execute_and_validate_poa_consensus(block.clone()).await?;
```
arminsabouri, line 697: Do we need to validate on both commits and pre commits?
Im thinking we fully validate on pre commits and only check if we have precommited before commiting

arminsabouri (commented):
Overall looking good just had a few things

crates/consensus/authority/src/pbft.rs
scottmillner: ah yep, removed in `process_commitment`

scottmillner:
int test passed locally

Screenshot 2024-07-14 at 4 42 49 PM

crates/consensus/authority/src/lib.rs
scottmillner: addressed: changed to use `validate_extra_data_header_single_signer`

crates/consensus/authority/src/lib.rs
scottmillner: can do this in separate ticket if we want

crates/consensus/authority/src/block_fetcher.rs
scottmillner: done!

BatiGencho (approved):

PR #280, opened 2024-07-01
Arm/feature/pk header

arminsabouri:


bin/test-suite/src/suite/consensus/frost/test_block_builder.rs
```diff
@@ -106,6 +107,17 @@ pub async fn block_builder(
     it_info_print!("Eoa tx: {:?}", last_tx_hash);
     tx_hashes_set.insert(last_tx_hash.transaction_hash);
 
+    // retrieve the current aggreate public key
```
scottmillner: aggreate > aggregate

crates/consensus/authority/src/lib.rs
```diff
@@ -144,6 +154,27 @@ impl Consensus for AuthorityConsensus {
             ConsensusError::ExtraDataInvalid
         })?;
 
+        // Past genesis NUMS point should never be used as the aggregated public key
+        if edh.aggregated_public_key ==
```
scottmillner: could pull into util method since used twice but also only used twice so just a thought

crates/consensus/authority/src/lib.rs
```diff
@@ -236,10 +279,24 @@ impl Consensus for AuthorityConsensus {
         validation::validate_header_extradata(header)?;
 
         // Attempt to deserialize the extra data header
-        let _edh = header.deserialize_extra_data_header().map_err(|e| {
+        let edh = header.deserialize_extra_data_header().map_err(|e| {
             error!("Failed to deserialize extra data header: {:?}", e);
             ConsensusError::ExtraDataInvalid
         })?;
+
+        if edh.aggregated_public_key ==
```
scottmillner: see common above about util method

crates/consensus/authority/src/lib.rs
arminsabouri: yeah good point. Also considering moving the NUMS pk creation step into its own util method as well

arminsabouri: Note to self: or a macro!

crates/consensus/authority/src/lib.rs
```diff
@@ -144,6 +153,20 @@ impl Consensus for AuthorityConsensus {
...
+                InvalidAggregatedPublicKeyError::NumsAggregatePublicKeyPastGenesis,
+            ));
+        }
+
+        // Validate public key placed in the extra data header
+        if edh.aggregated_public_key != *aggregate_public_key.expect("aggregated public key") {
+            return Err(ConsensusError::InvalidAggregatedPublicKey(
+                InvalidAggregatedPublicKeyError::InvalidAggregatedPublicKey,
+            ));
+        }
```
BatiGencho, line 175: why dont we do it a bit cleaner here and not expect. 

```
aggregate_public_key.and_then(|key| { edh.aggregated_public_key == key }).unwrap_or_else(|| Err(ConsensusError::InvalidAggregatedPublicKey(
                InvalidAggregatedPublicKeyError::InvalidAggregatedPublicKey,
            )) )?
```

You might want to check the proposed above a bit, but something within those lines

crates/consensus/authority/src/lib.rs
```diff
@@ -619,6 +663,7 @@ impl StorageInner {
     ) -> Result<SealedHeader, BlockExecutionError> {
         let Block { header, body, .. } = block;
         let body = BlockBody { transactions: body, ommers: vec![], withdrawals: None };
+        let consensus_pkg = botanix_consensus_pkg.as_ref().expect("consensus pkg");
```
BatiGencho: Whats the point of taking a ref, then unwrapping it ? Cannot we do this down below where it is being passed as a fn arg:

`botanix_consensus_pkg.as_ref().map(|pkg| pkg.bitcoint_checkpoint.0.block_hash()).unwrap_or_else(|| ... )?`

crates/consensus/authority/src/lib.rs
```diff
@@ -696,6 +742,7 @@ impl StorageInner {
                 &sealed_block.header.clone(),
                 &authority_signers,
                 &genesis_authorities,
+                Some(&botanix_consensus_pkg.as_ref().expect("consensus pkg").aggregate_public_key),
```
BatiGencho, line 742: see a bit above

BatiGencho (dismissed):

crates/consensus/authority/src/lib.rs
arminsabouri: turns out this didnt need to be a option at all

PR #281, opened 2024-07-02
Arm/feature/dynamic commitment pbft

arminsabouri:


crates/consensus/authority/src/pbft.rs
```diff
@@ -30,6 +30,22 @@ use tokio::sync::{
 };
 use tracing::{debug, error, info, warn};
 
+/// Returns the minimum number of pre-commitments required for a block
+/// This is not consenusus critical but will prevent the number of forked blocks
+/// n being the number of authorities
+#[inline]
+pub fn min_pre_commitments(n: u16) -> u16 {
```
BatiGencho: Just a nit: maybe we could not have these loose func but rather bundle them under some struct with static methods like:

```
pub struct PbftCommitmentCriteria {
#[inline]
 pub fn mint_pre_commitments(n: u16) ...
 #[inline]
 pub fn min_commitments(n: u16_ ...
}
```

just so we have the whole lot bundled in one place. You can ignore this comment too

BatiGencho (approved):

crates/consensus/authority/src/pbft.rs
scottmillner: i like that idea as well

scottmillner (approved):

crates/consensus/authority/src/pbft.rs
arminsabouri: I like this as well. implemented

PR #282, opened 2024-07-04
Add script to generate Rust docs

stevenroose:
Thought this might be useful. I do this myself but rely on shell history to find the correct command, and like this it includes everything :)

For those that don't understand, this generates a directory with the rust docs for all our crates and all dependencies that you can browse and search in a browser without needing internet connection. It's great. I use this one and the stdlib one from rustup all the time.


BatiGencho (approved):

stevenroose:
Can't figure it out, is something actually failing here?

arminsabouri (approved):

PR #285, opened 2024-07-09
extended chain.toml with is-testnet boolean

BatiGencho:


arminsabouri (approved):

scottmillner (dismissed):

crates/node-core/src/args/genesis_args.rs
```diff
@@ -56,13 +56,22 @@ pub struct FedMemberPubKey {
...
+/// Federation member public key and socket address
 #[derive(Clone, Debug, Deserialize, Serialize)]
 #[serde(deny_unknown_fields, rename_all = "kebab-case")]
-pub struct GenesisTomlConfig {
+pub struct GenesisNetworkConfig {
     /// Network name
-    #[allow(dead_code)]
     pub name: String,
+    /// Indicates whether it is a test network
+    pub is_testnet: bool,
```
arminsabouri: My idea was the use this bool to switch between 3636 and the mainnet chainspec (which we dont have defined in code yet). And remove the idea of pasing chain as a param in cli flags. Can we do that with this boolean?

arminsabouri (changes requested):
Just noticed that we we're not using this is_testnet for its intended usage.

arminsabouri (dismissed):

.github/workflows/lint.yml
```diff
@@ -1,145 +1,146 @@
...
-      - uses: dtolnay/rust-toolchain@stable
-      - uses: taiki-e/install-action@cargo-hack
-      - uses: Swatinem/rust-cache@v2
-        with:
-          cache-on-failure: true
-          cache-all-crates: true
-      - name: Install Protobuf Compiler
-        run: sudo apt-get update && sudo apt-get install -y protobuf-compiler
-      - run: cargo hack check
+    # Commenting out for now, will re-enable after final upstream merge
```
scottmillner, line 50: this is the only actual change in this file (the others are just formatting)

i don't think we're gaining much from this test but ill add back once we finish the last upstream merge

PR #286, opened 2024-07-10
cleanup, clippy, warnings removed, slack removed

BatiGencho:


arminsabouri (dismissed):
Nice work! Should we setup a clippy + warning job now that we have them all resolved?

PR #293, opened 2024-07-16
 Refactor most of the Minting contract log parsing and handling code 

stevenroose:


stevenroose:
If you want to read some code that is actually readable, you can always review the second commit, which contains the bulk of this change, and then verify that the last commit only does `make fmt` ;)

contracts/src/Minting.sol
```diff
@@ -6,18 +6,34 @@ contract Minting {
     uint constant GAS_SIMPLE_TRANSFER = 21000;
     uint constant GAS_AMOUNT_UPDATE = 2003;
     uint constant GAS_REVERT_TRUE = 3;
-    uint constant GAS_MINT_AMOUNT_EVENT = 1133;
+
```
arminsabouri, line 9: Im probably not the best person to review this part
tagging @scottmillner to take a look

contracts/src/Minting.sol
```diff
@@ -6,18 +6,34 @@ contract Minting {
...
+    // > indexed parameter, so that cost is included most of the time. Then we
+    // > pay 8 times the number of 32 byte words written to the chain. Because
+    // > this region is stored in memory before being emitted, the memory
+    // > expansion cost must be accounted for also.
+    uint constant BASE_GAS_MINT_EVENT = 1157;
 
     /// Some new coins have been minted.
     event Mint(
         address indexed account,
         uint256 amount,
```
arminsabouri: Do we use amount anywhere? Seems like mint amount should suffice

crates/consensus/authority/src/block_builder.rs
```diff
@@ -181,22 +185,41 @@ where
...
+            for receipt in receipts {
+                if idx == 0 && receipt.is_none() {
+                    break; // Prunning block, skip
+                }
+                if let Some(receipt) = receipt {
+                    if !receipt.success {
+                        continue;
+                    }
+                    for log in &receipt.logs {
+                        let pegin_match = try_parse_mint_event(log).expect("passed EVM check");
```
arminsabouri, line 194: Comment justifiying the expect
```suggestion
                        // Mint event should have already been validated in during evm execution (in processor.rs) 
                        let pegin_match = try_parse_mint_event(log).expect("passed EVM check");
```

crates/consensus/authority/src/block_builder.rs
```diff
@@ -181,22 +185,41 @@ where
...
+                            info!(target: "consensus::authority", "Parsing and sending minting event to btc_server");
+                            for pegin in &pegin_data.meta {
+                                //TODO(stevenroose) should this happen here?
+                                if let Err(e) = call_notify_pegin(&mut self.btc_server, pegin).await
+                                {
+                                    error!(target: "consensus::authority", ?e, "failed to notify btc_server of pegin");
+                                    return;
+                                }
+                                info!(target: "consensus::authority", "notifying btc server about pegin utxo");
+                            }
```
arminsabouri, line 205: Should we continue here?
There is no reason to try to parse a burn event if already have a pegin.
Or maybe a if elseif

crates/consensus/authority/src/block_fetcher.rs
```diff
@@ -206,26 +208,47 @@ where
...
+                                                //TODO(stevenroose) should this happen here?
+                                                if let Err(e) =
+                                                    call_notify_pegin(btc_server, &pegin).await
+                                                {
+                                                    error!(target: "consensus::authority", ?e, "failed to notify btc_server of pegin");
+                                                    return;
+                                                }
+                                                info!(target: "consensus::authority", "notifying btc server about pegin utxo");
+                                            }
+                                        }
```
arminsabouri, line 237: Same comment here?

crates/consensus/authority/src/block_fetcher.rs
```diff
@@ -206,26 +208,47 @@ where
...
+                                    if !receipt.success {
+                                        continue;
+                                    }
+                                    for log in &receipt.logs {
+                                        let pegin_match =
+                                            try_parse_mint_event(log).expect("passed EVM check");
+                                        if let Some(pegin_data) = pegin_match {
+                                            info!(target: "consensus::authority", "Parsing and sending minting event to btc_server");
+                                            for pegin in &pegin_data.meta {
+                                                //TODO(stevenroose) should this happen here?
```
arminsabouri, line 228: Not a req for change.
I agree with this. Duplicate code. Ideally we have some Pegin/Pegout Processor util or struct

crates/revm/src/processor.rs
```diff
@@ -221,86 +220,60 @@ where
...
         result: &ExecutionResult,
         botanix_consensus_pkg: Option<BotanixConsensusPackage>,
-        sender_address: Address, // used for reverted pegouts
-        sender_value: U256,      // used for reverted pegouts
-    ) -> Result<(), BlockExecutionError> {
+    ) -> Result<(), MintContractError> {
         let consensus_pkg = botanix_consensus_pkg.as_ref();
+        let btc_network = consensus_pkg
+            .map(|package| package.btc_network)
+            .unwrap_or_else(|| bitcoin::Network::Regtest);
```
arminsabouri: Maybe we should just panic here if no network is defined. If you are processing the botanix logs then this should be defined

crates/revm/src/processor.rs
```diff
@@ -221,86 +220,60 @@ where
...
-                                pegin_data.account,
-                                pegin_data.amount,
-                                MintContractErrorType::Pegin.to_string(),
+            let pegin_data = match try_parse_mint_event(log)? {
+                None => continue,
+                Some(p) => p,
+            };
+
+            let bitcoin_checkpoint = consensus_pkg.expect("is some").bitcoin_checkpoint;
+            if pegin_data.bitcoin_block_height >= bitcoin_checkpoint.1 {
```
arminsabouri: Is this correct? The error returned in the height is below the checkpoint. But here we are checking if the pegin block height is great than or eq to the checkpoint.

crates/revm/src/processor.rs
```diff
@@ -221,86 +220,60 @@ where
...
+                        bitcoin_checkpoint.1, pegin_data.bitcoin_block_height,
+                    ),
+                    revert_address: pegin_data.account,
+                    revert_amount: pegin_data.amount,
+                });
+            }
+            let aggregate_public_key = consensus_pkg.expect("is some").aggregate_public_key;
+            match pegin_data.validate(&bitcoin_checkpoint, &aggregate_public_key) {
+                Ok(aggregate_value) => {
+                    if aggregate_value == pegin_data.amount {
```
arminsabouri: Should this be mint_amount? Or are those the same thing now?

crates/consensus/authority/src/block_builder.rs
scottmillner: i guess you could have a contract that batches both pegins and pegouts in a single transaction so there could be pegin and pegout receipts in one tx...that is technically possible

crates/revm/src/processor.rs
scottmillner: it's the same thing now

`Minting.sol` has been refactored to only emit a single event that contains both the original amount and the final mint amount

`try_parse_mint()` will set `PeginData.amount` to the final mint amount

and amount will be less than aggregate value bc tx cost is deducted from original amount

contracts/src/Minting.sol
scottmillner: we don't but it's cheap to include in the event and my preference is to keep it
makes it very transparent if anyone really wanted to dig through logs and make sure everything is above board

scottmillner:
acking it all looks good
just needs an error message to be updated...see armin's comment

crates/revm/src/processor.rs
scottmillner: the check is correct but the error message isn't

- `bitcoin_checkpoint` is the header at the required depth (`let height = height.saturating_sub(pegin_conf_depth as usize - 1);`)
- so the pegin `bitcoin_block_height` should be equal or less than the checkpoint

ive updated the error message to reflect that and added a comment to explain

crates/revm/src/processor.rs
scottmillner: done!

contracts/src/Minting.sol
scottmillner: actually forgot you can only have 4 topics so will just use mint
that's why the int tests were failing: the topic hash was incorrect since the last topic wasn't being included

arminsabouri (changes requested):
tNack
getting this on a pegin
```
Botanix Minting contract event validation failed: InvalidPeginData { error: "pegin height 350 greater than checkpoint of 353", revert_address: 0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266, revert_amount: 9999932007753125000 }
```

Not sure why the int. test is passing

arminsabouri:
> tNack getting this on a pegin
> 
> ```
> Botanix Minting contract event validation failed: InvalidPeginData { error: "pegin height 350 greater than checkpoint of 353", revert_address: 0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266, revert_amount: 9999932007753125000 }
> ```
> 
> Not sure why the int. test is passing

I believe the fix needed for this is here: https://github.com/botanix-labs/Macbeth/pull/314

When I manually add that check my pegin passes

rwlockbg (dismissed):

crates/revm/src/processor.rs
```diff
@@ -221,93 +220,69 @@ where
...
+                        pegin_data.bitcoin_block_height, bitcoin_checkpoint.1,
+                    ),
+                    revert_address: pegin_data.account,
+                    revert_amount: pegin_data.amount,
+                });
+            }
+            let aggregate_public_key = consensus_pkg.expect("is some").aggregate_public_key;
+            match pegin_data.validate(&bitcoin_checkpoint, &aggregate_public_key) {
+                Ok(aggregate_value) => {
+                    if aggregate_value == pegin_data.amount {
```
arminsabouri: `if aggregate_value == pegin_data.amount`
Cab you explain the logic here? Is it due to the refund? 
Is it possible that a refund reciever isnt defined and the refund never occurs?

arminsabouri: Additionally what is its greater? Should this be >= ?

crates/revm/src/processor.rs
```diff
@@ -369,77 +344,56 @@ where
...
-                            };
-                            let new_result = ExecutionResult::Revert {
-                                gas_used: result.gas_used(),
-                                output: output_match,
-                            };
-
+                    match Self::botanix_mint_contract_checks(&result, botanix_consensus_pkg) {
+                        Ok(()) => Ok(ResultAndState { result, state }),
+                        Err(e) => {
+                            error!("Botanix Minting contract event validation failed: {:?}", e);
```
arminsabouri, line 354:
```suggestion
                            error!("Botanix Minting contract burn event validation failed: {:?}", e);
```

crates/revm/src/processor.rs
```diff
@@ -369,77 +344,56 @@ where
...
-                                } else {
-                                    let mut updated_state = state.clone();
+                                }
+                                MintContractError::InvalidPegoutData(_) => {
                                     Self::increment_balance_by_address(
-                                        address,
-                                        amount,
-                                        &mut updated_state,
+                                        sender,
+                                        EthersU256::from_little_endian(pegout_amount.as_le_slice()),
```
arminsabouri, line 373: Should gas be account for here?

arminsabouri: Otherwise if someone can keep calling burn without having some account balance decremented won't that create supply that doesnt exist?

crates/revm/src/processor.rs
scottmillner: yes, it is do to the refund so the aggregate amount should always be lower than the `amount` passed to the contract call

the `refundAddress` has to be passed to the contract call since it's part of the method signature, if it's not passed the call will fail

yes, it should be `>=`...nice catch!
done!

crates/revm/src/processor.rs
scottmillner: gas is paid by the sender of the call to `mint()` which is sidecar at the moment but could be anyone

the EVM enforces that you have enough funds to cover gas and that has already been decremented from the sender account since this is a valid tx it terms of the EVM. It's just our additional logic that reverts it.  So gas has been handled correctly by the EVM before it gets to our custom validation logic.  All our logic needs to do is decrement the amount that was transferred to the destination address which we parse from `emit Mint(destination, amount, bitcoinBlockHeight, metadata);` in the call method.

TL;DR
it's impossible to call `burn()` without having enough ether to cover gas due to normal EVM rules

crates/revm/src/processor.rs
arminsabouri: Nice!

arminsabouri:
@scottmillner Do you mind running thru some manual pegin/pegouts on this PR? As well as the functional test suite?

PR #295, opened 2024-07-17
Feature: utxo syncing

rwlockbg:
https://github.com/botanix-labs/botanix/issues/492


bin/btc-server/proto/btc_server.proto
```diff
@@ -75,21 +77,27 @@ message TxOut {
...
-    OutPoint outpoint = 1; 
-    uint32 output = 2;
+    // The txid of the utxo in hex.
+    bytes utxo_txid = 1;
+    // The output index of the utxo.
+    uint32 utxo_vout = 2;
+    // The user's ethereum address.
     string eth_address = 3;
+    // The txout pubkey of the utxo.
+    bytes output_pubkey = 4;
```
arminsabouri: what is pubkey here? An aggregate pubkey?
Or do you scriptpubkey?

bin/btc-server/src/main.rs
```diff
@@ -314,11 +314,10 @@ impl App {
...
     fn from(item: database::Utxo) -> Self {
         rpc::Utxo {
-            outpoint: Some(rpc::OutPoint {
-                txid: AsRef::<[u8]>::as_ref(&item.outpoint.txid).to_vec(),
-                vout: item.outpoint.vout,
-            }),
-            output: item.output.value.to_sat() as u32,
+            utxo_txid: AsRef::<[u8]>::as_ref(&item.outpoint.txid).to_vec(),
+            utxo_vout: item.outpoint.vout,
+            output_pubkey: item.output.script_pubkey.to_bytes(),
```
arminsabouri: I would rename `output_pubkey` to `output_scriptpubkey`

bin/btc-server/src/main.rs
```diff
@@ -314,11 +314,10 @@ impl App {
 impl From<database::Utxo> for rpc::Utxo {
     fn from(item: database::Utxo) -> Self {
         rpc::Utxo {
-            outpoint: Some(rpc::OutPoint {
-                txid: AsRef::<[u8]>::as_ref(&item.outpoint.txid).to_vec(),
-                vout: item.outpoint.vout,
-            }),
-            output: item.output.value.to_sat() as u32,
+            utxo_txid: AsRef::<[u8]>::as_ref(&item.outpoint.txid).to_vec(),
+            utxo_vout: item.outpoint.vout,
```
arminsabouri, line 318: Any particular reason why we want to break down outpoint to its respective parts?
output generally has a well defined serialization format

crates/consensus/authority/src/block_fetcher.rs
```diff
@@ -98,6 +134,244 @@ where
...
+                    // excluse our own peer_id
+                    Some(authority_peer_id)
+                } else {
+                    None
+                }
+            })
+            .collect::<Vec<PeerId>>()
+    }
+
+    async fn check_all_peers_initially_connected(&mut self) -> bool {
```
arminsabouri: This seem to get repeated in bunch of places. Can we abstract this away to a util?

crates/consensus/authority/src/utils.rs
```diff
@@ -20,6 +26,48 @@ use reth_rpc_types::BlockHashOrNumber;
...
+    BadMerkleRoot,
+}
+
+pub(crate) fn compute_utxo_merkle_root(
+    peer_utxos: &[Utxo],
+) -> Result<Option<bitcoin::hashes::sha256::Hash>, UtxoError> {
+    let mut utxos = peer_utxos
+        .iter()
+        .map(|u| {
+            // TODO: do we need to consider the other components of the utxo
```
arminsabouri: I dont see any reason why we need to?

crates/consensus/authority/src/utxo_syncer.rs
```diff
@@ -0,0 +1,275 @@
...
+            error!(target: "consensus::authority::utxo_syncer::get_peer_messages_rx", "Failed to send GetPeerMessagesStream frost command {}", e);
+        }
+        let peer_messages_rx = peer_messages_rx.await.map_err(|e| {
+            error!(target: "consensus::authority::utxo_syncer::get_peer_messages_rx", "Error getting receiver handle = {:?}", e);
+            Error::FrostRecv(e)
+        })?;
+        Ok(peer_messages_rx)
+    }
+
+    async fn get_authority_peers(&self) -> Vec<PeerId> {
```
arminsabouri: Can we add util to storage for this?
`get_authority_peer_ids`

crates/consensus/authority/src/utxo_syncer.rs
```diff
@@ -0,0 +1,275 @@
...
+            error!(target: "consensus::authority::utxo_syncer::get_all_connected_authority_peers", "Failed to send GetAllConnectedPeers frost command {:?}", e);
+        }
+        let connected_peers = connected_peers_rx.await.map_err(|e| {
+            error!(target: "consensus::authority::utxo_syncer::get_all_connected_authority_peers", "Error getting receiver handle = {:?}", e);
+            Error::FrostRecv(e)
+        })?;
+        Ok(connected_peers)
+    }
+
+    pub async fn start_task(&mut self) {
```
arminsabouri, line 114: This is more a architecture question. Does the utxo syncing need to be a continous task? To me its more of a one shot event that happens after the main block syncing finishes. Maybe the best way to go about this is to have the task here await a pipeline event (not sure which one exactly). Once it recieves that event it begins. After it finishes it should emit its own event. 

Additionally there should be some kind of is_syncing bool on some kind of handler (similar to how network handler) has it setup. That way other tasks (block_builder, block_fetcher, pbft, frost) know when to perform their respective tasks.

crates/consensus/authority/src/utxo_syncer.rs
```diff
@@ -0,0 +1,275 @@
...
+        info!(target: "consensus::authority::utxo_syncer::start_task", "Starting Utxo Sync Task");
+
+        // await all peers to be connected
```
arminsabouri, line 117: Don't we only need one peer to sync utxos off?

crates/consensus/authority/src/utxo_syncer.rs
```diff
@@ -0,0 +1,275 @@
...
+    pub(crate) frost_handle: ToFrostMan,
+    /// Shared authority storage
+    pub(crate) storage: Storage,
+    /// Btc Server client
+    pub(crate) btc_server: BtcServerExtendedClient,
+    /// compressor
+    pub(crate) compressor: Compressor,
+}
+
+impl<ToFrostMan> UtxoSyncTask<ToFrostMan>
```
arminsabouri, line 47: Nit: a light weight description for what this task is

crates/net/network/src/frost/mod.rs
```diff
@@ -90,6 +93,29 @@ impl fmt::Display for DkgResponse {
     }
 }
 
+/// Response structure for internal communication
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct UtxoSetResponse {
+    /// the utxo set request sender
+    pub sender: PeerId,
+    /// the utxo set target
+    pub target: PeerId,
```
arminsabouri: I think sender is already part of the payload recieved that is delivered to the peer. Im not sure we need to include it here as well. Samething goes for target. if you recieved a UTXO set payload then you were the target. That we also dont need to perfom any additional validation on this payload (is the sender the actual sender)

arminsabouri: Unless im missing something I think we can just do with `data`

crates/consensus/authority/src/utxo_syncer.rs
rwlockbg: The idea of having a background task here is that other peers might want to get our own utxos from us. To me it kinda make sense to have this task running in a loop throughout the entire life of thecycle fo the node as reconnecting or newly joining nodes might send a  message to us at any point asking for the utxo set.

Reg. you second comment, yes, I agree that is_syncing might also check the utxo syncing state using some sort of atomic boolean. I need to think about it.

crates/consensus/authority/src/utxo_syncer.rs
rwlockbg: As described above, the task here is rather serving throughout the node life cycle other peers requesting our own utxos. We need to have a continuous task serving those upon request.

bin/btc-server/src/main.rs
rwlockbg: renamed

bin/btc-server/proto/btc_server.proto
rwlockbg: I mean output_scriptpubkey, renamed as suggested

crates/consensus/authority/src/utils.rs
rwlockbg: me neither, was just a comment to revisit it .I have removed it now.

crates/consensus/authority/src/utxo_syncer.rs
rwlockbg: done

crates/consensus/authority/src/block_fetcher.rs
rwlockbg: done, moved to utils

bin/btc-server/src/main.rs
rwlockbg: This was already broken down by Sam this way. I think he wanted to have them separately as to avoid the double serialization / deserialization that is to follow. Prost -> Vec<u8> and vice versa. I could change it if you want

crates/consensus/authority/src/utxo_syncer.rs
rwlockbg: done

crates/net/network/src/frost/mod.rs
rwlockbg: done   good point

bin/btc-server/src/database.rs
```diff
@@ -44,6 +47,29 @@ impl Utxo {
...
+        Utxo::new(
+            OutPoint::new(txid, vout),
+            TxOut {
+                value: Amount::from_sat(value.output_value as u64),
+                script_pubkey: script.into(),
+            },
+            if value.eth_address.is_empty() {
+                None
+            } else {
+                Some(parse_eth_address(value.eth_address).unwrap())
```
arminsabouri, line 68:
```suggestion
                // FIXME: remove the unwrap
                Some(parse_eth_address(value.eth_address).unwrap())
```

bin/btc-server/src/main.rs
arminsabouri: I think its fine for now.

crates/consensus/authority/src/block_fetcher.rs
```diff
@@ -99,6 +137,188 @@ where
             network_client,
             network_handle,
             client,
+            frost_handle,
+            compressor: Compressor::new(),
+        }
+    }
+
+    fn get_best_block_edh_from_peers(&self) -> Result<Option<ExtraDataHeader>, Error> {
+        let best_block_number = self.client.best_block_number().map_err(Error::Provider)?;
```
arminsabouri, line 146: is client here the database provider? In which case I dont think network requests are being made

crates/consensus/authority/src/block_fetcher.rs
```diff
@@ -108,12 +328,35 @@ where
         let consensus = Arc::new(self.consensus.clone());
         let full_block_client = FullBlockClient::new(self.network_client.clone(), consensus);
 
+        // await all peers to be connected
+        loop {
+            if check_all_peers_initially_connected(&self.frost_handle).await {
+                break;
+            }
+            // short sleep
+            tokio::time::sleep(std::time::Duration::from_millis(500)).await;
```
arminsabouri: We shouldnt need to wait for alll peers to be connected until we can fetch and process blocks. If one peer is down we dont want the whole network to halt. Thats why we have a consensus protocol that can tolerate 2/3 of the network being byzantine

crates/consensus/authority/src/block_fetcher.rs
```diff
@@ -108,12 +328,35 @@ where
...
+        }
+
+        // ensure the node is not syncing
+        loop {
```
arminsabouri, line 332: This check already exists below. Do we want to recreate it here?

crates/consensus/authority/src/block_fetcher.rs
```diff
@@ -99,6 +137,188 @@ where
...
+        // get peer messages receiver
+        let mut peer_messages_rx = match self.get_peer_messages_rx().await {
+            Ok(peer_messages_rx) => peer_messages_rx,
+            Err(e) => {
+                error!(target: "consensus::authority::block_fetcher::sync_utxo_set", "Error getting peer messages receiver = {:?}", e);
+                panic!("Error getting peer messages receiver = {:?}", e);
+            }
+        };
+
+        loop {
```
arminsabouri, line 194: Does this function need a loop? I only say that b/c the caller is also in a loop.
maybe we just return out of this function. If the UTXO set is still not sync'd the caller can call back into this function

arminsabouri (commented):
Starting to look block fetcher. Really impressive code! 
Just had a few comments / questions

crates/consensus/authority/src/block_fetcher.rs
rwlockbg: It is, but the idea was that before that yo uhave some code that checks for synced status. Once the node is synced, the database provider will get the same best block edh as the network client, right ?  And yes, it is the db client used here

rwlockbg: We could also use the network client here, I suppose ?

bin/btc-server/src/database.rs
rwlockbg: added

crates/consensus/authority/src/block_fetcher.rs
rwlockbg: yes, my  bad, good point. Code removed

crates/consensus/authority/src/block_fetcher.rs
rwlockbg: The idea was that we sync initially, the utxos are then allowed to sync if needed. Then we loop and fetch blocks, but what if we desync again ,would not we want to wait to be synced again before we continue receiving new blocks ? I discussed this with Steven some time ago and this was his suggestion. Pls advise here

crates/consensus/authority/src/block_fetcher.rs
rwlockbg: Again, we discussed with Steven and his suggestion was to not let go the block fetcher fetch unless we have our utxos synced here. The loop is there to encapsulate this desired effect as long as its not finalized.

crates/consensus/authority/src/block_fetcher.rs
arminsabouri: yeah yes that makes sense. I would revise the function name then to `get_best_block_edh`.

crates/consensus/authority/src/utxo_syncer.rs
arminsabouri: Ah ok I understand the reasoning here a bit more.

bin/test-suite/src/suite/consensus/frost/test_utxo_sync.rs
```diff
@@ -0,0 +1,161 @@
...
+            .and_then(|servers| servers.iter().nth(instance as usize).map(|val| val.port))
+            .ok_or_else(|| Error::InvalidBtcServerPort)?;
+        let c = client::BtcServerClient::connect(format!("http://localhost:{}", port))
+            .await
+            .map_err(Error::ServerConnect)?;
+        clients.push(c);
+    }
+
+    // run the dkg
+    let _ = do_dkg(&mut clients).await?;
```
arminsabouri, line 57: Dkg should be happening outside the context of a specific test.

bin/test-suite/src/suite/consensus/frost/test_utxo_sync.rs
```diff
@@ -0,0 +1,161 @@
...
+            .get_utxo_merkle_root(tonic::Request::new(client::Empty {}))
+            .await
+            .unwrap()
+            .into_inner()
+            .merkle_root;
+        hashset.insert(utxo_set);
+    }
+    // all the btc_servers should have the same merkel commitment to the utxo set
+    assert_eq!(hashset.len(), 1);
+    assert_ne!(first_utxo_commitment, hashset.iter().next().unwrap().to_owned());
```
arminsabouri, line 158: Should we be re-connecting the one peers after this and checking their UTXO set is the same as the others?

arminsabouri (commented):
I had some questions around the int. test

bin/btc-server/src/database.rs
```diff
@@ -387,14 +414,33 @@ impl Db {
         }
     }
 
+    /// Resetting all utxos
+    pub fn reset_utxos(&self, utxos: &[Utxo]) -> Result<(), Error> {
+        self.clear_utxos()?;
+        for utxo in utxos.iter() {
+            self.store_utxo(utxo)?;
+        }
```
arminsabouri, line 422: Is there a way to use serde for a atomic operation here?

bin/test-suite/src/suite/consensus/frost/test_utxo_sync.rs
```diff
@@ -0,0 +1,161 @@
...
+    // create NUM_UTXOS pegins
+    for _ in 0..NUM_UTXOS {
+        let eth_address = ethers::core::types::Address::random();
+        pegins.eth_addresses.push(eth_address);
+        pegins.txids.push(rand::random::<[u8; 32]>());
+    }
+
+    // create btc server clients
+    let mut clients: Vec<BtcServerClient<Channel>> = vec![];
+    for instance in 0..suite.global_context.instances {
```
arminsabouri, line 43: BTC server clients should be accessible via `self.local_context.btc_servers`

crates/consensus/authority/src/block_fetcher.rs
rwlockbg: renamed

bin/btc-server/src/database.rs
rwlockbg: This is addressed in another MR: https://github.com/botanix-labs/Macbeth/pull/319

arminsabouri:
Closing 
Tracking this effor here #342

PR #309, opened 2024-07-23
fix(health_checks): pull peer socket address from network manager / fed config

rwlockbg:
- sockets are directly being taken from the networkHandle
- socket being transported on PeerConfirmed is removed as its not needed and reduces message size
- I could not get the peer_id , but have dropped all other stuff like pub keys and authority index since we can get them all based on the auth. confuguration
- added some clippy cleanup

### DIFF
--- a/crates/consensus/authority/src/healthcheck_task.rs
+++ b/crates/consensus/authority/src/healthcheck_task.rs
@@ -1,5 +1,6 @@
 use std::{
     collections::{HashMap, HashSet},
+    net::SocketAddr,
     sync::Arc,
     time::Instant,
 };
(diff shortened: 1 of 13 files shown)


crates/net/network/src/frost/protocol.rs
```diff
@@ -308,34 +299,26 @@ impl Stream for FrostProtoConnection {
...
-                if let Err(e) = peer_message_forwarder.send(FrostProtocolEvent::PeerConfirmed(
-                    peer_id,
-                    authority_index,
-                    peer_socket_addr,
-                )) {
-                    error!(target: "network::frost::protocol", "Failed to forward received pong message from peer id {:?}, auth index = {}, socket = {:?}. Error = {:?}", peer_id, authority_index, peer_socket_addr, e);
+            FrostProtoMessageKind::PingMessage(peer_id, authority_index) => {
+                info!(target: "network::frost::protocol", "Received ping message from peer with id = {:?}, auth index = {} Replying with pong...", peer_id, authority_index);
+                if let Err(e) = peer_message_forwarder
+                    .send(FrostProtocolEvent::PeerConfirmed(peer_id, authority_index))
```
arminsabouri: One Question about this. Cant we pull peer id off the request it self? Following this, can't we then find that authorities index from their pk

rwlockbg: yes, we can, I am adding that.

scottmillner (dismissed):

crates/net/network/src/frost/protocol.rs
rwlockbg: I could not get the peer_id , but have dropped all other stuff like pub keys and authority index since we can get them all based on the auth. confuguration

rwlockbg: There is no connection request available in our manager.rs. All we have are the messages which are being received in a deserialized by rlpx format. At this point the peer id is not retrievable as we dont have enough information hence we need to pass it over in the message.

scottmillner (approved):

crates/consensus/authority/src/healthcheck_task.rs
```diff
@@ -115,25 +116,33 @@ where
...
-                panic!("Error getting receiver handle");
-            }
-        };
+        // get all connected peers (could be authority and none-authority)
+        let all_connected_peers =
+            self.network_handle.get_all_peers().await.ok().unwrap_or_default();
+
+        // get all authority peers in the federation (at this point we must be connected to all of
+        // them)
+        let authority_peers = self
```
arminsabouri, line 125: Note for a later refactor. This code seems repeated in a bunch of places. we can probably just do this once in StorageInner::new() and access it off storage.

crates/net/network/src/frost/protocol.rs
```diff
@@ -308,34 +291,25 @@ impl Stream for FrostProtoConnection {
...
-                if let Err(e) = peer_message_forwarder.send(FrostProtocolEvent::PeerConfirmed(
-                    peer_id,
-                    authority_index,
-                    peer_socket_addr,
-                )) {
-                    error!(target: "network::frost::protocol", "Failed to forward received pong message from peer id {:?}, auth index = {}, socket = {:?}. Error = {:?}", peer_id, authority_index, peer_socket_addr, e);
+            FrostProtoMessageKind::PingMessage(peer_id) => {
+                info!(target: "network::frost::protocol", "Received ping message from peer with id = {:?} Replying with pong...", peer_id);
+                if let Err(e) =
+                    peer_message_forwarder.send(FrostProtocolEvent::PeerConfirmed(peer_id))
```
arminsabouri, line 297: Just a question not a request for change but is the peer id being used anywhere?

arminsabouri (approved):

PR #311, opened 2024-07-26
fix(poa_node): create edh limit on max inputs

rwlockbg:
Task: https://github.com/botanix-labs/botanix/issues/537
Int. test: https://github.com/botanix-labs/botanix/issues/534


scottmillner (dismissed):

bin/btc-server/src/cordinator.rs
```diff
@@ -197,7 +201,15 @@ impl App {
...
             })?;
         // Filter the ones that are still pending and conflict with pending txs.
-        let pending_inputs = self.txindex.lock().await.pending_inputs();
+        let pending_inputs: HashSet<OutPoint> = self
+            .txindex
+            .lock()
+            .await
+            .pending_inputs()
+            .iter()
+            .take(max_allowed_inputs)
```
arminsabouri: The problem here could be that certain UTXOs may get starved. 
Is there anyway to present BDK's coinselection algorithm the full utxo set and have the algorithm choose the best tx that only spends 100 inputs?

rwlockbg: I am not sure about this :) Maybe you could have a quick look and give me some guidelines

arminsabouri: Yeah I can help with this

arminsabouri: There doesnt seem like there is a way to restrict the size or number of inputs using the BDK coin selection algos. 
Perhaps we can select the first lowsest value UTXOs first. I'm not sure yet. It may be the case that we are also trying to optimize for something that isnt a problem yet.

arminsabouri:
Can we add https://github.com/botanix-labs/botanix/issues/534 as part of this PR

rwlockbg:
> Can we add [botanix-labs/botanix#534](https://github.com/botanix-labs/botanix/issues/534) as part of this PR

Yes, I will add it there.

crates/consensus/authority/src/lib.rs
```diff
@@ -72,7 +72,7 @@ pub use builder::AuthorityConsensusBuilder;
 /// This was calulated with the following formula
 /// version + optional_fields bitmask + signers pk + witness (vec of sigs) + blockhash +
 /// utxo_commit + block_witness + agg_pk For specific details see [ExtraDataHeader]
-pub const MAX_EDH_SIZE: usize = 8005;
+pub const MAX_EDH_SIZE: usize = 5 * 8005; // 8005 - 100 spents
```
arminsabouri: Not sure I understand this calculation?

arminsabouri (changes requested):
Some things to fix. I am going to assign over to my self as rwlock is out of town this week

bin/test-suite/src/suite/consensus/frost/test_edh_size_limit.rs
```diff
@@ -0,0 +1,327 @@
...
+        let _ = bitcoind_rpc.unload_wallet(Some(&wallet));
+    }
+    let create_res = bitcoind_rpc.create_wallet(BITCOIND_WALLET_NAME, None, None, None, None);
+    if create_res.is_err() {
+        // wallet already exists, load wallet
+        let _ = bitcoind_rpc.load_wallet(BITCOIND_WALLET_NAME);
+    }
+    let address =
+        bitcoind_rpc.get_new_address(None, None).expect("get new address").assume_checked();
+    bitcoind_rpc.generate_to_address((NUM_PEGINS).into(), &address).expect("generate to address");
```
arminsabouri: Do you mean to create `NUM_PEGINS ` blocks? 
101 should be sufficient

bin/btc-server/src/cordinator.rs
rwlockbg: @scottmillner @0xBEEFCAF3 Do we still want this change that I originally did or not ? Acc. to Armins comment, it looks as if we rather not cap here ?

scottmillner: should we revisit the value of this ticket and what we gain from it?

rwlockbg: removed

arminsabouri:
[fix: remvoved utxo capping (reverted back)](https://github.com/botanix-labs/Macbeth/pull/311/commits/1d45358509cfcd8dc3c0657a9ffe2c9fbf80073a) 
Can we just drop the original commit instead of having two commits: one that introduces a change and another one that removes that change

arminsabouri:
Int. test going forward: it would be great if the test being introduced in this PR can be refactored to do a multi pegin / multi pegout.

scottmillner:
superceded by: https://github.com/botanix-labs/Macbeth/pull/362

PR #312, opened 2024-07-26
fix: should check btc_server auth status after connecting

rwlockbg:
https://github.com/botanix-labs/botanix/issues/567


arminsabouri (approved):
tACK

PR #313, opened 2024-07-26
Fix/arm/pbft pre commits resets

arminsabouri:


crates/consensus/authority/src/pbft.rs
```diff
@@ -43,6 +43,11 @@ impl PbftCommitmentCriteria {
     /// n being the number of authorities
     #[inline]
     pub(crate) fn min_pre_commitments(n: u16) -> u16 {
+        // In the case of a two or three person federation we will require 2 pre-commitments
+        // This will be mostly used in local dev setups
+        if n == 2 || n == 3 {
```
rwlockbg, line 48: I think this can be written in a more rusty way like : if (2..=3).contains(n) { 2 } else { n- 2 }

rwlockbg: you can just ignore the above, it looks silly

crates/consensus/authority/src/pbft.rs
```diff
@@ -536,21 +541,52 @@ where
...
+            // i.e recieve the fee reward.
+            let number_of_valid_sigs =
+                saved_block.header().check_authority_sig_add(&self.config.authorities)?;
+            info!(target: "consensus::authority::pbft::init_block_proposal" ,"number of valid sigs: {}", number_of_valid_sigs);
+            if number_of_valid_sigs >=
+                PbftCommitmentCriteria::min_commitments(self.config.authorities.len() as u16)
+            {
+                info!(target: "consensus::authority::pbft::init_block_proposal" ,"We have enough commitments, time to produce a block");
+                let block_witness =
+                    saved_block.header().get_block_witness()?.expect("set the witness above");
```
rwlockbg, line 576: maybe put a TOOD here reg. later handling the expect

rwlockbg (dismissed):

crates/consensus/authority/src/pbft.rs
arminsabouri: Hmm to me this is a bit harder to read tbh

crates/consensus/authority/src/pbft.rs
arminsabouri: done

PR #315, opened 2024-07-29
Fix/arm/pbft pre commits

arminsabouri:
Best way to review is commit by commit

The main changes are
* Allowing nodes to pre-commit even tho they are in a commiting state. the reason behind this is other nodes can reset or could just be poorly connected so they may still be stuck on pre-commits. If a suffecient # of nodes progress to commits and a number of nodes are still waiting for precommits then we will stall. For example if n = 6 and all but one node is waiting for commits then we will stall. Furthermore there is no hard to precommiting to the same block twice. 
* Swapping out the BTE `contains` with block_by_hash. Im not sure if contains was doing what I thought it was doing. It may only look up a block in tree not the canon blockchain


arminsabouri:
Looks like I have a bug. e2e are failing

scottmillner (approved):

PR #316, opened 2024-07-29
added ConnEstablished failed event, some cleanup

rwlockbg:


crates/net/network/src/frost/protocol.rs
```diff
@@ -109,14 +109,14 @@ impl ConnectionHandler for FrostConnectionHandler {
...
-            })
-            .ok();
+        let connection_established_event = FrostProtocolEvent::ConnectionEstablished {
+            direction,
+            peer_id,
+            to_connection: remote_peer_tx,
+        };
+        if let Err(e) = self.state.events.send(connection_established_event) {
+            error!(target: "network::frost::protocol::into_connection", "Failed to send ConnectionEstablished event: {:?}", e);
+        }
```
arminsabouri, line 119: Do we want to return with an error here? or just log?

arminsabouri (approved):

crates/net/network/src/frost/protocol.rs
rwlockbg: This is the the most important event, I would say yes, it needs to be an error as this is place in code where the rlxp protocol makes the durable connection to the other peer.

arminsabouri: Will we violate the trait if we return an Result here?

arminsabouri: Lets merge and open a ticket for this. Will be good to just start with some logging

PR #319, opened 2024-07-30
refactor(btc_server): send_pegin notification rpc should be a bulk operation

rwlockbg:
Ticket description: https://github.com/botanix-labs/botanix/issues/368


arminsabouri (dismissed):
Nice!!

arminsabouri:
@rwlockbg looks like int. found a bug in the test `test_many_inputs_signing `

arminsabouri:
Problem is probably in tryFrom conversion going from RPCUtxo -> UTXO

bin/btc-server/src/database.rs
```diff
@@ -387,6 +400,23 @@ impl Db {
         }
     }
 
+    fn store_utxos_atomically(&self, utxos: &[&Utxo]) -> Result<bool, Error> {
+        self.utxos
+            .transaction(|tx| {
+                for utxo in utxos.iter() {
+                    let op = utxo.outpoint;
+                    if tx.get(op.to_bytes())?.is_none() {
```
arminsabouri: Shouldnt we be checking if utxo exists in the database as whole? Not just this db transaction

bin/btc-server/src/database.rs
```diff
@@ -387,6 +400,23 @@ impl Db {
         }
     }
 
+    fn store_utxos_atomically(&self, utxos: &[&Utxo]) -> Result<bool, Error> {
+        self.utxos
+            .transaction(|tx| {
```
arminsabouri: Maybe rename this to database_tx so it doesnt get confusing with other variables in this codebase called tx

bin/btc-server/src/database.rs
```diff
@@ -487,6 +521,60 @@ impl From<Error> for tonic::Status {
...
+            .map_err(|_| Error::RpcToDbMap("Unparsable Txid".to_string()))?;
+        let vout = outpoint.vout;
+
+        // txout
+        let tx_out = value.output.ok_or_else(|| Error::RpcToDbMap("TxOut is None".to_string()))?;
+        let script_pubkey = tx_out
+            .script_pubkey
+            .ok_or_else(|| Error::RpcToDbMap("Script Pub Key is None".to_string()))?;
+        let script = Script::from_bytes(&script_pubkey.script);
+        let tx_out_val = Amount::from_sat(tx_out.value as u64);
```
arminsabouri: is the u64 conversion neccecary? Can't we define value to be u64 in the protobuf definitions?

bin/btc-server/src/database.rs
rwlockbg: done

bin/btc-server/src/database.rs
rwlockbg: The value is infact u64. I have removed the unneeded conversion. thx

bin/btc-server/src/database.rs
rwlockbg: Can you elaborate more on that here pls

arminsabouri (approved):

PR #322, opened 2024-07-31
refactor: pass ntp-server from cli

scottmillner:


bin/reth/src/commands/poa/mod.rs
```diff
@@ -99,6 +99,10 @@ pub struct PoaNodeCommand<Ext: clap::Args + fmt::Debug = NoArgs> {
     #[arg(long, value_name = "IS_TESTNET", default_value = "true")]
     pub is_testnet: bool,
 
+    /// The NTP server url
+    #[arg(long, value_name = "NTP_SERVER", default_value = "time.cloudflare.com")]
+    pub ntp_server: String,
```
rwlockbg, line 109: Can we have this as a Url from the url crate ?

rwlockbg (dismissed):

bin/reth/src/commands/poa/mod.rs
arminsabouri: +1 to using a URL. There should already be a parser for this

arminsabouri (approved):

arminsabouri:
We can write a custom parser for this in the future

PR #324, opened 2024-07-31
fixed healthcheck task

rwlockbg:


arminsabouri (dismissed):
ApproachACK

arminsabouri:
@rwlockbg I think the block builder int. is experiencing the same bug you described locally. PBFT doesnt seem to be kicking in. Rather it seems to not submit any txs. Need to debug further

scottmillner (dismissed):

test_runner.sh
```diff
@@ -5,7 +5,7 @@ set -o pipefail
 
 # Define the list of strings
 # temporarily removed "e2e_peer_disconnect" since healthcheck service is disabled
```
scottmillner: nit: remove comment

rwlockbg: removed, thanks

rwlockbg:
@0xBEEFCAF3 @scottmillner I have modified the integration test now so it physically disconnects the peers upon request. Previously it was just removing them from the peers list. Since I am not sure how to reconnect with them now (the network handler does not have such a command), I have asked Matthias from Reth for help.

arminsabouri (approved):
tACK 3d58d8d372b6ae21e577596f06e86e66b694798b

PR #326, opened 2024-07-31
Fix/arm/bitcoind sync

arminsabouri:


scottmillner (approved):

crates/btc-wallet/src/bitcoind.rs
```diff
@@ -81,19 +81,14 @@ impl BitcoindClient {
     }
 
     pub async fn is_synced(&self) -> Result<bool, BitcoindError> {
-        match self.rpc.get_index_info().map_err(BitcoindError::BlockIndexStatusFailed) {
```
rwlockbg, line 84: Any reason why this used to work and now it does not anymore ?

rwlockbg: Can we maybe still keep the old code and if it fails, we try to call `get_blockchain_info` ? 
Or is the latter guaranteed to work with all bitcoinds ?

arminsabouri: The old code was giving false positives. I think block index is available before IBD is finished

rwlockbg: fair enough. Approved

rwlockbg (approved):

PR #328, opened 2024-08-01
fix(poa): added deployed minting contract path

rwlockbg:


scottmillner (dismissed):

bin/reth/src/commands/poa/mod.rs
```diff
@@ -907,6 +913,20 @@ where {
 
         let _ = ext.on_node_started(components);
 
+        // check Minting.sol deployed bytecode matches known bytecode
```
arminsabouri: This piece of code should be commented out somewhere above. Can we comment that back in?
Its further up in poa/mod.rs b/c we shouldnt start the consensus engine or the rpc engine in case this check fails

arminsabouri (changes requested):

bin/reth/src/commands/poa/mod.rs
rwlockbg: done

birozuru:
all good?

arminsabouri (commented):
Do we need to update the federation template config?

scottmillner:
> Do we need to update the federation template config?

done!

arminsabouri (approved):

PR #330, opened 2024-08-02
fix(btc-server):  fix for stream no longer needed err message

rwlockbg:


bin/btc-server/src/extended_client.rs
```diff
@@ -75,7 +80,15 @@ pub struct BtcServerExtendedClient {
 impl BtcServerExtendedClient {
     /// Create a new Bitcoin Server Client with extended authentication credentials
     pub async fn new(url: String, jwt_secret: Option<JwtSecret>) -> Result<Self, GrpcClientError> {
-        let client = BtcServerClient::connect(url).await.map_err(GrpcClientError::Transport)?;
+        let uri = url.parse::<Uri>().map_err(|e| GrpcClientError::InvalidUri(e.to_string()))?;
+        let chan = tonic::transport::Channel::builder(uri)
+            .timeout(Duration::from_secs(20))
+            .connect_timeout(Duration::from_secs(20))
```
arminsabouri, line 86: Should this not be shorter? Im just thinking if a block builder fails to make_tx < 20 seconds they will produce an out of turn block?

rwlockbg: This has nothing to do with block times, these are the same timeouts we apply to the btc server in order to keep it idle to pick up connections and not timeout in case the request takes longer. Not sure what the defaults are but we allow for longer timeouts on connections.

PR #337, opened 2024-08-07
test: add test for botanix_consensus_package()

scottmillner:


crates/primitives/src/header_ext.rs
```diff
@@ -805,4 +810,16 @@ mod tests {
...
+    #[test]
+    fn test_botanix_consensus_package() {
+        let mut header = Header::default();
+        let edh = ExtraDataHeader::default();
+        header.add_extra_data_header(&edh);
+        let btc_network = bitcoin::Network::Testnet;
+        let bitcoind_factory = MockBitcoindFactory::new(BitcoindConfig::default());
+
+        let res = header.botanix_consensus_package(btc_network, bitcoind_factory);
+        assert!(res.is_ok());
```
arminsabouri, line 828: Nit: 
* Are there any more assertions worth doing here? (network is correct, block header is what we expect, etc...)
* Is it worth asserting for any error cases?

scottmillner: added assertions to ensure result is what we expect

as far as error cases don't think we gain much bc:

- the method is just deserializing the edh and we have good test coverage around that
- then it's just side effects (calling bitcoind)
- then stitching together the bitcoind call results

the assertions cover the stitching part

arminsabouri (approved):

PR #338, opened 2024-08-09
fix: kill btc servers regardless of int. test results

arminsabouri:


birozuru (dismissed):
Looks good. 
The script on the vm will then act as a back up to this just in case.

PR #346, opened 2024-08-15
feature: add chain version to edh

scottmillner:


crates/primitives/src/extra_data_header.rs
```diff
@@ -117,6 +122,8 @@ impl ExtraDataHeader {
     /// Create a new extra data header
     pub fn new(
         version: u32,
+        // Chain version that determines the valid chain
+        chain_version: u32,
```
arminsabouri, line 126: Nit: instead of it being passed in new it can be hardcoded below. That way we have less code changes all around

```rust
Self {
            version,
            chain_version: CHAIN_VERSION,
           ...
```

scottmillner: one benefit i do see of having it as a const is when writing tests that should fail
for example: `chain_version: CHAIN_VERSION + 1`
then we don't have to update those tests at all if the version changes

scottmillner: and if the version changes we only have to update in one place
so in the long run, i think having a constant reduces the amount of code changes

arminsabouri: makes sense. Thanks!

arminsabouri (approved):

rwlockbg (approved):

PR #356, opened 2024-08-21
added cargo audit and coverage to ci/cd

rwlockbg:


crates/revm/src/factory.rs
```diff
@@ -45,12 +45,6 @@ impl<EvmConfig: ConfigureEvm, BF> EvmProcessorFactory<EvmConfig, BF> {
         self
     }
 
-    /// adds bitcoin network information
-    pub fn with_bitcoin_config(mut self, btc_network: bitcoin::Network) -> Self {
-        self.btc_network = Some(btc_network);
-        self
-    }
-
```
arminsabouri: This should be remove now on main. Just need to rebase

PR #358, opened 2024-08-22
620 fix pegout fees should be taken out of pegout request

arminsabouri:
## Description
Users need to paying for L1 pegout fees. If fees are deducted from the federation wallet change outputs then we will slowly de-peg the sidechain. 
The solution is simple. Perform coin selection and create an optional change output. Afterwards deduct each pegout output by the absolute fee / pegouts.len. And then increase the change output value by the absolute value.

### DIFF
--- a/bin/btc-server/src/coin_selection.rs
+++ b/bin/btc-server/src/coin_selection.rs
@@ -6,7 +6,7 @@ use bdk::{
 };
 use bitcoin::{
     psbt::{Error as PsbtError, ExtractTxError, Psbt},
-    Amount, FeeRate, OutPoint, ScriptBuf, Transaction, TxOut,
+    Amount, FeeRate, OutPoint, ScriptBuf, TxOut,
 };
 use reth_btc_wallet::TAPROOT_KEYSPEND_SATISFACTION_WEIGHT;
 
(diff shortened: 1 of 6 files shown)


bin/btc-server/src/coin_selection.rs
```diff
@@ -0,0 +1,250 @@
...
+        if let Some(mut ch) = change.clone() {
+            ch.value += absolute_fee;
+            Some(ch)
+        } else {
+            None
+        }
+    };
+
+    let updated_psbt =
+        reth_btc_wallet::transaction::create_psbt(selected_inputs, pegouts, updated_changed);
```
SatwikPrabhu, line 138:
```suggestion
    // Untested
    let updated_change = match change.clone() {
        Some(ch) => Some(ch.value + absolute_fee),
        None => None,
    };
    
    let updated_psbt =
        reth_btc_wallet::transaction::create_psbt(selected_inputs, pegouts, updated_change);
```

arminsabouri: Sorry for the super late reply. The problem with this change is that it only returns the updated change amount. I actually need the full `TxOut` with the update value.

SatwikPrabhu: Yep. Got it. Missed that earlier.

SatwikPrabhu (approved):

PR #366, opened 2024-08-23
Txs coming into reth mempool should get proxied to cbft

scottmillner:


crates/transaction-pool/src/pool/mod.rs
```diff
@@ -423,6 +424,30 @@ where
...
+
+                // create TransactionSigned from Transaction
+                let transaction_signed =
+                    transaction.transaction().to_recovered_transaction().into_signed();
+                let mut tx_bytes = Vec::new();
+                transaction_signed.encode_enveloped(&mut tx_bytes);
+
+                // spawn a task to send the tx to CometBFT mempool
+                // we don't care about the result since CometBFT will send a check_tx message
+                tokio::spawn(async move {
```
arminsabouri: I dont know if this is neccecary. the async method will respond immediately regardless of the checkTx response

arminsabouri (changes requested):

crates/transaction-pool/src/pool/mod.rs
scottmillner: yeah i thought i had to call `.await` since the method was async but can drop that and ignore the result

done!

arminsabouri (approved):

PR #368, opened 2024-08-26
Feature/cbft extended vote 642

scottmillner:


crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -312,4 +327,133 @@ where
...
+    #[error("invalid version")]
+    /// Invalid ExtendedVote version
+    InvalidVersion,
+}
+
+pub struct ExtendedVote {
+    pub version: Version,
+    pub bitcoin_block_hash: bitcoin::hash_types::BlockHash,
+    pub aggregated_public_key: secp256k1::PublicKey,
+}
```
arminsabouri: How do you feel about this living in its own file under the comet_bft dir

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -312,4 +327,133 @@ where
...
+    #[error("invalid data format")]
+    /// Invalid data format
+    Decoding(#[from] encode::Error),
+    #[error("invalid version")]
+    /// Invalid ExtendedVote version
+    InvalidVersion,
+}
+
+pub struct ExtendedVote {
+    pub version: Version,
```
arminsabouri: can this just be u16?

scottmillner: yeah maybe a bit overkill with the struct but i like having a default method i can call so if the version is ever changed, i only have to update the default method...i use it in tests as well

made it u16 and added version_default() on ExtendedVote

crates/consensus/authority/src/comet_bft/abci.rs

arminsabouri (approved):

arminsabouri:
@scottmillner I hope you don't mind I truncated some of these commits. Things like fixiing git conflicts

PR #376, opened 2024-08-29
Feature/arm/abci commit

arminsabouri:


scottmillner (approved):

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -470,40 +479,63 @@ where
...
-            )))
-            .expect("to send");
-
         let mut exec_results = vec![];
-        for _ in 0..sealed_block_with_peg.block().body.len() {
-            // TODO this needs to be populated with the actual results of the execution
+        for tx in sealed_block_with_peg.block().body.iter() {
             // https://docs.cometbft.com/v0.38/spec/abci/abci++_app_requirements#transaction-results
-            exec_results.push(ExecTxResult::default());
+            exec_results.push(ExecTxResult {
```
scottmillner, line 490: do we not want to serialize the tx and set that as the value in `ExecTxResult.data`?
the underlying question is how does comet set `Block.data` which should be the `txs` according to the docs?

arminsabouri: notsure what data is used for. I'd have to take a deeper look at that

PR #377, opened 2024-09-03
feature(pegout): metadata should be pegout version

scottmillner:


crates/botanix-lib/src/mint_validation.rs
```diff
@@ -294,4 +311,7 @@ mod test {
             Address::from_str("0xa65812bac44dadb79c3e4930dbd98d5a75376b2a").unwrap()
         );
     }
+
+    // TODO: add decode_log_payload for pegout
```
arminsabouri, line 314: Nice todo !

crates/botanix-lib/src/mint_validation.rs
```diff
@@ -231,6 +231,23 @@ pub fn try_parse_burn_event(
...
+    if metadata.len() != 1 {
+        return Err(ParseBurnEventError::InvalidPegoutData(PegoutDataError::Invalid(
+            "invalid metadata length",
+        )));
+    }
+
+    if metadata[0] != PegoutData::version() {
+        return Err(ParseBurnEventError::InvalidPegoutData(PegoutDataError::Invalid(
+            "invalid metadata version",
+        )));
```
arminsabouri: If we upgrade the network with v1 then some older client will always reject any version that is not v0. As long as a supermajority upgrade then we can upgrade to v1. But this would be a hardfork. Instead for a softfork what we want is to provide the same fields (destination, amount) and append future additional fields. So in short It should be fine to not check the version here. In the future upgraded client will know that there are additional fields pass the old (maybe deprecated ones).

arminsabouri (changes requested):

crates/botanix-lib/src/mint_validation.rs
scottmillner: ah yep, refactored to check and print legacy version if detected
can add legacy support when needed

crates/botanix-lib/src/peg_contract.rs
```diff
@@ -239,6 +239,11 @@ impl PegoutData {
 
         Ok(Self { amount, destination: network_checked_destination, network: btc_network })
     }
+
+    // current version of the pegout which is checked during validation
```
arminsabouri:
```suggestion
    // current version of the pegout data structure
```

crates/botanix-lib/src/mint_validation.rs
```diff
@@ -231,6 +231,22 @@ pub fn try_parse_burn_event(
...
+    ))?;
+
+    if metadata.len() != 1 {
+        return Err(ParseBurnEventError::InvalidPegoutData(PegoutDataError::Invalid(
+            "invalid metadata length",
+        )));
+    }
+
+    if metadata[0] != PegoutData::version() {
+        info!("Legacy pegout submitted, version: {}", metadata[0].to_string());
```
arminsabouri:
```suggestion
        info!("unexpected pegout version submitted, version: {}", metadata[0].to_string());
```

arminsabouri (approved):
Nits only

PR #378, opened 2024-09-05
Wrap txs in envelope with non-deterministic data

scottmillner:
This pr creates a tx that includes non-deterministic data so vote extensions can be removed.  This will allow us to use `block sync` (historical syncing) since  we can't use `vote_extensions` during historical sync bc we can't reproduce the data.


crates/consensus/authority/src/comet_bft/transactions_envelope.rs
```diff
@@ -0,0 +1,275 @@
...
+    /// Invalid data format
+    InvalidFormat(#[from] bitcoin::consensus::encode::Error),
+}
+
+// Type that wraps txs with additional non-deterministic data needed for consensus
+#[derive(Debug, Clone)]
+pub(crate) struct TransactionsEnvelope {
+    pub(crate) format: u16,
+    pub(crate) non_deterministic_data: NonDeterministicData,
+    pub(crate) txs: Vec<Bytes>,
```
arminsabouri: I think the one problem with this approach is that Comet txs are always going to be size 1 per block right? With this approach developers won't be able to easily parse the txs in Comet's RPC responses.
Are there any senarios where devs would want to read the data from Comet's RPC? @PhoenixCodeCraft @jeffgeudens ?

scottmillner: not quite following

the txs size is dynamic...looking at the logic in `process_proposal` we loop through bytes to deserialize to SignedTransactions

txs will always be an arbitrary Bytes array that needs to be deserialized based on the shape we decide
it'll be the same issue when we inject a tx that contains the non-deterministic data

first tx needs to be parsed differently than the rest

maybe im misunderstanding the question?

in my mind, with the wrapper (which im not going to use), the block still has a field `txs` that is Bytes

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -423,25 +439,45 @@ where
         let block_time = request.time.expect("block time");
         let cbft_block_hash = FixedBytes::<32>::from_slice(&request.hash.to_vec().as_slice());
 
+        // extract first tx which contains non-deterministic data and validate
+        let non_deterministic_data_bytes = txs_bytes.first().expect("bytes to exist").clone();
```
arminsabouri: should we reject here? If a peer sends bytes that dont deserialize correctly?

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -423,25 +439,45 @@ where
...
+        let reader_inner: Vec<u8> =
+            vec![non_deterministic_data_bytes].into_iter().flatten().collect();
+        let mut reader = &mut io::Cursor::new(reader_inner);
+        let non_deterministic_data = NonDeterministicData::deserialize(&mut reader)
+            .expect("non deterministic data to be deserialized");
```
arminsabouri: Samething here

arminsabouri (approved):
ApproachACK. Very nice 👍 
Just had two questions

crates/consensus/authority/src/comet_bft/abci.rs

PR #379, opened 2024-09-05
Arm/feature/comet light client

arminsabouri:
Leaving this in draft. The blocker right now is FCU takes ~5s of execution time. And blocks are coming in faster than that. So the RPC nodes are always going to be being in blocks. However, syncing via pipeline works out the box quite nice


crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -299,6 +299,11 @@ where
         info!("prepare_proposal request: {:?}", request);
         let _txs_bytes = request.txs;
```
scottmillner, line 301: we still need to add a tx with non-deterministic data but my pr has that logic and needs to be added here

this is more a reminder to do this during/after rebase and merge

scottmillner:
approach ack
awesome sauce!! 🔥 🔥 🔥

PR #380, opened 2024-09-05
refactor(cbft): build block if cache miss in finalize block

scottmillner:


crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -515,21 +517,78 @@ where
...
+
+                // get txs skipping the first non-deterministic data tx
+                let txs =
+                    match transactions_signed_from_bytes(txs_bytes.clone().iter().skip(1).cloned())
+                    {
+                        Ok(txs) => txs,
+                        Err(e) => panic!("Error decoding transactions in finalize block: {:?}", e),
+                    };
+
+                let storage = self.storage.inner.blocking_read();
```
arminsabouri, line 564: Note to self. We really need to pull most of this stuff out of storage (authorities, bitcoind_factory, etc.. ) They dont really need to be protected by a rwlock.

arminsabouri (approved):

PR #384, opened 2024-09-13
unit tests for all abci methods being used

scottmillner:
- commit() needs to be an integration test bc it requires the driver to be running which requires a btc server among other things


crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -238,6 +238,15 @@ where
         payload_config
     }
 
+    pub(crate) fn non_deterministic_data_bytes(&self) -> prost::bytes::Bytes {
```
arminsabouri, line 241: `prost::bytes::Bytes` seems like a type specific to grpc and tonic related things. Can we just return &[u8] or std::Bytes ?

scottmillner:
superceded by https://github.com/botanix-labs/Macbeth/pull/398

PR #386, opened 2024-09-19
added docker-compose for cometbft

rwlockbg:


cometbft/init.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+set -e
+
+if [ ! -f "/cometbft/config/priv_validator_key.json" ] || [ ! -f "/cometbft/config/node_key.json" ]; then
+    echo "Initializing CometBFT..."
+    cometbft init -k "secp256k1" --home /cometbft
+else
+    echo "CometBFT already initialized, skipping init process..."
+    exec cometbft node \
```
arminsabouri: Should this be if else? It seems like we always want the else block executed

cometbft/Makefile
```diff
@@ -0,0 +1,10 @@
+.PHONY: build-cometbft-local start-cometbft-local stop-cometbft-local 
+
+build-cometbft-local:
+	docker-compose -f docker-compose-cometbft-local.yml up --build -d
+
+start-cometbft-local:
+	docker-compose -f docker-compose-cometbft-local.yml up -d
+
+stop-cometbft-local:
+	docker-compose  -f docker-compose-cometbft-local.yml down
```
arminsabouri: Could we add one more command for restarting the nodes?
specifically executing the `cometBFT unsafe-reset-all --home X` command

cometbft/init.sh
rwlockbg: agreed, change

cometbft/Makefile
rwlockbg: yep, added this

arminsabouri (approved):

PR #387, opened 2024-09-19
feature(cometbft-local): spin up 2 local instances of cometbft nodes

birozuru:
- Spin up two cometbft nodes for local testing on docker compose


arminsabouri:
dup of #386 ?

PR #389, opened 2024-09-19
added back authority tests, removed warnings, cleanup, retry, removed sync logic etc.

rwlockbg:


bin/reth/src/cli/mod.rs
```diff
@@ -97,6 +97,7 @@ impl Cli {
 }
 
 impl<Ext: clap::Args + fmt::Debug> Cli<Ext> {
+    /// Runs a command using a launcer with context
```
arminsabouri: Typo
```suggestion
    /// Runs a command using a launcher with context
```

Makefile
```diff
@@ -550,7 +550,8 @@ start-poa-server-1:
 	--frost.min_signers 2 \
 	--frost.max_signers 2 \
 	--p2p-secret-key "${NODE_1_DIR}/discovery-secret" \
-	--port 30303
+	--port 30303 \
+	--abci-port=36658
```
arminsabouri: Can we use 36658 for poa server 2? The convention we started to use locally is every poa node after the first will increment the first digit of the default port numbers. i.e default abci port is 26658 or poa-node-2 would use 36658 and poa-node-3:46658, poa-node-n:(n+2)6658

crates/consensus/authority/src/lib.rs
```diff
@@ -481,38 +467,36 @@ mod tests {
...
+
+        let result =
+            consensus.validate_extra_data_header(&header, &authority_signers, Some(&different_pk));
+        assert_eq!(
+            result.err().unwrap(),
+            ConsensusError::InvalidAggregatedPublicKey(
+                InvalidAggregatedPublicKeyError::InvalidAggregatedPublicKey
+            )
+        );
+    }
```
arminsabouri, line 499: There is a couple other tests we can cover in another PR:
* Where pk is nums past genesis
* Where pk is missing

arminsabouri (approved):
Couple Nits. Otherwise looks great!

crates/consensus/authority/src/lib.rs
rwlockbg: I will add those in a follow up MR!

bin/reth/src/cli/mod.rs
rwlockbg: fixed

Makefile
rwlockbg: understood. I have assigned 26658 for poa-node-1 , 36658 for poa-node2

PR #390, opened 2024-09-19
ensure all unit tests pass

scottmillner:


crates/primitives/src/botanix/utils.rs
```diff
@@ -88,7 +88,8 @@ pub fn generate_taproot_scriptpubkey(public_key: &secp256k1::PublicKey) -> Scrip
     // Note that the public key is already tweaked with the eth address and the taptree merkel root
     // so we can use the dangerous_assume_tweaked method to create the script
     // In the case of a change output being created no eth address tweak is provided
-    let xonly = bitcoin::XOnlyPublicKey::from_slice(&public_key.serialize()).unwrap();
```
arminsabouri, line 91: Just making sure. This always panicked correct?

arminsabouri (approved):
Awesome stuff

crates/primitives/src/botanix/utils.rs
scottmillner: yeah, `public_key.serialize()` returns 33 bytes and `bitcoin::XOnlyPublicKey::from_slice` expects 33 bytes
and `public_key.x_only_public_key().0.serialize()` gives us the `x_only` as 32 bytes

PR #392, opened 2024-09-20
Chore/cleanup rpc crate

arminsabouri:
* passes factories to rpc crate
* Remove all non-poa commands and subcommands


scottmillner:
some cli unit tests are failing now that weren't before

```
TRY 2 FAIL [   1.206s] reth cli::tests::parse_color_mode
TRY 3 FAIL [   0.348s] reth cli::tests::parse_env_filter_directives
TRY 2 FAIL [   1.302s] reth cli::tests::parse_logs_path
```

maybe they're not needed now?

rwlockbg (approved):
Looks fine, but in case we ever decide to rebase from upstream, we might end up with many conflicts again ....

PR #394, opened 2024-09-23
Fix/dkg/round2 shares

arminsabouri:
Its crutial in round2 that we send each participant their respective partial share. Which is our polynomial evaluated at their index. however if every partipant has all the shares this can lead to a key cancellation or rouge key attack.

Documentation is provided here: https://frost.zfnd.org/tutorial/dkg.html
Some TODOs for my self:

- [x] update frost networking protocol
- [x] more unit tests!
- [x] zero out round 1 secret after dkg is successful

### DIFF
--- a/bin/btc-server/src/dkg.rs
+++ b/bin/btc-server/src/dkg.rs
@@ -70,7 +70,9 @@ where
 
     pub(crate) async fn add_round2_dkg(
         &self,
+        // Peer's identifier
         frost_id: frost::Identifier,
+        // Peer's round 2 package
         package: frost::keys::dkg::round2::Package,
     ) -> Result<(), DKGError> {
         if self.db.get_key_package()?.is_some() {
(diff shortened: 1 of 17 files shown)


bin/btc-server/src/dkg.rs
```diff
@@ -78,36 +78,30 @@ where
...
-                            &round1_packages,
-                            &round2_packages,
-                        )?;
-
-                        self.db.set_key_package(pk_res.0.clone())?;
-                        self.db.set_pubkey_package(pk_res.1.clone())?;
-                        self.db.flush()?;
-                    }
-                }
+        // Look for our package and store it
```
scottmillner: little confused by this comment bc we do a check above to make sure we don't store our own package, right?

scottmillner (approved):

bin/btc-server/src/dkg.rs
arminsabouri: ah yes this is left over from the previous implementation

PR #398, opened 2024-09-25
Unit tests for all used abci methods

scottmillner:


scottmillner:
porting this comment from previous pr:

@0xBEEFCAF3 
`prost::bytes::Bytes` seems like a type specific to grpc and tonic related things. Can we just return &[u8] or std::Bytes ?

scottmillner:
> porting this comment from previous pr:
> 
> @0xBEEFCAF3 `prost::bytes::Bytes` seems like a type specific to grpc and tonic related things. Can we just return &[u8] or std::Bytes ?

at the moment the only use case is for grpc bytes and i wanted to encapsulate all the logic in one area (the method)
i don't see a future use case where we need to convert the ndd to another type

so would like to keep method as is and then refactor if future use cases arise

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -825,3 +826,368 @@ where
...
+        let expected_ndd = NonDeterministicData::new(
+            abci_client.bitcoin_blockhash(),
+            abci_client.aggregate_public_key(),
+        );
+        let response_ndd_bytes = response.txs.first().expect("to have tx").clone();
+        let reader_inner: Vec<u8> = vec![response_ndd_bytes].into_iter().flatten().collect();
+        let mut reader = &mut io::Cursor::new(reader_inner);
+        let response_ndd = NonDeterministicData::deserialize(&mut reader).expect("to deserialize");
+
+        // todo: deserialize tx
```
arminsabouri, line 1046: You could also check the hex strings

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -825,3 +826,368 @@ where
...
+        request.proposer_address = proposer_address;
+
+        request.time = Some(Timestamp::default());
+        request.hash = prost::bytes::Bytes::copy_from_slice(FixedBytes::<32>::random().as_slice());
+
+        let response = abci_client.process_proposal(request);
+
+        // this fails bc prevrandao isn't being set in the evm env during tests
+        // but all the custom code is executed successfully up to `build_and_execute`
+        assert_eq!(response.status, VERIFY_REJECT);
```
arminsabouri, line 1100: Should we also add a test for invalid tx? Maybe a nonce re-use?

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -825,3 +826,368 @@ where
...
+        let proposer_address = prost::bytes::Bytes::copy_from_slice(Address::ZERO.0.as_slice());
+        request.proposer_address = proposer_address;
+
+        request.time = Some(Timestamp::default());
+        request.hash = prost::bytes::Bytes::copy_from_slice(FixedBytes::<32>::random().as_slice());
+
+        // need to call finalize block first to generate a block in the cache to commit
+        let _finalize_block_response = abci_client.finalize_block(request);
+
+        // will panic bc the driver isn't running
```
arminsabouri, line 1190: Lets add a todo to test the driver

arminsabouri (approved):

rwlockbg (approved):

PR #403, opened 2024-09-26
feature(docker-local): local testing setup 2/2

birozuru:


docker-local/docker-compose.yml
```diff
@@ -0,0 +1,218 @@
...
+    depends_on:
+      - signing-server-1
+      - consensus-node-1
+    volumes:
+      - ./poa-1:/reth/botanix_testnet:rw
+      - ./signing-server-1/jwt.hex:/reth/botanix_testnet/jwt.hex
+    restart: on-failure
+
+
+  consensus-node-1:
```
rwlockbg: I would prefer adding cometbft in the naming somehow to make it more clear

birozuru: Okay will do

rwlockbg (approved):

PR #405, opened 2024-09-27
Setup RPC node without light client

scottmillner:
I've tested this and rpc node is performing as expected and rpc endpoints are functioning as expected

### DIFF
--- a/bin/reth/src/commands/poa/mod.rs
+++ b/bin/reth/src/commands/poa/mod.rs
@@ -841,14 +841,6 @@ impl<Ext: clap::Args + fmt::Debug> PoaNodeCommand<Ext> {
             }
         };
         if !is_fed_node {
-            // TODO: remove once non-validator node is setup
-            // info!(target: "reth::cli", "Starting PoA Block Fetcher Task");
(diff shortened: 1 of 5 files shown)


rwlockbg (approved):

bin/reth/src/commands/poa/mod.rs
```diff
@@ -853,13 +841,6 @@ impl<Ext: clap::Args + fmt::Debug> PoaNodeCommand<Ext> {
             }
         };
         if !is_fed_node {
-            info!(target: "reth::cli", "Starting PoA Block Fetcher Task");
-            executor.spawn_critical(
```
arminsabouri, line 857: Can we also remove block fetcher as a exported member of authority-consensus?

arminsabouri: Then we can also remove it from the authority-builder as well

arminsabouri (approved):
ApproachACK

bin/reth/src/commands/poa/mod.rs
scottmillner: done and done!

PR #410, opened 2024-09-30
Feature/btc server/pending pegout endpoint

arminsabouri:
https://github.com/botanix-labs/botanix/issues/664
https://github.com/botanix-labs/botanix/issues/578


bin/btc-server/src/server.rs
```diff
@@ -221,6 +221,27 @@ where
         Ok(tonic::Response::new(rpc::Empty {}))
     }
 
+    /// Admin util endpoing to get all pending pegouts
```
scottmillner, line 224: nit: `endpoing > endpoint`

scottmillner (approved):

bin/btc-server/src/server.rs
arminsabouri: interesting codespell never didn't pick up on this

PR #413, opened 2024-10-01
Fix/dkg p2p messages

arminsabouri:
Fix is broken into 3 seperate commits. Each can be reviewed seperately.
The main issue here is the coordinator only in certain conditions would send its round2 commitments to its peers. That fix is in the 3rd commit. 

Tested with a local 2-2 set up


rwlockbg (dismissed):

birozuru:
LGTM!

birozuru (dismissed):

PR #414, opened 2024-10-01
Notify pegouts in abci driver

scottmillner:
leaving a note here for future reference:

When testing locally, the pegin ultimately fails bc adding the tx to the mempool fails on line:

519 @ `crates/transaction-pool/src/pool/mod.rs` in `add_transactions` @

```
let mut added =
            transactions.into_iter().map(|tx| self.add_transaction(origin, tx)).collect::<Vec<_>>();
```

`self.add_transaction()` returns `Ok(txHash)` but then `collect()` is failing

doesn't make sense bc a successful `txHash` is returned so not gonna dive any deeper for now

want to see if this occurs when deployed to testnet


crates/consensus/authority/src/utils.rs
```diff
@@ -123,6 +125,31 @@ pub(crate) async fn call_notify_pegin(
...
+pub(crate) async fn call_notify_pegout(
+    btc_server: &mut BtcServerExtendedClient,
+    pegouts: &[PegoutWithId],
+    height: u64,
+) -> Result<(), GrpcClientError> {
+    if pegouts.is_empty() {
+        return Ok(());
+    }
+
+    // TODO: do we want to modify NotifyPegoutRequest to take a list of pegouts?
```
scottmillner, line 137: @0xBEEFCAF3 ^^

rwlockbg (approved):

crates/consensus/authority/src/utils.rs
arminsabouri: Ah yes that would have made sense

arminsabouri (approved):

PR #416, opened 2024-10-02
fix(abci): prepare_proposal clears mempool of txs

scottmillner:
still need to test


scottmillner:
closing bc confirmed mempool is cleared once block is committed to chain

the issue i was seeing was due to another issue...empty blocks we're being produced and the tx in the mempool wasn't being included

PR #418, opened 2024-10-03
Get frost coordinator from epoch block hash

scottmillner:


crates/consensus/authority/src/frost_task.rs
```diff
@@ -50,6 +50,8 @@ pub(crate) struct FrostNotification {
     pub(crate) signing_session_id: Vec<u8>,
     /// The agglomerated psbts
     pub(crate) psbt: Vec<u8>,
+    /// Optional epoch block hash used to determine the coordinator
+    pub(crate) epoch_block_hash: Option<Vec<u8>>,
```
arminsabouri, line 54: Why make this optional?

crates/consensus/authority/src/frost_task.rs
```diff
@@ -226,6 +228,7 @@ where
                         .initate_signing_session(
                             frost_notification.signing_session_id,
                             frost_notification.psbt,
+                            frost_notification.epoch_block_hash.expect("epoch block hash to exist"),
```
arminsabouri, line 231: It seems like we are using epoch block hash as an idenfitier for the signing session? Maybe signing session id and epoch block hash are one? 

I understand why we would have to pass this down everywhere. i.e if you recieve round 1/2 messages you need to know if you are infact the coordinator and singers would need to also know who to send messages to. 

Maybe it makes more sense to create a validate_ssid(block_hash) where we take a block hash and check that it is infact the last epoch block that was produced

crates/consensus/common/src/utils.rs
```diff
@@ -46,13 +46,24 @@ pub fn validate_chain_version(edh_chain_version: u32) -> Result<(), ConsensusErr
...
-pub fn is_inturn(authorities_len: u64, signer_index: u64, block_time: u64) -> bool {
-    let timestamp = unix_timestamp(); // Keep the timestamp in seconds
+/// Uses the epoch block hash to determine inturness
+pub fn is_inturn(
+    authorities_len: u64,
+    signer_index: u64,
+    block_time: u64,
+    epoch_block_hash: Vec<u8>,
+) -> bool {
+    let hash_u64 = block_hash_to_u64(epoch_block_hash); // Convert the first 8 bytes of the hash into a u64
```
arminsabouri, line 62: this hash should just be a u256 no? Can we just use 256 arithmitic?

crates/consensus/common/src/utils.rs
```diff
@@ -93,17 +104,20 @@ pub fn get_in_turn_interval(
     )
 }
 
-/// Returns the index of the authority which is currently in turn based on the seconds passed
+/// Returns the index of the authority which is currently in turn based on the epoch block hash
 pub fn current_inturn_index(
     authorities_len: u64,
-    reference_timestamp: u64,
+    epoch_block_hash: Vec<u8>,
```
arminsabouri, line 110: Nit: use fixedBytes<32>

crates/net/network/src/frost/messages.rs
```diff
@@ -396,6 +409,9 @@ impl FrostProtoMessage {
                 // psbt
                 buf.put_u32_le(resource.psbt.len() as u32); // Use u32 to support larger data sizes
                 buf.put_slice(&resource.psbt);
+                // epoch block hash
+                buf.put_u32_le(resource.epoch_block_hash.len() as u32); // Use u32 to support larger data sizes
```
arminsabouri, line 413: I dont think we need to pass this epoch block hash with each frost message. Esp since epoch block hash and ssid seem to serve the same purpose. Ideally what happens here is a signer / coord gets a psbt with a ssid. They pull the latest epoch block from db and compare. Then validate that the inturn person is actually inturn

crates/consensus/authority/src/frost_task.rs
scottmillner: i wasn't sure if all the frost notification needed it but gonna refactor so it's not optional

arminsabouri: Also I would go with FixedBytes primitive here

crates/consensus/authority/src/frost_task.rs
scottmillner: need to sync up on this

the reason i had to pass it around everywhere is that several methods use `get_coordinator` or `current_inturn_index` for example all which need access to the block hash

so if we don't want to include it on all the frost messages, we need to make `epoch_block_hash` part of `SigningStateMachine` so those methods have access to it or come up with a different way

crates/net/network/src/frost/messages.rs
scottmillner: ~~see note above above needing access to the hash~~

~~the methods that need access to the `epoch_block_hash` are on `SigningStateMachine` but that struct doesn't have access to the db to make the comparison (and we still need to add `epoch_block_hash` to that struct first)~~

~~how about we sync up on this and whiteboard it out a bit...im gonna need to start a fresh pr most likely~~

we synced up and gonna pass a struct that impls a trait which accesses source of 32 byte randomness
and figure out later what that source is

scottmillner:
closing and will be superseded by a new pr

PR #419, opened 2024-10-04
Get frost coordinator from random source

scottmillner:
Adds a random source provider that is passed to `SigningStateMachine` and used to determine authority in turn

### DIFF
--- a/crates/consensus/authority/src/random_source_provider.rs
+++ b/crates/consensus/authority/src/random_source_provider.rs
@@ -1,0 +1,17 @@
+use reth_revm::primitives::FixedBytes;
+
+/// Trait that provides a random source of 32 bytes.
+pub trait RandomSource {
+    /// Returns a random source.
+    fn random_source(&self) -> FixedBytes<32>;
+}
+
+/// Struct that provides a random source of 32 bytes.
+pub struct RandomSourceProvider;
+
+impl RandomSource for RandomSourceProvider {
+    // TODO: Implement a proper random source.
+    fn random_source(&self) -> FixedBytes<32> {
+        FixedBytes::default()
+    }
+}
--- a/crates/consensus/authority/src/random_source_provider.rs
+++ b/crates/consensus/authority/src/random_source_provider.rs
@@ -1,13 +1,24 @@
+//! Module for providing random sources.
+use std::fmt::Debug;
+
 use reth_revm::primitives::FixedBytes;
 
 /// Trait that provides a random source of 32 bytes.
-pub trait RandomSource {
+pub trait RandomSource: Debug {
     /// Returns a random source.
     fn random_source(&self) -> FixedBytes<32>;
 }
 
 /// Struct that provides a random source of 32 bytes.
+#[derive(Debug)]
 pub struct RandomSourceProvider;
+
+impl RandomSourceProvider {
+    /// Creates a new `RandomSourceProvider`.
+    pub fn new() -> Self {
+        Self
+    }
+}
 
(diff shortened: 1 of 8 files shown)


crates/consensus/authority/src/random_source_provider.rs
```diff
@@ -0,0 +1,28 @@
...
+
+use reth_revm::primitives::FixedBytes;
+
+/// Trait that provides a random source of 32 bytes.
+pub trait RandomSource: Debug {
+    /// Returns a random source.
+    fn random_source(&self) -> FixedBytes<32>;
+}
+
+/// Struct that provides a random source of 32 bytes.
```
scottmillner, line 12: assuming 32 bytes which has some logic downstream that expects that but will refactor as needed

crates/consensus/common/src/utils.rs
```diff
@@ -46,51 +49,34 @@ pub fn validate_chain_version(edh_chain_version: u32) -> Result<(), ConsensusErr
...
     signer_index: u64,
-    reference_timestamp: u64,
     block_time: u64,
-) -> CoordinatorInterval {
-    assert!(block_time > 0, "block_time must be greater than 0");
-    // Calculate the length of one complete cycle
-    let cycle_length = authorities_len * block_time;
+    random_source: FixedBytes<32>,
+) -> bool {
+    // convert types to U256 since random_source is 32 bytes and do arithmetic
```
scottmillner, line 66: assuming 32 bytes but will refactor as needed

crates/consensus/authority/src/random_source_provider.rs
arminsabouri: Make generic?

scottmillner: via discussion, will assume 32 bytes for now

crates/consensus/authority/src/random_source_provider.rs
```diff
@@ -0,0 +1,28 @@
...
+    /// Creates a new `RandomSourceProvider`.
+    pub fn new() -> Self {
+        Self
+    }
+}
+
+impl RandomSource for RandomSourceProvider {
+    // TODO: Implement a proper random source.
+    fn random_source(&self) -> FixedBytes<32> {
+        FixedBytes::default()
```
arminsabouri, line 26: what is the default implementation? All 0 bytes?

crates/consensus/common/src/utils.rs
```diff
@@ -46,51 +49,34 @@ pub fn validate_chain_version(edh_chain_version: u32) -> Result<(), ConsensusErr
...
-pub type CoordinatorInterval = (u64, u64, u64, u64);
-
-/// Returns the inturn interval for a signer index based on the seconds passed
-pub fn get_in_turn_interval(
+/// Returns true if the authority is in turn
+pub fn is_inturn(
     authorities_len: u64,
     signer_index: u64,
-    reference_timestamp: u64,
     block_time: u64,
```
arminsabouri: Would rename to something more generic like period or time_range

arminsabouri (dismissed):

crates/consensus/authority/src/random_source_provider.rs
scottmillner: yeah `default()` just returns a `const` that's all 0 bytes

crates/consensus/common/src/utils.rs
scottmillner: went with `time_range`

arminsabouri (approved):

PR #420, opened 2024-10-07
Frost task should listen to canon events and kick of signing state machine

scottmillner:


crates/consensus/authority/src/frost_task.rs
```diff
@@ -248,6 +256,69 @@ where
...
+                            if coordinator.is_some() {
+                                info!("Received canon state notification but we're not the coordinator");
+                                continue;
+                            } else {
+                                // create psbt and send init signing message
+                                let bitcoin_checkpoint = tip
+                                    .header
+                                    .header()
+                                    .deserialize_extra_data_header()
+                                    .expect("to deserialize extra data header")
```
arminsabouri: Do we want to log an error and continue here instead? Not sure when a invalid block would be made canon but perhaphs the frost task can continue

crates/consensus/authority/src/frost_task.rs
```diff
@@ -248,6 +256,69 @@ where
...
+
+            // Receive canon state notifications
+            while let Ok(notification) = self.canon_state_notification_receiver.try_recv() {
+                match notification {
+                    CanonStateNotification::Commit { new } => {
+                        // check if epoch block and if we are the coordinator
+                        // if so, send init signing message
+                        let tip = new.tip();
+                        if tip.is_poa_epoch() {
+                            // the coordinator is always expected to be us in this case, i.e.
```
arminsabouri: I might be misunderstanding the comment. 
Wouldnt this always be some? Or will get_coordinator only return Some when the caller is the coord? In which case it could probably use a better name

arminsabouri (dismissed):
Approach/Code ACK
Just had a question

crates/consensus/authority/src/frost_task.rs
scottmillner: this comment was copied from the previous code

`get_coordinator` returns None if the coordinator and Some if it's not us

snippet from the method:

```
/// Gets the current federation coordinator. Returns None if it is us, otherwise Some if someone
/// Uses a random 32 byte source to determine the current inturn authority
pub(crate) async fn get_coordinator(&self) -> Result<Option<(PeerData, u64)>, Error> {
```    

agree it's confusing and i'll update the comment to be more clear

arminsabouri: Yeah I agree this is confusing. A better util would be `is_coordinator` just returns true if this peer is the coordinator Or  get_coordinator doesnt return a Option instead the caller check if the self.peerid == returned_data.peer_id

crates/consensus/authority/src/frost_task.rs
scottmillner: ah yep, that's better
done!

arminsabouri:
@scottmillner I think we got a type error in the int. test suite 
Should be a quick fix

crates/consensus/authority/src/frost_task.rs
scottmillner: `is_coordinator` already exists so just gonna use that instead, not sure why i didn't think of that before lol
done

scottmillner:
> @scottmillner I think we got a type error in the int. test suite Should be a quick fix

my bad, fixed

PR #421, opened 2024-10-08
Test/additional/signing assertions

arminsabouri:


bin/test-suite/src/suite/consensus/frost/test_signing.rs
```diff
@@ -158,6 +158,24 @@ pub async fn do_signing(
     Ok(final_tx)
 }
 
+/// Assert that all clients have the same UTXO set
```
scottmillner, line 164: i like this test
concise yet clever lol, nice use of HashSet

scottmillner (dismissed):

scottmillner (approved):

PR #423, opened 2024-10-08
Validate pegout requests using pegout ids

scottmillner:
for now just doing an exact match on pegout amount which likely be refactored in the future to account for fees

### DIFF
--- a/crates/consensus/authority/src/frost_task.rs
+++ b/crates/consensus/authority/src/frost_task.rs
@@ -247,16 +247,25 @@ where
                 })?;
 
             // check if a corresponding output exists in the psbt
-            if let Ok(transaction) = psbt.clone().extract_tx() {
-                if let None = transaction.output.iter().find(|output| {
-                    output.script_pubkey == destination.script_pubkey() && output.value == amount
-                }) {
-                    error!(target: "consensus::authority::frost_task::validate_psbt_by_ids", "Failed to find matching output in psbt");
+            match psbt.clone().extract_tx() {
+                Ok(transaction) => {
+                    match transaction.output.iter().find(|output| {
+                        output.script_pubkey == destination.script_pubkey() &&
+                            output.value == amount
+                    }) {
+                        Some(_) => {
+                            info!(target: "consensus::authority::frost_task::validate_psbt_by_ids", "Found matching output in psbt");
+                        }
+                        None => {
+                            error!(target: "consensus::authority::frost_task::validate_psbt_by_ids", "Failed to find matching output in psbt");
+                            return Err(PsbtValidationError::FailedToValidatePsbtByIds);
+                        }
+                    }
+                }
+                Err(e) => {
+                    error!(target: "consensus::authority::frost_task::validate_psbt_by_ids", "Failed to extract transaction from psbt {:?}", e);
                     return Err(PsbtValidationError::FailedToValidatePsbtByIds);
-                };
-            } else {
-                error!(target: "consensus::authority::frost_task::validate_psbt_by_ids", "Failed to extract transaction from psbt");
-                return Err(PsbtValidationError::FailedToValidatePsbtByIds);
+                }
             }
         }
 
--- a/crates/consensus/authority/src/frost_task.rs
+++ b/crates/consensus/authority/src/frost_task.rs
@@ -24,7 +24,10 @@ use reth_network::{
     },
     NetworkHandle,
 };
-use reth_primitives::{botanix::mint_validation::try_parse_burn_event, header_ext::HeaderExt};
+use reth_primitives::{
+    botanix::{mint_validation::try_parse_burn_event, peg_contract::PegoutData},
+    header_ext::HeaderExt,
+};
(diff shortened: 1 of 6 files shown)


crates/consensus/authority/src/frost_task.rs
```diff
@@ -184,6 +206,58 @@ where
...
+                .map_err(|e| {
+                    error!(target: "consensus::authority::frost_task::validate_psbt_by_ids", "Failed to parse burn event {:?}", e);
+                    PsbtValidationError::FailedToValidatePsbtByIds
+                })?
+                .ok_or_else(|| {
+                    error!(target: "consensus::authority::frost_task::validate_psbt_by_ids", "Failed to get pegout data from burn event");
+                    PsbtValidationError::FailedToValidatePsbtByIds
+                })?;
+
+            // check if a corresponding output exists in the psbt
```
scottmillner: @0xBEEFCAF3 is this check sufficient?

im just looping through and finding an output that matches the pegout data
i don't have any way of knowing which output the pegout is, correct?

as in, we don't have a mapping of output index to pegout id, correct?

crates/consensus/authority/src/frost_task.rs
```diff
@@ -184,6 +206,58 @@ where
...
+    pub(crate) async fn validate_psbt_by_ids(&self, psbt: Psbt) -> Result<(), PsbtValidationError> {
+        let pegout_ids = psbt
+            .outputs
+            .iter()
+            .filter_map(|output| match output.pegout_id() {
+                Some(pegout_id) => PegoutId::from_bytes(pegout_id.as_slice()).ok(),
+                _ => None,
+            })
+            .collect::<Vec<_>>();
+
```
arminsabouri: if len of pegout_ids is 0 something is off. Lets Error out here

crates/consensus/authority/src/frost_task.rs
arminsabouri: Yes this looks correct to me

crates/consensus/authority/src/frost_task.rs
```diff
@@ -184,6 +206,58 @@ where
...
+                    error!(target: "consensus::authority::frost_task::validate_psbt_by_ids", "Failed to parse burn event {:?}", e);
+                    PsbtValidationError::FailedToValidatePsbtByIds
+                })?
+                .ok_or_else(|| {
+                    error!(target: "consensus::authority::frost_task::validate_psbt_by_ids", "Failed to get pegout data from burn event");
+                    PsbtValidationError::FailedToValidatePsbtByIds
+                })?;
+
+            // check if a corresponding output exists in the psbt
+            if let Ok(transaction) = psbt.clone().extract_tx() {
```
arminsabouri: Nit: match here do remain consistent with the rest of this file's convention

crates/consensus/authority/src/frost_task.rs
scottmillner: done!

crates/consensus/authority/src/frost_task.rs
scottmillner: done!

crates/consensus/authority/src/frost_task.rs
```diff
@@ -57,6 +68,12 @@ pub(crate) struct FrostNotification {
     pub(crate) psbt: Vec<u8>,
 }
 
+#[derive(Debug, thiserror::Error)]
+pub(crate) enum PsbtValidationError {
+    #[error("Failed to validate psbt by ids")]
+    FailedToValidatePsbtByIds,
```
arminsabouri: Maybe we add a string here `(FailedToValidatePsbtByIds(String))` so we can seperate out which cases we are error'ing out

crates/consensus/authority/src/frost_task.rs
```diff
@@ -184,6 +206,72 @@ where
...
+                    error!(target: "consensus::authority::frost_task::validate_psbt_by_ids", "Failed to get pegout data from burn event");
+                    PsbtValidationError::FailedToValidatePsbtByIds
+                })?;
+
+            // check if a corresponding output exists in the psbt
+            match psbt.clone().extract_tx() {
+                Ok(transaction) => {
+                    match transaction.output.iter().find(|output| {
+                        output.script_pubkey == destination.script_pubkey() &&
+                            output.value == amount
```
arminsabouri: Can you please leave a TODO here that strict amount checking will need to account for the fees coming out of the pegout outputs in the future

arminsabouri:
@scottmillner how hard would it be to mock the db here and unit test this? This should definetly recieve some test coverage

crates/consensus/authority/src/frost_task.rs
scottmillner: done!

crates/consensus/authority/src/frost_task.rs
scottmillner: done!

arminsabouri (approved):

PR #431, opened 2024-10-11
Add pegout tests

scottmillner:


crates/primitives/src/botanix/mint_validation.rs
```diff
@@ -346,6 +346,85 @@ mod test {
         );
     }
 
-    // TODO: add decode_log_payload for pegout
-    // integration tests do cover this but unit test is nice to have
+    #[test]
+    fn decode_burn_log_payload() {
```
scottmillner, line 350: not an essential test since test below covers that same logic
but does show how we parse logs so added it

arminsabouri (approved):
nice!!!

PR #446, opened 2024-10-16
chore:Added local_setup.sh to automates and spin the nodes

nagarajm87:
* Added `local_script.sh` to spin all the nodes.
* Updated `Readme` for running the script.


binaries/bitcoin-tx
arminsabouri, line 1: Do we need these are bitcoin binaries. Seems like we really just need bitcoind

local_setup.sh
```diff
@@ -0,0 +1,145 @@
...
+    echo "Starting bitcoind with configuration from $BITCOIN_NODE_DIR/bitcoin.conf..."
+   ./binaries/bitcoind -datadir="$BITCOIN_NODE_DIR" -conf="$BITCOIN_NODE_DIR/bitcoin.conf" &
+
+    # Wait for bitcoind to start
+    sleep 5
+    echo "Creating wallet 'mywallet' in $BITCOIN_NODE_DIR..."
+    ./binaries/bitcoin-wallet -chain=regtest -wallet=mywallet -datadir="$BITCOIN_NODE_DIR" create
+
+    echo "Loading wallet 'mywallet'..."
+    ./binaries/bitcoin-cli -rpcport=18443 -rpcuser=test123 -rpcpassword=test123 loadwallet "mywallet"
```
arminsabouri: Could probably use the env vars  
-rpcuser=$RPC_USER -rpcpassword=$RPC_PASS

local_setup.sh
```diff
@@ -0,0 +1,145 @@
...
+
+    echo "Generating 200 blocks..."
+    ./binaries/bitcoin-cli -rpcport=18443 -rpcuser=test123 -rpcpassword=test123 -generate 200
+    
+    sleep 5
+    echo "Running 'make make start-btc-server_1'..."
+    make start-btc-server-1 &      
+    sleep 60
```
arminsabouri: Do we need to sleep for a minute here? esp since the last command is running the background

local_setup.sh
```diff
@@ -0,0 +1,145 @@
...
+    echo "Running 'make make start-btc-server_2'..."
+    make start-btc-server-2 &  
+
+    sleep 10
+    echo "Running 'make start-poa-server_1' ..."
+    make make start-poa-server-1  &
+    sleep 60
```
arminsabouri: Same thing here

local_setup.sh
```diff
@@ -0,0 +1,145 @@
...
+    sleep 5
+    echo "Starting CometBFT node1..."
+    ./binaries/cometbft start --home "$COMET_NODE_DIR/node1" &
+    echo "CometBFT node started with home directory $HOME_DIR/".
+}
+
+# Function to clean up (delete directories and .env file)
+stop() {
+    echo "Cleaning up..."
+    ./binaries/bitcoin-cli -regtest -rpcport=18443 -rpcuser=test123 -rpcpassword=test123 stop
```
arminsabouri: Could also send sigint and not keeo a bitcoin-cli binary.

local_setup.sh
```diff
@@ -0,0 +1,145 @@
...
+     
+    pkill -9 -f btc-server
+    pkill -9 -f cometbft
+
```
arminsabouri, line 177: Do we also need to kill the poa nodes?

local_setup.sh
```diff
@@ -0,0 +1,145 @@
+#!/bin/bash
+
+
+HOME_DIR="$HOME/develop"
```
arminsabouri: Maybe we can create the temp dirs in /tmp or somehere in the cwd (and the gitignore it)  ?

arminsabouri (changes requested):
Im glad we finally got around to doing something like this. This is super impressive. I had a couple questions and some requests for changes

binaries/bitcoin-tx
nagarajm87: I kept the other bitcoin binaries to create wallet and generate some blocks.

local_setup.sh
nagarajm87: will update this :)

local_setup.sh
nagarajm87: I added a one-minute delay, so when running make start-btc-server-1, it starts with the compilation process, then proceeds to the launch phase.

local_setup.sh
nagarajm87: will update it :)

binaries/bitcoin-tx
rwlockbg: Naga, lets not include binaries here - this will blow up the pull duration, plus it sonly for ubuntu / macos. If we want to have the binaries, we need to publish them for all platforms. If you however want them , there is a lfg module for gitub for handling large files.

local_setup.sh
rwlockbg: agree, let that be a new  directory under root

rwlockbg: a temp one

local_setup.sh
rwlockbg: the btc server needs at most 10 secs or so to boot up, its pretty fast.

local_setup.sh
```diff
@@ -0,0 +1,145 @@
...
+}
+
+# Function to start the environment and run the CometBFT testnet and Bitcoin commands
+start() {
+    # Call the 'init' function to ensure everything is set up before starting
+    echo "Starting the bitcoin node..."
+    sleep 5
+    # Run Bitcoin-related commands
+    echo "Starting bitcoind with configuration from $BITCOIN_NODE_DIR/bitcoin.conf..."
+   ./binaries/bitcoind -datadir="$BITCOIN_NODE_DIR" -conf="$BITCOIN_NODE_DIR/bitcoin.conf" &
```
arminsabouri: Could also consider passing the paths to the bitcoind and comet binaries.

binaries/bitcoin-tx
nagarajm87: I have removed the binaries, user need to provide the path

local_setup.sh
nagarajm87: I have updated, sleep :)

local_setup.sh
nagarajm87: updated created temp folder :)

local_setup.sh
nagarajm87: done :)

local_setup.sh
nagarajm87: resolved :)

rwlockbg (dismissed):
Looks good ! In the feature it would be nice to add some extra scripting for pre-configuring the nodes connectivity in both genesis and config.tomls for each cometbft as this would help the dev enormously.

arminsabouri:
@nagarajm87 can you rebase this on top of the main branch?

local_setup.sh
```diff
@@ -0,0 +1,196 @@
...
+init() {
+      DEFAULT_BINARY_PATH="{PROVIDE_BINARIES_PATH}"
+      BINARY_PATH=${1:-$DEFAULT_BINARY_PATH}
+     echo "Using binary path: $BINARY_PATH"
+    if [[ ! -d "$BINARY_PATH" ]]; then
+        echo "Error: Binary path '$BINARY_PATH' does not exist."
+        exit 1
+    fi
+
+    Check if the directories already exist
```
arminsabouri: Should this be comment?

arminsabouri (dismissed):
UTack! Just one question. Otherwise good to go

scottmillner (dismissed):

scottmillner:
looks good! id say get the `lint/crate-checks` job passing then should be good to merge!

local_setup.sh
nagarajm87: will update it :)

scottmillner:
for the last commit `rebased latest`:
- when rebasing, just rebase and force push so no new commit is added
- i see there's a small change in there (a comment): if you still need that change just `git reset --soft HEAD~1`, add that change to the previous changes and force push

rwlockbg (approved):

PR #447, opened 2024-10-16
Feature/frost should abort previous signing session one a new round

scottmillner:


rwlockbg (dismissed):

crates/consensus/authority/src/signing.rs
```diff
@@ -739,35 +740,31 @@ where
         };
         info!(target: "consensus::authority::signing::signer_process_round1", "coordinator index {:?}", coordinator_id);
 
+        // check coordinator is sending the request
+        let coordinator_frost_identifier = coordinator_peer_data
+            .frost_identifier
+            .expect("Coordinator should have a frost identifier");
+        if coordinator_frost_identifier.serialize().to_vec() != identifier {
+            error!(target: "consensus::authority::signing::signer_process_round1", "Round 1 signing request not from coordinator");
```
arminsabouri: Lets warn! here

crates/consensus/authority/src/signing.rs
```diff
@@ -739,35 +740,31 @@ where
         };
         info!(target: "consensus::authority::signing::signer_process_round1", "coordinator index {:?}", coordinator_id);
 
+        // check coordinator is sending the request
+        let coordinator_frost_identifier = coordinator_peer_data
+            .frost_identifier
+            .expect("Coordinator should have a frost identifier");
+        if coordinator_frost_identifier.serialize().to_vec() != identifier {
```
arminsabouri, line 747: I think we want to also do this check on round 2 if we are not already

arminsabouri (dismissed):

rwlockbg (dismissed):

crates/consensus/authority/src/signing.rs
scottmillner: done!

crates/consensus/authority/src/signing.rs
scottmillner: done!

crates/consensus/authority/src/signing.rs
```diff
@@ -656,24 +661,24 @@ where
...
-                // session was previously already registered, maybe it got retriggered
-                // check if it was the same coordinator
-                // check if it is still valid time-wise
-                if signing_session.coordinator_index != self.frost_config.authority_index as u64 {
-                    // session is no longer valid, remove it from cache and return
+                // a coordinator should never re-trigger an existing session
+                if signing_session.coordinator_index == self.frost_config.authority_index as u64 {
+                    // clear session and lose ability to be coordinator
+                    // this could happen if previous session failed but wasn't removed
+                    self.abort_signing().await?;
```
arminsabouri, line 672: I cant remember where we left of on this. Did we decide to keep this? 
I remeber this was an issue where a coord could be mid-signing and have their signing session cleared by an accidental init signing req

arminsabouri (approved):
ApproachACK. Just one question

crates/consensus/authority/src/signing.rs
scottmillner: yeah that could happen but we decided that somethign is seriously wrong if it does
bc:
- only during an epoch block and when you're in-turn, does the node init a new signing session
- so the code should prevent a coordinator node from re-init an existing signing session
- if it does try abort and error

i do think once we start testing, maybe the behavior isn't exactly what i think but this makes sense to me at the moment

PR #450, opened 2024-10-17
Validate additional outputs are change

scottmillner:


bin/btc-server/src/signer.rs
```diff
@@ -402,4 +377,49 @@ where
...
+        let mut psbt_pegout_ids: Vec<PegoutId> = vec![];
+        for output in psbt.outputs.iter() {
+            match output.pegout_id() {
+                Some(id) => match PegoutId::from_bytes(&id) {
+                    Ok(id) => psbt_pegout_ids.push(id),
+                    Err(_) => return Err(SigningFinalizeError::MissingPsbtPegout),
+                },
+                // check extra outputs are change outputs: redeem script should be change script
+                // pubkey derived from aggregated public key
+                None => {
```
scottmillner: @0xBEEFCAF3 
not sure if this logic is correct?

bin/btc-server/src/signer.rs
```diff
@@ -402,4 +377,49 @@ where
...
     }
+
+    pub(crate) fn pending_pegout_ids(pegouts: Vec<PegoutRequest>) -> Vec<PegoutId> {
+        pegouts.iter().map(|p| p.id).collect::<Vec<PegoutId>>()
+    }
+
+    // Check all pending pegouts are being settled in this tx
+    // and additional outputs are change outputs
+    pub(crate) fn validate_outputs(&self, psbt: &Psbt) -> Result<(), SigningFinalizeError> {
+        let pending_pegouts = self.db.get_pending_pegouts()?;
```
arminsabouri: We could check here if `self.db.get_public_key_package()?` exists. If it does not throw error

bin/btc-server/src/signer.rs
arminsabouri: Looks good to me!

bin/btc-server/src/signer.rs
```diff
@@ -402,4 +377,49 @@ where
...
+                None => {
+                    let agg_pk = self
+                        .db
+                        .get_public_key_package()?
+                        .expect("valid public key package")
+                        .verifying_key()
+                        .to_secp_pk()
+                        .expect("valid secp pk");
+                    let expected_script_pubkey = generate_taproot_change_scriptpubkey(&agg_pk);
+                    let script_pubkey = output.redeem_script.as_ref().expect("valid redeem script");
```
arminsabouri: not sure if redeem script is what we are looking for. There should a script pubkey field 
https://docs.rs/bitcoin/latest/bitcoin/struct.TxOut.html

bin/btc-server/src/signer.rs
```diff
@@ -402,4 +377,49 @@ where
...
+                    if script_pubkey != &expected_script_pubkey {
+                        return Err(SigningFinalizeError::PsbtIncludesExtraOutput);
```
arminsabouri: how do you feel about a invalid change out error variant

bin/btc-server/src/signer.rs
arminsabouri: just found thing

arminsabouri (commented):
Good stuff!!

bin/btc-server/src/signer.rs
scottmillner: done!

bin/btc-server/src/signer.rs
scottmillner: done!

bin/btc-server/src/signer.rs
scottmillner: done!

bin/btc-server/src/signer.rs
```diff
@@ -402,4 +376,55 @@ where
...
+                None => additional_outputs.push(output.clone()),
+            };
+        }
+
+        for pegout_id in pending_pegout_ids.iter() {
+            if !psbt_pegout_ids.contains(&pegout_id) {
+                return Err(SigningFinalizeError::MissingPendingPegout(pegout_id.clone()));
+            }
+        }
+
```
scottmillner: Originally i wanted to extract the `TxOut` corresponding to the psbt `Output` but you can't so here's my logic:
- ive looped through all the pegouts and collected any that don't have a `PegoutId`
- there should only be one of the above and it should be a change output
- then loop through extracted_tx outputs and find matching scriptpubkey

bin/test-suite/src/suite/consensus/frost/test_signing.rs
```diff
@@ -170,11 +170,11 @@ pub async fn all_clients_have_same_utxo_set(
     let mut utxo_merkle_root = HashSet::new();
     for c in clients.iter_mut() {
         let root = c
```
scottmillner: this was breaking so updated it
@0xBEEFCAF3 double check this is what you want, thanks!

bin/btc-server/src/signer.rs
```diff
@@ -402,4 +376,55 @@ where
...
+                None => return Err(SigningFinalizeError::MissingKeyPackage),
+            },
+            Err(_e) => return Err(SigningFinalizeError::MissingKeyPackage),
+        };
+
+        let pending_pegouts = self.db.get_pending_pegouts()?;
+        let pending_pegout_ids = pending_pegouts.iter().map(|p| p.id).collect::<Vec<PegoutId>>();
+
+        let mut psbt_pegout_ids: Vec<PegoutId> = vec![];
+        let mut additional_outputs: Vec<Output> = vec![];
```
arminsabouri: Nit: change outputs

```suggestion
        let mut change_outputs: Vec<Output> = vec![];
```

bin/btc-server/src/signer.rs
```diff
@@ -402,4 +376,55 @@ where
...
+                Some(pk) => pk,
+                None => return Err(SigningFinalizeError::MissingKeyPackage),
+            },
+            Err(_e) => return Err(SigningFinalizeError::MissingKeyPackage),
+        };
+
+        let pending_pegouts = self.db.get_pending_pegouts()?;
+        let pending_pegout_ids = pending_pegouts.iter().map(|p| p.id).collect::<Vec<PegoutId>>();
+
+        let mut psbt_pegout_ids: Vec<PegoutId> = vec![];
```
arminsabouri: Nit: we could predefine the capacity:
Vec::with_capacity(outputs.length)

bin/btc-server/src/signer.rs
```diff
@@ -402,4 +376,55 @@ where
...
+        // check extra outputs are change outputs:
+        // psbt should only have one change output
+        if additional_outputs.len() != 1 {
+            return Err(SigningFinalizeError::InvalidChangeOutput);
+        }
+
+        // TxOut scriptpubkey should be scriptpubkey derived from aggregated public key
+        let agg_pk = public_key_package.verifying_key().to_secp_pk().expect("valid secp pk");
+        let expected_script_pubkey = generate_taproot_change_scriptpubkey(&agg_pk);
+        let tx = psbt.clone().extract_tx()?;
```
arminsabouri: Nit: can we remove the last clone here?

bin/btc-server/src/signer.rs
```diff
@@ -402,4 +376,55 @@ where
...
+        let has_change = tx.output.iter().any(|o| o.script_pubkey == expected_script_pubkey);
```
arminsabouri: There are txs that may not have change

I think the check here should be 
```rust
if change_outputs.len() > 0 && tx.output.iter().any(|o| o.script_pubkey == expected_script_pubkey).len() == 0 {
            return Err(SigningFinalizeError::InvalidChangeOutput);

}

File: bin/test-suite/src/suite/consensus/frost/test_signing.rs (hunk 6 above)
Comment (arminsabouri) [reply bin/test-suite/src/suite/consensus/frost/test_signing.rs]: Yeah sorry about that I must have introduced that. FWIW it should be resolved on main now

**Review by arminsabouri — COMMENTED** [review]
Overall looking good. Just a couple nitpicks and a comment about requiring change to be present.

File: bin/btc-server/src/util.rs [hunk 11]
```diff
@@ -209,12 +212,17 @@ pub enum ValidatePSBTError {
...
     flags: u8,
     min_signers: u16,
-    db: &database::Db,
-) -> Result<(), ValidatePSBTError> {
+    app: &crate::App<BitcoinRpcApi>,
+) -> Result<(), ValidatePSBTError>
+where
+    BitcoinRpcApi: RpcApi,
+    BitcoinRpcApi: Sync,
+    BitcoinRpcApi: std::marker::Send,
```
arminsabouri:
```suggestion
    BitcoinRpcApi: RpcApi + Sync + std::marker::Send,
```

bin/btc-server/src/util.rs
```diff
@@ -209,12 +212,17 @@ pub enum ValidatePSBTError {
...
 /// never add more than a quorum of signers `ROUND2_TRANSITION`: Validates partial signatures during
 /// the transition to round 2, ensuring signers match and Frost IDs align.
-pub fn validate_psbt(
+pub fn validate_psbt<BitcoinRpcApi: 'static>(
     psbt: &Psbt,
     flags: u8,
     min_signers: u16,
-    db: &database::Db,
-) -> Result<(), ValidatePSBTError> {
+    app: &crate::App<BitcoinRpcApi>,
```
arminsabouri: Why pass the whole app down? It seems like even with the changes introduced we only need the db

arminsabouri (commented):
Just noticed two more things

bin/btc-server/src/util.rs
scottmillner: ah yep i meant to revert that change but forgot, only need db, thanks for catching it!

bin/btc-server/src/signer.rs
scottmillner: the change output(s) won't have an id so can do `outputs.length - 1` if we're sure only 1 change output

bin/btc-server/src/signer.rs
scottmillner: `psbt` is a shared reference passed to the method and `extract_tx()` moves `psbt` so need to clone 
and i don't want `validate_outputs` to take ownership of `psbt` so the caller can continue to use it if needed

bin/btc-server/src/signer.rs
scottmillner: `There are txs that may not have change` - that's why im using `.any()` and if there are none then `has_change` is false and errors

we already have the following check above so don't need `change_outputs.len() > 0`
```
if change_outputs.len() != 1 {
            return Err(SigningFinalizeError::InvalidChangeOutput);
        }
```
so think this is good how it is

bin/btc-server/src/signer.rs
scottmillner: done!

bin/btc-server/src/util.rs
scottmillner: no longer needed since `validate_outputs` only needs db

bin/btc-server/src/util.rs
scottmillner: done!

bin/btc-server/src/signer.rs
arminsabouri: Makes sense. Thanks

arminsabouri (approved):
Thanks for addressing the feedback. ACK!

PR #455, opened 2024-10-22
feat(repo): Added implementation for fed-utils for `get_balance` and `sweep_balance`

nagarajm87:
* Added implementation `client` and `rpc_method`.
* Added implementation `get_balance` and `sweep_balance`.
* updated `utils-cli` with new commands.
* Handling `error` crate. 
* Added test cases.
* Added method to support config.toml.
* Added method for deserialize private_key_validator.json


scottmillner:
looking good so far! will review again when not draft

nagarajm87:
1. I have kept `wallet.rs` bcoz it contains some test cases, which was written by @scottmillner , once its reviewed, if everything is good, will delete it. 
2. For `sweep_balance` do we need to add `amount` params ? 
3. I have used chain_id `3636` to do transactions, its working for `sweep_balance`

rwlockbg:
@nagarajm87 Can you pls rebase as there is 1 conflict ?

scottmillner:
~~do you mind adding a README that gives a general description of how the wallet is used?~~
~~i realize i forgot to do that, sorry~~

~~for example, instead of the user saving a config, the user can set env vars~~

RESOLVED

scottmillner:
~~just a general question:~~
~~how did you decide on env vars over the user setting a config that will be saved?~~

~~it looks like the only env var available to the wallet is `SECRET_KEY_PATH`~~
~~which means the user has to pass in `chain_id`, `receiver_address` every time? is that correct?~~

~~not saying either approach is better just want to get group consensus on what we think node operators will prefer~~
@0xBEEFCAF3 @rwlockbg 

RESOLVED

bin/federation-utils/src/client.rs
```diff
@@ -0,0 +1,139 @@
...
+}
+
+impl Client {
+    /// Asynchronous function to create a new client instance
+    pub async fn new(url: &str) -> Result<Self, WalletError> {
+        let provider = Provider::<Http>::try_from(url)
+            .map_err(|e| WalletError::IoError(format!("Failed to connect: {:?}", e)))?;
+        Ok(Client {
+            client: Arc::new(provider),
+            provider_url: url.to_string(),
```
scottmillner: this is hardcoded to 1 which is ethereum mainnet 

the user could use this wallet with any evm compatible chain but imagine it will be used with a chain running our codebase

bc of that i'd prefer not to hardcode the chain id or we could default to ours

thoughts?

bin/federation-utils/src/cli/mod.rs
```diff
@@ -1,53 +1,39 @@
 use clap::{Parser, Subcommand};
-
 #[derive(Parser, Debug)]
 #[command(name = "Wallet CLI")]
```
scottmillner: do you think we should allow the user to set any additional env vars besides `SECRET_KEY_PATH`?

rwlockbg (dismissed):

nagarajm87:
> just a general question: how did you decide on env vars over the user setting a config that will be saved?
> 
> it looks like the only env var available to the wallet is `SECRET_KEY_PATH` which means the user has to pass in `chain_id`, `receiver_address` every time? is that correct?
> 
> not saying either approach is better just want to get group consensus on what we think node operators will prefer @0xBEEFCAF3 @rwlockbg

Regarding the environment variable, the user or node operator can set it directly, or the node operator can provide the SECRET_KEY_PATH manually—either option works. Alternatively, we could create a config file and load the required data from there, and I’m open to suggestions on which approach would be best.

bin/federation-utils/src/cli/mod.rs
nagarajm87: user can pass, `chain_id`, `privded_url`, `address`, & `SECRET_KEY_PATH`,  one thing need to confirmation regarding `amount` for sweep_balance.

bin/federation-utils/src/client.rs
nagarajm87: @scottmillner  thank you :) I have updated it.

bin/federation-utils/src/cli/mod.rs
scottmillner: i mean do we always want to make the user pass those values?
from what i could see, there is only one env var being used, is that correct?

scottmillner:
> > just a general question: how did you decide on env vars over the user setting a config that will be saved?
> > it looks like the only env var available to the wallet is `SECRET_KEY_PATH` which means the user has to pass in `chain_id`, `receiver_address` every time? is that correct?
> > not saying either approach is better just want to get group consensus on what we think node operators will prefer @0xBEEFCAF3 @rwlockbg
> 
> Regarding the environment variable, the user or node operator can set it directly, or the node operator can provide the SECRET_KEY_PATH manually—either option works. Alternatively, we could create a config file and load the required data from there, and I’m open to suggestions on which approach would be best.

~~curious on which option you think node operators prefer?~~

RESOLVED

bin/federation-utils/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+```bash
+utils [OPTIONS] <COMMAND>
+Commands:
+  get-balance    Get balance of an address
+  sweep-balance  Sweep the balance to another address
+  help           Print this message or the help of the given subcommand(s)
+
+Options:
+  -c, --chain-id <CHAIN_ID>          Chain ID (default is 1) [default: 11155111]
```
rwlockbg: @nagarajm87 I think this should be 3636 (our chain Id)

bin/federation-utils/src/cli/mod.rs
```diff
@@ -1,53 +1,39 @@
...
 pub(crate) struct Cli {
     #[command(subcommand)]
     pub(crate) command: Commands,
+    /// Chain ID (default is 11155111)
+    #[arg(short, long, default_value_t = 11155111)]
```
rwlockbg: again, I think this should be 3636 (our chain  id)

bin/federation-utils/src/client.rs
```diff
@@ -0,0 +1,146 @@
...
+            address: None,          // Default as None
+            provider_url: url.to_string(),
+        };
+
+        Ok(client)
+    }
+
+    /// Synchronous function to create a new client instance using a synchronous block_on
+    pub fn new_sync(url: &str, chain_id: u64) -> Result<Client, WalletError> {
+        let runtime = tokio::runtime::Builder::new_current_thread()
```
rwlockbg: Question: why do we always create a new runtime for these commands ? The runtime could  be initialized on the main method if needed. First, we dont need all features of it (enable_all()) and I also think we dont need tihs to be inside the funcs. How about we move it to main. Putting #[tokio::main] implicitly starts the runtime

bin/federation-utils/src/main.rs
```diff
@@ -6,61 +6,35 @@
...
-            let pretty_config = serde_json::to_string_pretty(&config).expect("Valid JSON");
-            println!("Wallet config: {}", pretty_config);
-        }
-        Commands::GenerateKey(_) => {
-            println!("Generating key...");
-        }
-        Commands::GetBalance(_) => {
+        Commands::GetBalance(get_balance) => {
             println!("Getting balance...");
+            let bal = handle_get_balance(get_balance, &cli.provider_url, cli.chain_id)
```
rwlockbg: if we add #[tokio::main], the cli utils could directly call async .await here and we could drop the runtime we are creating everywhere. Creating a runtime is a heavy operation and is not needed. Is there a reason why we want our cli to be fully sync ?

rwlockbg:
> > > just a general question: how did you decide on env vars over the user setting a config that will be saved?
> > > it looks like the only env var available to the wallet is `SECRET_KEY_PATH` which means the user has to pass in `chain_id`, `receiver_address` every time? is that correct?
> > > not saying either approach is better just want to get group consensus on what we think node operators will prefer @0xBEEFCAF3 @rwlockbg
> > 
> > 
> > Regarding the environment variable, the user or node operator can set it directly, or the node operator can provide the SECRET_KEY_PATH manually—either option works. Alternatively, we could create a config file and load the required data from there, and I’m open to suggestions on which approach would be best.
> 
> curious on which option you think node operators prefer?

I believe having both might be confusing. Best is to have a .toml file that contains all info needed and we just pass over the cli a path where the file is located. This is how geth and all the others work. Of course, if need be, we could have the args also added, but that  might be too much. Adding ENVs on top of that is an overkill.

bin/federation-utils/src/cli/mod.rs
scottmillner: for `sweep_balance`: i think it'd be nice to allow an `amount` param that is an Option<> so they can sweep less than the full balance if they want

if they don't pass `amount`, the full balance will be sweeped

arminsabouri:
@nagarajm87 to squash relevant commits

bin/federation-utils/src/cli/mod.rs
```diff
@@ -5,49 +5,118 @@ use clap::{Parser, Subcommand};
...
+pub(crate) struct GetBalance {
+    /// Address to check the balance of
+    #[arg(short, long)]
+    pub address: Option<String>,
+}
 
-    /// Path to the wallet config file (optional)
-    #[arg(long)]
-    pub(crate) provider_url: Option<String>,
+#[derive(Parser, Debug)]
```
scottmillner, line 40: should we get remove `env = "SECRET_KEY_PATH"` since we have a config now?

bin/federation-utils/src/cli/mod.rs
```diff
@@ -18,30 +19,104 @@ pub(crate) struct Cli {
 }
 #[derive(Subcommand, Debug)]
 pub(crate) enum Commands {
+    /// Init create config.toml in home directory.
```
scottmillner, line 23: i think we should add a GetConfig command so the user can see what their config values are? thoughts?

bin/federation-utils/src/main.rs
```diff
@@ -6,69 +6,102 @@
...
-            } = setup;
-            let receiver_addres_hash =
-                receiver_address.as_ref().map(|addr| addr.parse().expect("Valid address"));
-            let config = WalletConfig::new(
-                *chain_id,
-                config_path.clone(),
-                secret_key_destination.clone(),
-                provider_url.clone(),
-                receiver_addres_hash,
-            )?;
```
scottmillner, line 52: let's get rid of this `unwrap` and handle the case where the user hasn't passed in the `config_path`

as is, this will break if the user doesn't pass in `config_path` which is an Option
if you run `cargo run --package federation-utils-cli --bin utils init`

bin/federation-utils/src/handler.rs
```diff
@@ -0,0 +1,167 @@
...
+}
+
+/// Deseialise Ed25519 or Secp256k1 private key and return stirng.
+fn deserialize_secret_key(seceret_path: &str) -> Result<String, WalletError> {
+    let json_data = fs::read_to_string(seceret_path)?;
+    let parsed_key: TendermintKey = serde_json::from_str(&json_data)
+        .map_err(|e| WalletError::CustomError(format!("Failed to parse JSON: {:?}", e)))?;
+
+    match parsed_key.priv_key.key_type.as_str() {
+        "tendermint/PrivKeyEd25519" => {
```
scottmillner: let's error here instead since a user should only be using secp curve in comet

bin/federation-utils/src/config.rs
```diff
@@ -0,0 +1,53 @@
...
+        }
+    } else {
+        Config::default()
+    }
+}
+//create config file insdie the home direcotry.
+pub(crate) fn get_default_config_path() -> PathBuf {
+    dirs::home_dir().expect("Failed to get home directory").join("fed-utils").join("config.toml")
+}
+
```
scottmillner, line 31: looks like if i use the `init` command, i will get default values only
i have no way to override those values unless i manually create the `config.toml` file, is that correct?

if so, and you don't think we should let the user override those values through a command, can you update the README to tell the user they have to manually create that file if they don't want the default values

nagarajm87: `init` command which creates `config.toml` file with default values  and user has to update  `config.toml` with theirs address or `seceret_path`  all values will be loaded from config.toml or else user can pass via cli command.

bin/federation-utils/src/handler.rs
nagarajm87: As we discussed this in yesterday meeting, will update it, thanks :)

bin/federation-utils/src/cli/mod.rs
nagarajm87: yes, we will loading via config.toml or by cli command

arminsabouri:
Remaining TODOs 
* Squash relevant commits
* Update fed utils README

scottmillner:
once the TODOs Armin listed are done, it's good to go on my end and will ✅

bin/federation-utils/Cargo.toml
```diff
@@ -15,15 +15,20 @@ name = "utils"
...
 url = "2.5.0"
 serde.workspace = true
 serde_json.workspace = true
 tempfile.workspace = true
-
-
+tokio = { version = "1.37.0", features = ["full"] }
+dirs = "5.0.1"
+toml = "0.8.19"
+ed25519-dalek = { version = "2.1.1", features = ["pkcs8"] }
```
arminsabouri: Do we need this dependency?

bin/federation-utils/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+### GetBlance :
+
+* Config toml
+```bash
+ ./target/debug/utils  -p {path_to_your_config.toml} get-balance 
+```
+* CLI Command
+```bash
+  ../../target/debug/utils -c 36363 -u http://localhost:8545  get-balance -a <your address>
```
arminsabouri: Assuming this this chain id?

```suggestion
  ../../target/debug/utils -c 3636 -u http://localhost:8545  get-balance -a <your address>
```

bin/federation-utils/README.md
```diff
@@ -0,0 +1,70 @@
...
+* CLI Command
+```bash
+  ../target/debug/utils -c 36363 -u http://localhost:8545  sweep-balance -s <SECRET_KEY_PATH> -r <RECEIVER_ADDRESS> -a <AMOUNT>
+```
+
+### Config.toml  
+```
+chain_id = 3636
+provider_url = "http://localhost:8545"
+from_address = ""                  // This addrress is used for get-balance.
```
arminsabouri: Small typo here
"address"

bin/federation-utils/src/cli/mod.rs
```diff
@@ -5,49 +5,119 @@ use clap::{Parser, Subcommand};
...
-    #[arg(long)]
-    pub(crate) chain_id: u64,
-
-    /// Path to the secret key output file
-    #[arg(long)]
-    pub(crate) secret_key_destination: Option<String>,
+pub(crate) struct GetBalance {
+    /// Address to check the balance of
+    #[arg(short, long)]
+    pub address: Option<String>,
```
arminsabouri: should this take an argument? I think we are always checking the balance of the federation wallet. Not any address

bin/federation-utils/src/cli/mod.rs
```diff
@@ -5,49 +5,119 @@ use clap::{Parser, Subcommand};
...
-    #[arg(long)]
-    pub(crate) config_path: Option<String>,
+    #[arg(short, long)]
+    pub receiver_address: Option<String>, // Changed to Option<String>
 
-    /// Address to receive funds (optional)
-    #[arg(long)]
-    pub(crate) receiver_address: Option<String>,
+    #[arg(short, long)]
+    pub amount: Option<String>,
```
arminsabouri: How can these be optional? Are there defaults?

bin/federation-utils/src/client.rs
```diff
@@ -0,0 +1,148 @@
...
+    }
+    //sweep balance from one account to another account
+    async fn sweep_balance(
+        &self,
+        to: Address,
+        amount: Option<U256>,
+    ) -> Result<TxHash, WalletError> {
+        let provider = Arc::clone(&self.client);
+
+        let wallet: LocalWallet = match &self.secret_key {
```
arminsabouri: The client struct should probably not get initialized if there is no secret key. And then you dont need these match statments

bin/federation-utils/src/client.rs
```diff
@@ -0,0 +1,148 @@
...
+
+        let mut chain_id: u64 = self.chain_id;
+
+        if self.chain_id == 0 {
+            chain_id = provider
+                .get_chainid()
+                .await
+                .map_err(|e| WalletError::GasError(format!("Failed to get chain Id: {}", e)))?
+                .as_u64();
+        }
```
arminsabouri: I'm not sure I understand this check. We should always expect a set of valid chain ids OR just pull it from provider when we are init'ing the client OR pass in a is-testnet flag that toggles between testnet and mainnet chain id

bin/federation-utils/src/client.rs
```diff
@@ -0,0 +1,148 @@
...
+
+        let send_amount = match amount {
+            Some(amt) => {
+                if amt + total_gas_fee > balance {
+                    return Err(WalletError::CustomError(
+                        "Insufficient balance to cover transfer amount and gas fees".to_string(),
+                    ));
+                }
+                amt
```
arminsabouri: Let the EVM tell you there is insufficient balance. You dont need to do this.

bin/federation-utils/src/client.rs
```diff
@@ -0,0 +1,148 @@
...
+        };
+
+        let from_address = wallet.address();
+        let balance = self.get_balance(from_address).await?; // Await get_balance
+
+        if balance.is_zero() {
+            return Err(WalletError::RpcError("Insufficient funds.".to_string()));
+        }
+
+        let tx: TypedTransaction = TransactionRequest::new().to(to).from(from_address).into();
```
arminsabouri, line 68: Ethers has much easier interface for sending from a local wallet 
See [this example](https://docs.infura.io/tutorials/ethereum/send-a-transaction/use-rust)

bin/federation-utils/src/client.rs
```diff
@@ -0,0 +1,148 @@
...
+    /// Chain ID for the blockchain network
+    pub chain_id: u64,
+    /// Secret key (private key) for the client
+    pub secret_key: Option<String>,
+    /// URL of the blockchain provider
+    pub provider_url: String,
+    /// Optional receiver address for transactions
+    pub receiver_address: Option<Address>,
+    /// Optional address for the client itself
+    pub address: Option<Address>,
```
arminsabouri: The address can always be generated from the sk. 
`address = keccak256(sk * G)` 
Let ethers do this

bin/federation-utils/src/client.rs
```diff
@@ -0,0 +1,148 @@
...
+    /// Shared HTTP provider client
+    pub client: Arc<Provider<Http>>,
+    /// Chain ID for the blockchain network
+    pub chain_id: u64,
+    /// Secret key (private key) for the client
+    pub secret_key: Option<String>,
+    /// URL of the blockchain provider
+    pub provider_url: String,
+    /// Optional receiver address for transactions
+    pub receiver_address: Option<Address>,
```
arminsabouri: Why are we storing this in a client struct? 
perhaps the client should just store the neccecary data needed to send txs, view balance, etc... and the methods such as `sweep_balance` just take in reciever_address as a param.

bin/federation-utils/src/config.rs
```diff
@@ -0,0 +1,53 @@
...
+impl Default for Config {
+    fn default() -> Self {
+        Config {
+            chain_id: 3636,
+            provider_url: "http://localhost:8545".to_string(),
+            from_address: Some("".to_string()),
+            receiver_address: Some("".to_string()),
+            secret_path: Some("".to_string()),
+            amount: Some("10000000000000000".to_string()),
```
arminsabouri: Defaults here are scary. These values should probably not be in config. 
Also do we need and/or use amount anywhere? we should be just sweeping

bin/federation-utils/src/handler.rs
```diff
@@ -0,0 +1,146 @@
...
+    let tx_hash = match client.sweep_balance(to_address, Some(amount)).await {
+        Ok(hash) => Ok(hash),
+        Err(e) => {
+            return Err(WalletError::RpcError(e.to_string()));
+        }
+    };
+    tx_hash
+}
+
+/// Deseialise Ed25519 or Secp256k1 private key and return stirng.
```
arminsabouri, line 151: Must be secp256k1

arminsabouri: panic otherwise

bin/federation-utils/src/main.rs
```diff
@@ -6,69 +6,100 @@
...
+                })?;
+
+            let amount_str =
+                sweep_balance.amount.clone().or(config.amount.clone()).ok_or_else(|| {
+                    WalletError::CustomError(
+                        "Amount must be provided via CLI or config".to_string(),
+                    )
+                })?;
+
+            let amount = U256::from_dec_str(&amount_str).expect("Failed to parse decimal to U256");
```
arminsabouri: Same comment about amount as above. Since we are sweeping do we need to pass in amount?

arminsabouri (changes requested):

bin/federation-utils/README.md
nagarajm87: yes, its `chain_id`, we can either use as `-c` or `-chain-id` its already mentioned in readme.

bin/federation-utils/Cargo.toml
nagarajm87: I have removed, not push yet, after disscussion yesterday meeting

bin/federation-utils/src/handler.rs
nagarajm87: yep, it will return error mentioning `key_type not supported`

bin/federation-utils/src/client.rs
nagarajm87: balance value which I receive from `rpc_call` from `evm`, then I am comparing it with amount passed by the user.

bin/federation-utils/src/cli/mod.rs
nagarajm87: need to discuss :)

bin/federation-utils/src/client.rs
nagarajm87: oops, i already removed it, not pushed yet. will update it

bin/federation-utils/src/main.rs
nagarajm87: need to discuss, amount params, allowing the user to either send their entire balance or specify a desired amount.

bin/federation-utils/src/cli/mod.rs
nagarajm87: These fields are optional in the CLI because they can be loaded from the config.toml file, if not provided directly as command-line arguments.

bin/federation-utils/src/main.rs
arminsabouri: The way we have spec'd this out was to send the entire balance

rwlockbg (dismissed):

rwlockbg:
@nagarajm87 need to rebase the conflicts again, sorry

arminsabouri:
@nagarajm87 Can you also squash the commits. Thanks

bin/federation-utils/src/client.rs
```diff
@@ -0,0 +1,142 @@
+use crate::errors::WalletError;
+use async_trait::async_trait;
+use ethers::{prelude::*, types::transaction::eip2718::TypedTransaction};
+use std::{str::FromStr, sync::Arc};
+/// wallet trait
+#[async_trait]
+pub trait Wallet {
+    /// getbalnce  for given address
+    async fn get_balance(&self, sceret_key: String) -> Result<U256, WalletError>;
```
arminsabouri: Secret key typo

bin/federation-utils/src/client.rs
```diff
@@ -0,0 +1,142 @@
...
+    async fn get_transaction_info(&self, tx_hash: TxHash) -> Result<Transaction, WalletError>;
+}
+
+/// rpc client struct
+#[derive(Debug)]
+pub struct Client {
+    /// Shared HTTP provider client
+    pub client: Arc<Provider<Http>>,
+    /// Chain ID for the blockchain network
+    pub chain_id: u64,
```
arminsabouri, line 22: Can we pull this from the provider or is there a another reason to pass this in

arminsabouri (commented):
Great progress! Looks like some code spell and other ci jobs failing. I'll let @scottmillner do another pass as well

bin/federation-utils/src/client.rs
nagarajm87: last time as discussed to keep it, `chain id` can be set via config or cli or can be fetch from provider, handling all cases :)

arminsabouri: but why?

bin/federation-utils/README.md
```diff
@@ -0,0 +1,74 @@
...
+```
+
+### GetBlance :
+
+* Using  config toml
+```bash
+ ./target/debug/utils  --config-path {path_config.toml} get-balance 
+```
+* CLI Command
+```bash
```
scottmillner, line 43: is --chain-id 36363 intentional or did you mean 3636?

nagarajm87: its 3636, my bad

scottmillner: all good

im gonna ✅ and if you update the readme, ill ✅ again

nice job!!

scottmillner (dismissed):

rwlockbg (approved):

scottmillner (approved):

arminsabouri (approved):

PR #472, opened 2024-11-04
fix(federation): validate signer count in federation.toml config #766

nagarajm87:
* Sanity check for the maximum number of signers loaded from federation.toml in the POA node.


arminsabouri (approved):
Nice work!

PR #473, opened 2024-11-06
feature(735-handle-missing-federation-member-pk): handling  federation toml missing public keys

nagarajm87:
* handling  `federation.toml` missing public keys.

### DIFF
--- a/bin/reth/src/commands/poa/mod.rs
+++ b/bin/reth/src/commands/poa/mod.rs
@@ -504,7 +504,7 @@ impl<Ext: clap::Args + fmt::Debug> PoaNodeCommand<Ext> {
         if let Some(max_signers) = rpc.max_signers {
             if federation_authorities.len() != max_signers as usize {
                 return Err(eyre::eyre!(
-                    "Error: max_signers does not match the length of federation_authorities"
+                    "max_signers does not match the length of federation_authorities"
                 ));
             }
         }
@@ -621,26 +621,34 @@ impl<Ext: clap::Args + fmt::Debug> PoaNodeCommand<Ext> {
 
         if let (Some(min_signers), Some(max_signers)) = (rpc.min_signers, rpc.max_signers) {
             if min_signers > max_signers {
-                return Err(eyre::eyre!(
-                    "Error: min_signers should be less than or equal to max_signers"
-                ));
+                return Err(eyre::eyre!("min_signers should be less than or equal to max_signers"));
             }
         }
         // create frost config if in federation mode
         let frost_config = if is_fed_node {
-            let authority_index = genesis_authorities
-                .iter()
-                .position(|a| a == &authority_pk)
-                .expect("Authority not found in genesis authorities");
+            let authority_index = match genesis_authorities.iter().position(|a| a == &authority_pk)
+            {
+                Some(index) => index,
+                None => {
+                    error!("Federation public keys not found ");
+                    Default::default()
+                }
+            };
             let config = FrostConfig::new(
                 authority_pk,
                 authority_index,
                 genesis_authorities.clone(),
-                node_config.rpc.min_signers.expect("min signers"),
-                node_config.rpc.max_signers.expect("max signers"),
+                node_config.rpc.min_signers.unwrap_or_else(|| {
+                    error!("min signers not specified");
+                    Default::default()
+                }),
+                node_config.rpc.max_signers.unwrap_or_else(|| {
+                    error!("max signers not specified");
+                    Default::default()
+                }),
             );
-            info!(target: "reth::cli", "Frost config initialized");
 
+            info!(target: "reth::cli", "Frost config initialized");
             Some(config)
         } else {
             None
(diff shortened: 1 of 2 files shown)


bin/reth/src/commands/poa/mod.rs
```diff
@@ -621,26 +621,34 @@ impl<Ext: clap::Args + fmt::Debug> PoaNodeCommand<Ext> {
...
-            let authority_index = genesis_authorities
-                .iter()
-                .position(|a| a == &authority_pk)
-                .expect("Authority not found in genesis authorities");
+            let authority_index = match genesis_authorities.iter().position(|a| a == &authority_pk)
+            {
+                Some(index) => index,
+                None => {
+                    error!("Federation public keys not found ");
+                    Default::default()
```
arminsabouri: Should we error out of the program here? If a node is saying they are a fed node but their key is not found maybe we should just exit

arminsabouri (changes requested):

bin/reth/src/commands/poa/mod.rs
```diff
@@ -781,10 +787,15 @@ impl<Ext: clap::Args + fmt::Debug> PoaNodeCommand<Ext> {
...
             RandomSourceProvider::new(),
             canon_state_notification_receiver,
-        )
-        .expect("Failed to create authority consensus builder")
-        .build()
-        .await;
+        ) {
+            Ok(consensus) => consensus.build().await,
+            Err(e) => {
+                return Err(eyre::eyre!(
```
scottmillner: `try_new` can fail for other reasons besides `missing federation public keys`

crates/node/core/src/args/federation_args.rs
```diff
@@ -84,19 +88,22 @@ impl FederationTomlConfig {
     ) -> Result<Vec<(secp256k1::PublicKey, SocketAddr)>, Error> {
         let federation_members = self
             .federation_member_public_key
-            .iter()
-            .map(|key| {
-                let public_key = secp256k1::PublicKey::from_str(&key.key)
-                    .expect("Invalid hex string for PublicKey");
+            .as_ref() // Get a reference to the inner Vec if it exists
```
rwlockbg: Can you remove this comment please ?

crates/node/core/src/args/federation_args.rs
```diff
@@ -84,19 +88,22 @@ impl FederationTomlConfig {
...
+            .map_or(Vec::new(), |members| {
+                members
+                    .iter()
+                    .map(|member| {
+                        let public_key = secp256k1::PublicKey::from_str(&member.key)
+                            .expect("Invalid hex string for PublicKey");
```
rwlockbg: Can we not expect here, but rather collect all Result<PubKey, Error> and check i any are Err typed and then create aan Error message just to know the exact reason ?

crates/node/core/src/args/federation_args.rs
```diff
@@ -84,19 +88,22 @@ impl FederationTomlConfig {
...
 
-                let soc_addr = key.socket_addr.parse::<SocketAddr>().unwrap();
-                (public_key, soc_addr)
-            })
-            .collect::<Vec<(secp256k1::PublicKey, SocketAddr)>>();
+                        let soc_addr = member.socket_addr.parse::<SocketAddr>().unwrap();
```
rwlockbg: same here - to avoid the unwraps somwhow

bin/reth/src/commands/poa/mod.rs
rwlockbg: agree, maybe you can check the err type (if possible) and then react to it ?

arminsabouri: +1

crates/node/core/src/args/federation_args.rs
```diff
@@ -45,7 +45,7 @@ pub struct FedMemberPubKey {
 #[serde(deny_unknown_fields, rename_all = "kebab-case")]
 pub struct FederationTomlConfig {
     /// federation members public keys
-    pub federation_member_public_key: Vec<FedMemberPubKey>,
+    pub federation_member_public_key: Option<Vec<FedMemberPubKey>>,
```
arminsabouri: I dont think this should ever be optional

arminsabouri (changes requested):

rwlockbg (dismissed):

bin/reth/src/commands/poa/mod.rs
```diff
@@ -621,26 +621,32 @@ impl<Ext: clap::Args + fmt::Debug> PoaNodeCommand<Ext> {
...
         // create frost config if in federation mode
         let frost_config = if is_fed_node {
-            let authority_index = genesis_authorities
-                .iter()
-                .position(|a| a == &authority_pk)
-                .expect("Authority not found in genesis authorities");
+            let authority_index =
+                genesis_authorities.iter().position(|a| a == &authority_pk).ok_or_else(|| {
+                    error!("Federation public keys not found");
+                    eyre::eyre!("Federation public keys not found")
```
arminsabouri: key* singular not plural. 

Maybe a better error message is "Your public key could not be found in the list of federation public keys"

arminsabouri (changes requested):

arminsabouri (approved):

PR #479, opened 2024-11-11
fix(repo): Moved to broadcast channels

rwlockbg:


crates/net/network/src/frost/protocol.rs
```diff
@@ -388,7 +388,7 @@ impl Stream for FrostProtoConnection {
                 }
             }
             FrostProtoMessageKind::CoordinatorRound2SigningPackage(data) => {
-                if let Err(e) = protocol_events_tx.send(FrostProtocolEvent::PeerMessage {
```
arminsabouri: is send also blocking when its bounded?

rwlockbg: send in async, blocking_send is sync. Since all of these methods are sync, we cannot use await , so I had to move to blocking_send to suit the sync nature of the method.

rwlockbg: forget the above, I have re-factored everything to use broadcast channels now.

crates/consensus/authority/src/signing.rs
```diff
@@ -842,7 +841,6 @@ where
 
         // try to finalize the signing
         if let Ok(_sign_payload) = self.finalize_signing(signing_session_id).await {
-            info!(target: "consensus::authority::signing::coordinator_process_round2", "signing finalized!");
```
arminsabouri: Why remove this?

arminsabouri (commented):
Is this an attempt to fix a bug or just an improvment? I'm missing the why?

crates/consensus/authority/src/signing.rs
rwlockbg: sorry , my bad, put it back in.

PR #483, opened 2024-11-12
fix(frost): Signers validate psbt by ids

scottmillner:


rwlockbg (dismissed):

crates/consensus/authority/src/frost_task.rs
```diff
@@ -407,57 +407,161 @@ where
...
+                                };
+                                match validate_psbt_by_ids(
+                                    self.storage.client.clone(),
+                                    self.storage.btc_network,
+                                    &psbt_res,
+                                )
+                                .await
+                                {
+                                    Ok(_) => {
+                                        info!(target: "consensus::authority::frost_task::SignerRound1SigningPackage", "Validated psbt successfully")
```
arminsabouri: nit: these logs tend to just get noisy over time. 
maybe we can use
```rust
if let err(e) = ...
```
log only on errors and cut down on the lines of code

arminsabouri (commented):
Nit: only a cosmetic change

crates/consensus/authority/src/frost_task.rs
scottmillner: it is done!

arminsabouri (approved):

PR #490, opened 2024-11-15
feat(btc-server): Add rpc methods needed to sync wallet state

scottmillner:


rwlockbg (dismissed):

bin/btc-server/proto/btc_server.proto
```diff
@@ -156,6 +160,40 @@ message GetGatewayAddressResponse {
     string gateway_address = 3;
 }
 
+// TODO: remove
+// message Witness {
```
arminsabouri: could just store as a Vec<raw bytes>

bin/btc-server/proto/btc_server.proto
```diff
@@ -156,6 +160,40 @@ message GetGatewayAddressResponse {
...
+//     repeated bytes content = 1;
+//     uint32 witness_elements = 2;
+//     uint32 indices_start = 3;
+// }
+
+message TxIn {
+    OutPoint previous_outpoint = 1;
```
arminsabouri, line 173: missing 2

bin/btc-server/proto/btc_server.proto
```diff
@@ -156,6 +160,40 @@ message GetGatewayAddressResponse {
...
+    repeated TxOut output = 4;
+}
+
+message TrackedTx {
+    bytes txid = 1;
+    Transaction tx = 2;
+    repeated uint32 pegout_idxs = 3;
+    repeated PendingPegout pegout_requests = 4;
+    repeated uint32 change_idxs = 5;
+    google.protobuf.Timestamp created = 6;
```
arminsabouri, line 192: I wonder if we should replace this with bitcoin block height. It would allow this to be verifiable. And we can include this in a the wallet state merkle root

bin/btc-server/src/server.rs
```diff
@@ -750,4 +749,87 @@ where
...
+                        witness: tx_in.witness.to_vec(),
+                    })
+                    .collect::<Vec<_>>();
+                let tx_outs = tx
+                    .tx
+                    .output
+                    .into_iter()
+                    .map(|tx_out| rpc::TxOut {
+                        value: tx_out.value.to_sat(),
+                        script_pubkey: Some(rpc::ScriptBuf {
```
arminsabouri: Should this be optional?

arminsabouri (commented):
Great stuff! :fire:

bin/btc-server/proto/btc_server.proto
scottmillner: i ended up going with a vec

bin/btc-server/proto/btc_server.proto
scottmillner: yeah that makes sense, do you mind if i do that in a separate pr?
created a ticket for it: https://github.com/botanix-labs/botanix/issues/811

bin/btc-server/src/server.rs
scottmillner: no but this is what was autogenerated from the proto build

any idea on why it makes it optional? @rwlockbg 

bc the proto message is defined as 
```
message TxOut {
    // The value of the output, in satoshis.
    uint64 value = 1;
    // The script which must be satisfied for the output to be spent.
    ScriptBuf script_pubkey = 2;
}
```

bin/btc-server/proto/btc_server.proto
scottmillner: oof i can't count, nice catch!
fixed!

bin/btc-server/src/server.rs
rwlockbg: Protobuf members are always optional since syntax version 3 (the latest). This is the whole idea. You  can do something similar e.g.:
```

impl TxOut {
    pub fn validate(&self) -> Result<(), String> {
        if self.value == 0 {
            return Err("value must be greater than 0".to_string());
        }
        if self.script_pubkey.is_empty() {
            return Err("script_pubkey must not be empty".to_string());
        }
        Ok(())
    }
}
```

rwlockbg (dismissed):

bin/btc-server/src/server.rs
scottmillner: thanks for the clarification!!

bin/btc-server/proto/btc_server.proto
```diff
@@ -64,6 +66,12 @@ service BtcServer {
     rpc GetSigningStatus(GetSigningStatusRequest) returns (GetSigningStatusResponse) {}
 
     rpc GetSessionIds(GetSessionIdsRequest) returns (GetSessionIdsResponse) {}
+
+    rpc GetTrackedTxs(Empty) returns (GetTrackedTxsResponse) {}
+
+    rpc ResetAllPendingPegouts(ResetAllPendingPegoutsRequest) returns (Empty) {}
+
+    rpc ResetAllTrackedTxs(ResetAllTrackedTxsRequest) returns (Empty) {}
```
arminsabouri: Is there any case where we would reset one not the other? Maybe we shoud have a reset wallet state endpoint?

bin/btc-server/src/server.rs
```diff
@@ -750,4 +782,168 @@ where
...
+        req: tonic::Request<rpc::ResetAllTrackedTxsRequest>,
+    ) -> Result<tonic::Response<rpc::Empty>, tonic::Status> {
+        self.validate_jwt(&req)?;
+        let req = req.into_inner();
+        info!("Received reset all tracked txs request");
+        let tracked_txs = req
+            .tracked_txs
+            .into_iter()
+            .map(|tx| {
+                let tx_prost = tx.tx.expect("valid tx");
```
arminsabouri: This is related to https://github.com/botanix-labs/Macbeth/pull/490/files#r1846309253
According to @rwlockbg we should be error checking here. Otherwise an incorrect payload could panic the server.

bin/btc-server/src/server.rs
```diff
@@ -750,4 +782,168 @@ where
...
+    ) -> Result<tonic::Response<rpc::GetTrackedTxsResponse>, tonic::Status> {
+        self.validate_jwt(&request)?;
+        let db_tracked_txs = self.db.get_tracked_txs().map_err(|e| {
+            error!("Failed to get tracked txs: {}", e);
+            internal!("Failed to get tracked txs: {}", e)
+        })?;
+        let tracked_txs = db_tracked_txs
+            .into_iter()
+            .map(|tx| {
+                // create internal tx
```
arminsabouri: This comment applies to all the type conversions happening in server.rs. Is there a more rusty/idomatic way to do this that would reduce the code in server.rs? Maybe we can implement TryFrom for the types we are interested in? 

For example TxIn
```rust
impl TryFrom<prost::TxIn> for TxIn {
    type Error = ConversionError; // Custom error type
    
    fn try_from(prost_tx_in: prost::TxIn) -> Result<Self, Self::Error> {
        let previous_outpoint = prost_tx_in
            .previous_outpoint
            .ok_or(ConversionError::MissingField("previous_outpoint"))?;
        
        Ok(TxIn {
            previous_output: OutPoint {
                txid: Txid::from_slice(&previous_outpoint.txid)
                    .map_err(|_| ConversionError::InvalidField("txid"))?,
                vout: previous_outpoint.vout,
            },
            script_sig: ScriptBuf::from_bytes(
                prost_tx_in.script_sig
                    .ok_or(ConversionError::MissingField("script_sig"))?
                    .script,
            ),
            sequence: Sequence::from_consensus(prost_tx_in.sequence),
            witness: Witness::from_slice(&prost_tx_in.witness),
        })
    }
}
```

arminsabouri: also related to https://github.com/botanix-labs/Macbeth/pull/490#discussion_r1848814254

bin/btc-server/proto/btc_server.proto
scottmillner: created `reset_wallet_state` and removed the others

bin/btc-server/src/server.rs
scottmillner: implemented `validate()` for `TrackedTx`, `TxIn`, `TxOut`

bin/btc-server/src/server.rs
scottmillner: implemented TryFrom for all needed types

bin/btc-server/src/database.rs
```diff
@@ -638,6 +641,37 @@ impl Db {
...
+
+    /// Resets all pending pegouts, and re-adding the functions arguments back in
+    pub fn reset_pending_pegouts(
+        &self,
+        pegout_requests: &[&pegout_scheduler::PegoutRequest],
+    ) -> Result<(), Error> {
+        self.clear_pending_pegouts()?;
+        for pegout_request in pegout_requests.iter() {
+            self.store_pending_pegout(pegout_request)?;
+        }
```
arminsabouri: If you take a look at storing UTXOs we are using sleds atomic txs to perform this operation atomically. We need to consider the same thing with resetting and adding wallet state information

bin/btc-server/src/pegout_scheduler.rs
```diff
@@ -97,6 +100,88 @@ impl Tx {
...
+impl TryFrom<rpc::TrackedTx> for Tx {
+    type Error = tonic::Status;
+
+    fn try_from(tx: rpc::TrackedTx) -> Result<Self, Self::Error> {
+        if let Err(e) = tx.validate() {
+            error!("Invalid tracked tx: {}", e);
+            return Err(tonic::Status::invalid_argument(e));
+        }
+
+        let tx_prost = tx.tx.expect("valid tx");
```
arminsabouri: Do we still want to use expects here? Or can we return and error
`Err(MissingField("bitcoin transaction"))`

bin/btc-server/src/pegout_scheduler.rs
```diff
@@ -97,6 +100,88 @@ impl Tx {
...
+                id: PegoutId::from_bytes(&pegout.pegout_id).expect("valid pegout id"),
+                spk: ScriptBuf::from_bytes(pegout.spk),
+                value: Amount::from_sat(pegout.amount),
+                botanix_height: pegout.height,
+            })
+            .collect::<Vec<_>>();
+
+        Ok(Tx {
+            txid: Txid::from_slice(&tx.txid).expect("valid txid"),
+            tx: internal_tx,
```
arminsabouri, line 198: Just a thought. Belongs an different PR. bit we can reduce the payload size dramatically if we dont include tx and instead pull it from bitcoind. since we already have txid. Same could be same about change idx.

bin/btc-server/src/server.rs
```diff
@@ -752,4 +760,65 @@ where
...
+
+    async fn reset_wallet_state(
+        &self,
+        req: tonic::Request<rpc::ResetWalletStateRequest>,
+    ) -> Result<tonic::Response<rpc::Empty>, tonic::Status> {
+        self.validate_jwt(&req)?;
+        let req = req.into_inner();
+        info!("Received reset wallet state request");
+
+        // handle utxos
```
arminsabouri, line 798: The atomic db  comment applies here as well

arminsabouri (commented):
Awesome tests!! my only real comment is about atomic db transactions.

bin/btc-server/src/pegout_scheduler.rs
```diff
@@ -97,6 +100,88 @@ impl Tx {
...
+        let tx_prost = tx.tx.expect("valid tx");
+        let tx_ins = tx_prost
+            .input
+            .into_iter()
+            .map(|tx_in| {
+                if let Err(e) = tx_in.validate() {
+                    error!("Invalid tx input: {}", e);
+                    tonic::Status::invalid_argument(e);
+                }
+                let previous_outpoint = tx_in.previous_outpoint.expect("valid previous outpoint");
```
rwlockbg, line 127: I would say that since you have the associated Error above, we can replace all these expects with proper errors being thrown out o the TryFrom ?

bin/btc-server/src/pegout_scheduler.rs
```diff
@@ -97,6 +100,88 @@ impl Tx {
...
+            })
+            .collect::<Vec<_>>();
+
+        Ok(Tx {
+            txid: Txid::from_slice(&tx.txid).expect("valid txid"),
+            tx: internal_tx,
+            pegout_idxs: tx.pegout_idxs.into_iter().map(|idx| idx as usize).collect(),
+            pegout_requests,
+            change_idxs: tx.change_idxs.into_iter().map(|idx| idx as usize).collect(),
+            created: SystemTime::UNIX_EPOCH +
```
SatwikPrabhu, line 203: Can we use chrono instead of time? Is there a particular reason that I may be missing here?

bin/btc-server/src/rpc/btc_server.rs
```diff
@@ -1618,6 +1681,94 @@ pub mod btc_server_server {
...
+                        let codec = tonic::codec::ProstCodec::default();
+                        let mut grpc = tonic::server::Grpc::new(codec)
+                            .apply_compression_config(
+                                accept_compression_encodings,
+                                send_compression_encodings,
+                            )
+                            .apply_max_message_size_config(
+                                max_decoding_message_size,
+                                max_encoding_message_size,
+                            );
```
SatwikPrabhu, line 1721:
```suggestion
                    let inner = self.inner.clone();
                    let fut = async move {
                        let method = GetTrackedTxsSvc(inner);
                        let codec = tonic::codec::ProstCodec::default();
                        let mut grpc = tonic::server::Grpc::new(codec)
                            .apply_compression_config(
                                self.accept_compression_encodings,
                                self.send_compression_encodings,
                            )
                            .apply_max_message_size_config(
                                self.max_decoding_message_size,
                                self.max_encoding_message_size,
                            );
```

bin/btc-server/src/rpc/btc_server.rs
```diff
@@ -1618,6 +1681,94 @@ pub mod btc_server_server {
...
+                        let method = ResetWalletStateSvc(inner);
+                        let codec = tonic::codec::ProstCodec::default();
+                        let mut grpc = tonic::server::Grpc::new(codec)
+                            .apply_compression_config(
+                                accept_compression_encodings,
+                                send_compression_encodings,
+                            )
+                            .apply_max_message_size_config(
+                                max_decoding_message_size,
+                                max_encoding_message_size,
```
SatwikPrabhu, line 1765: Same as https://github.com/botanix-labs/Macbeth/pull/490/files#r1866537075

bin/btc-server/src/rpc/btc_server.rs
scottmillner: this file is autogenerated by prost-build (see comment at top of file)
so any changes like these will be overridden when ever the btc-server binary is compiled

if you're curious, prost-build uses the `build.rs` file
and you can see the build dependencies used for btc-server under `build-dependencies` in the `Cargo.toml`

bin/btc-server/src/pegout_scheduler.rs
scottmillner: chrono is an external dependency which we don't need for this binary (although it's in the `Cargo.toml` but should be removed)
- a general rule i try to follow is not to use a dependency if i can achieve the same goal with the std library
- unless the dependency allows for a much better approach

reducing dependencies will reduce binary size and dependency issues

in this case, the difference using std or chrono is trivial so i go with the std library

bin/btc-server/src/rpc/btc_server.rs
scottmillner: see above comment to see how this file is generated and why we don't make changes directly to this file

bin/btc-server/src/pegout_scheduler.rs
scottmillner: good idea, left a TODO to pull from bitcoind

bin/btc-server/src/pegout_scheduler.rs
scottmillner: the `validate()` call at the beginning of the map body does this:
- it validates all the optional fields contain valid values and if not, returns an error that is passed to `tonic::Status` which is the associated type error

i've added comments to make this clear

and the overall approach im using (mainly bc im inside a map call):
- if expect would err, return Tonic::status error
- then use expect knowing it's safe

bin/btc-server/src/pegout_scheduler.rs
scottmillner: the `validate()` call at the beginning of the map body does this:
- it validates all the optional fields contain valid values and if not, returns an error that is passed to `tonic::Status` which is the associated type error

i could move the `expect` calls into the `validate` method and return the some values but i wanted the `validate()` method to have only one purpose

so the pattern im using (mainly bc im inside a map call):
- if expect would err, return Tonic::status error
- then use expect knowing it's safe

bin/btc-server/src/database.rs
scottmillner: done!

bin/btc-server/src/server.rs
scottmillner: done!

bin/btc-server/src/pegout_scheduler.rs
```diff
@@ -97,6 +100,115 @@ impl Tx {
...
+
+        let tx_ins = tx_prost
+            .input
+            .into_iter()
+            .map(|tx_in| {
+                // validate tx_in optional fields contain valid values
+                // so the expect calls below are safe
+                if let Err(e) = tx_in.validate() {
+                    error!("Invalid tx input: {}", e);
+                    tonic::Status::invalid_argument(e);
```
rwlockbg: nice ! Just wondering if we should not use our macros for invalid_argument bad_Argument etc ?

bin/btc-server/src/pegout_scheduler.rs
rwlockbg: fine by me

rwlockbg (dismissed):
Very nice stuff! ApprovedQ

rwlockbg:
@scottmillner You might need to rebase and re-run the pipelines again as they seem to be badly stuck for some workflows!

bin/btc-server/src/pegout_scheduler.rs
scottmillner: done!

rwlockbg (dismissed):

rwlockbg (approved):

scottmillner:
i ran the int tests locally and they all passed

PR #497, opened 2024-11-19
fix(btc-server): missing pk

arminsabouri:
Fixes https://github.com/botanix-labs/botanix/issues/813
The problem we are fixing here is that we have added non-change outputs as change in the pegout scheduler.
The root cause is still unsure. Its possible that certain pending pegouts were added during the time of signing. 
However skipping non tr outputs during pegout scheduling should help

### DIFF
--- a/bin/btc-server/src/pegout_scheduler.rs
+++ b/bin/btc-server/src/pegout_scheduler.rs
@@ -10,8 +10,12 @@ use std::{
     time::{Duration, SystemTime},
 };
 
-use bitcoin::{Amount, Block, BlockHash, OutPoint, Transaction, TxOut, Txid};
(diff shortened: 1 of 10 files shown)


bin/btc-server/src/pegout_scheduler.rs
```diff
@@ -356,6 +356,10 @@ impl PegoutScheduler {
             // Add back the change to the utxo set
             let mut change_utxos = vec![];
             for (outpoint, output) in tx.change() {
+                if !output.script_pubkey.is_p2tr() {
```
arminsabouri: make this more accurate by re-generating the change spk

scottmillner (approved):

PR #501, opened 2024-11-20
feat(btc-server): Sync entire wallet state

scottmillner:
this pr wires up the wallet state sync ability but does not "turn it on": there's no task that invokes `sync_wallet_state()`

### DIFF
--- a/bin/reth/src/commands/poa/mod.rs
+++ b/bin/reth/src/commands/poa/mod.rs
@@ -841,26 +841,6 @@ impl<Ext: clap::Args + fmt::Debug> PoaNodeCommand<Ext> {
             //         healthcheck_task.expect("health check task exists").start_task().await;
             //     }),
             // );
-
-            // just for local testing purposes
-            // TODO: create WalletStateSync::start_task to handle this logic
-            // executor.spawn_critical(
-            //     "Wallet Sync Task",
-            //     Box::pin(async move {
-            //         loop {
-            //             tokio::time::sleep(Duration::from_secs(20)).await;
-
-            //             if let Err(err) = wallet_sync
-            //                 .as_ref()
-            //                 .expect("wallet sync task exists")
-            //                 .sync_wallet_state()
-            //                 .await
-            //             {
-            //                 error!("Error syncing wallet state: {:?}", err);
-            //             }
-            //         }
-            //     }),
-            // );
         }
 
         let initial_target = node_config.debug.tip;
--- a/bin/reth/src/commands/poa/mod.rs
+++ b/bin/reth/src/commands/poa/mod.rs
@@ -13,6 +13,7 @@ use futures::{stream_select, StreamExt, TryFutureExt};
 use reth_authority_consensus::{
     random_source_provider::RandomSourceProvider,
     utils::{is_known_minting_contract, retry_exec},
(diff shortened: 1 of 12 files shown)


bin/reth/src/commands/poa/mod.rs
```diff
@@ -839,6 +841,26 @@ impl<Ext: clap::Args + fmt::Debug> PoaNodeCommand<Ext> {
             //         healthcheck_task.expect("health check task exists").start_task().await;
             //     }),
             // );
+
+            // just for local testing purposes
```
rwlockbg: maybe remove this fro the repo if it does not go into prod ?

crates/consensus/authority/src/frost_task.rs
```diff
@@ -323,13 +400,29 @@ where
...
-                    PeerMessageResponse::Utxo(mut response) => {
+                    PeerMessageResponse::WalletState(mut response) => {
+                        // Only handle response if it has no state: responses with state are also
+                        // sent to WalletStateSyncEngine::sync_wallet_state
+                        // which updates the wallet state. This code block
+                        // handles sending our wallet state to a peer
+                        //
+                        // TODO: create separate messages for asking for wallet state and sending
+                        // wallet state
+                        if !response.utxos.is_empty() ||
```
rwlockbg: I think we might do some sort of validation func here that wraps all conditions together ?

crates/net/network/src/frost/messages.rs
```diff
@@ -551,6 +613,31 @@ impl FrostProtoMessage {
 
                 FrostProtoMessageKind::Healthcheck(HealthcheckRequest { sender, receiver })
             }
+            FrostProtoMessageId::WalletState => {
```
rwlockbg, line 616: Further down we have some ser/deser tests for the other message types, maybe we could add the WalletState to them too just for sanity and completeness?

rwlockbg (commented):
Nice work! In fact, its the old utxo sync that we had but with the pending pegouts extended. I wonder if we should create some integration tests to confirm it all works or leave this as a separate PR ?

scottmillner:
> Nice work! In fact, its the old utxo sync that we had but with the pending pegouts extended. I wonder if we should create some integration tests to confirm it all works or leave this as a separate PR ?

i was planning to do this is another pr since i want to stop work on this temporarily and focus on async pegouts

crates/consensus/authority/src/frost_task.rs
scottmillner: done!

crates/net/network/src/frost/messages.rs
scottmillner: done!

rwlockbg: nice!

crates/consensus/authority/src/frost_task.rs
rwlockbg: nice!

bin/reth/src/commands/poa/mod.rs
rwlockbg: thx

rwlockbg (approved):

PR #502, opened 2024-11-20
fix(abci): fixed retain height in abci commit function #814

nagarajm87:
*  fixed retain height in abci commit function.

### DIFF
--- a/crates/consensus/authority/src/comet_bft/abci.rs
+++ b/crates/consensus/authority/src/comet_bft/abci.rs
@@ -744,6 +744,21 @@ where
     /// docs: https://docs.cometbft.com/v0.38/spec/abci/abci++_methods#commit
     fn commit(&self) -> ResponseCommit {
         info!("commit request received");
+        let latest_block_height = match self.storage.client.last_block_number() {
+            Ok(height) => height as i64,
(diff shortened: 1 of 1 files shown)


crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -744,6 +744,21 @@ where
...
+                error!("Failed to fetch the latest block number, defaulting to 0: {:?}", e);
+                0
+            }
+        };
+
+        let candidate_blocks = match self.block_cache.write() {
+            Ok(guard) => guard,
+            Err(err) => {
+                error!("Failed to write to block cache: {:?}", err);
+                return ResponseCommit { retain_height: latest_block_height };
```
arminsabouri: This doenst seem right.
The documentation reads: 
> Blocks below this height may be removed. Defaults to 0 (retain all).

If latest block height is anything besides 0 we will prune historical blocks

arminsabouri (changes requested):

scottmillner (dismissed):

rwlockbg (approved):

scottmillner (approved):

PR #506, opened 2024-12-03
feat(repo): Added state sync impl

rwlockbg:


crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -376,6 +442,276 @@ where
...
+        let client = self.storage.client.clone();
+        match client.get_snapshots() {
+            Ok(snapshots) => snapshots.into_iter().fold(
+                ResponseListSnapshots { snapshots: vec![] },
+                |mut acc, snapshot| {
+                    acc.snapshots.push(Snapshot {
+                        height: snapshot.height(),
+                        format: SNAPSHOT_MESSAGE_FORMAT,
+                        chunks: snapshot.chunk_ids().len() as u32,
+                        hash: prost::bytes::Bytes::copy_from_slice(&snapshot.block_hash().0),
```
arminsabouri: should this be a hash of the snapshot itself? is there benefit for having this hash be the block hash?

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -376,6 +442,276 @@ where
...
+        if let Some(snapshot) = request.snapshot {
+            if snapshot.format != SNAPSHOT_MESSAGE_FORMAT {
+                return ResponseOfferSnapshot { result: SnapshotOfferResult::REJECT_FORMAT as i32 };
+            }
+
+            if snapshot.chunks == 0 {
+                return ResponseOfferSnapshot { result: SnapshotOfferResult::REJECT as i32 };
+            }
+
+            if snapshot.hash == prost::bytes::Bytes::default() {
```
arminsabouri, line 654: since the snapshot hash is a block hash shoudl we check if we already have this block?
and if we decide that the snapshot hash should be some unique hash that reps the snapshot we should also check if this snapshot hash exists in our db

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -376,6 +442,276 @@ where
...
+
+        if let Some(snapshot) = request.snapshot {
+            if snapshot.format != SNAPSHOT_MESSAGE_FORMAT {
+                return ResponseOfferSnapshot { result: SnapshotOfferResult::REJECT_FORMAT as i32 };
+            }
+
+            if snapshot.chunks == 0 {
+                return ResponseOfferSnapshot { result: SnapshotOfferResult::REJECT as i32 };
+            }
+
```
arminsabouri, line 653: Should we be checking for valid chunk bytes? empty bytes? Is there any structure to them?

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -376,6 +442,276 @@ where
...
+    /// https://docs.cometbft.com/v0.38/spec/abci/abci++_methods#applysnapshotchunk
+    fn apply_snapshot_chunk(
+        &self,
+        request: RequestApplySnapshotChunk,
+    ) -> ResponseApplySnapshotChunk {
+        info!("apply_snapshot_chunk request");
+        let client = self.storage.client.clone();
+
+        // get the last snapshot sync id - there should always be one provided the offer_snapshot
+        // has alredy run
```
arminsabouri: alredy -> already 
typo

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -376,6 +442,276 @@ where
...
+                    error!(target: "consensus::authority::abci", ?e, "Failed to insert block");
+                }
+            }
+
+            // set canonical head
+            client.set_canonical_head(sealed_block_with_senders.header.clone());
+            client.set_safe(sealed_block_with_senders.header.clone());
+            client.set_finalized(sealed_block_with_senders.header.clone());
+
+            // Should we update the fork choice?
```
arminsabouri: rm comment.

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -376,6 +442,276 @@ where
...
+                .blocking_recv()
+                .expect("to receive confirmation from fork choice updater");
+        }
+
+        return ResponseApplySnapshotChunk {
+            result: ApplySnapshotResult::Accept as i32,
+            refetch_chunks: vec![],
+            reject_senders: vec![],
+        };
+    }
```
arminsabouri, line 1018: Is it possible for multiple apply_snapshot_chunk to be processed in parallel? I dont think it is but I sitll want to surface this question. Do we need to apply a state sync lock?

arminsabouri (commented):
first part of the review on abci stuff

crates/storage/db/src/tables/mod.rs
```diff
@@ -307,6 +311,24 @@ tables! {
     /// Canonical only Stores transaction receipts.
     table Receipts<Key = TxNumber, Value = Receipt>;
 
+    /// Store snapshot id to snapshot data.
+    table Snapshots<Key = SnapshotId, Value = Snapshot>;
+
+    /// Store chunk id to chunk data.
+    table Chunks<Key = ChunkId, Value = SnapshotChunk>;
+
+    /// Stores the uncles/ommers of the block.
```
scottmillner: nit: comment needs to be updated

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -376,6 +442,276 @@ where
...
+        if snapshot.last_applied_chunk_index() >= request.index as u64 {
+            error!("Last applied chunk index is greater than the incoming chunk index");
+            return ResponseApplySnapshotChunk {
+                result: ApplySnapshotResult::RetrySnapshot as i32,
+                refetch_chunks: vec![],
+                reject_senders: vec![],
+            };
+        }
+
+        // set the last applied chunk index
```
scottmillner, line 912: how are we enforcing that chunks are added in order?
above i see we check that the `last_applied_chunk_index` is not greater than the `request.index` but we don't check that it's sequential?

for example: if last chunk is 100, shouldn't the request.index have to be 101?

crates/consensus/authority/src/snapshot_manager.rs
```diff
@@ -0,0 +1,167 @@
...
+                            last_snapshot_id,
+                            sealed_block.number,
+                            chunk.to_vec(),
+                        )?;
+                        new_chunks.push(chunk_id);
+                        client.update_snapshot(last_snapshot_id, sealed_block.number, chunk_id)?;
+                    }
+                    client.create_block_chunks_register(sealed_block.number, new_chunks)?;
+
+                    // check if we need to delete older snapshots (Retention policy)
```
scottmillner, line 359: eventually as the chain grows and we drop snapshots, this means a peer needing to sync the whole chain or from an old enough block, couldn't use state sync

i guess they could just sync using the CASS data we provide, right?

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -376,6 +442,276 @@ where
...
+            if snapshot.chunks == 0 {
+                return ResponseOfferSnapshot { result: SnapshotOfferResult::REJECT as i32 };
+            }
+
+            if snapshot.hash == prost::bytes::Bytes::default() {
+                return ResponseOfferSnapshot { result: SnapshotOfferResult::REJECT as i32 };
+            }
+
+            let latest_header =
+                client.latest_header().expect("should have latest").expect("should have header");
```
scottmillner, line 681: i have a similar comment below about applying chunks

for `snapshot.height` we only enforce that it's greater then our latest header

- don't we want to sync snapshots sequentially? or can we process them out of order? 

- or does comet always give us snapshots sequentially based on height?

what happens if:
- snapshot.height = 10 comes in
- then snapshot.height = 8
- and our latest header height is 7

crates/consensus/authority/src/snapshot_manager.rs
```diff
@@ -0,0 +1,167 @@
...
+                    })?;
+
+                    if serialized_compressed_sealed_block.is_empty() {
+                        error!(target: "consensus::authority::snapshot_manager::run", "serialized_compressed_sealed_block is empty");
+                        continue;
+                    }
+
+                    // check the block height vs. the last snapshot height
+                    let mut last_snapshot_id = match client.get_last_snapshot_height()? {
+                        Some((last_snapshot_id, last_snapshot_height)) => {
```
scottmillner, line 274: do we need to enforce we are creating snapshots in sequential order?
what if `last_snapshot_height` = 5 and `sealed.block.number` = 12

not sure how this could happen but wanted to mention

scottmillner:
did my first pass through and left some comments

wanted to say awesome work man!! 🔥 💪

crates/consensus/authority/src/comet_bft/abci.rs
rwlockbg: This is what commet is sending is:

```
pub struct Snapshot {
    /// The height at which the snapshot was taken
    #[prost(uint64, tag = "1")]
    pub height: u64,
    /// The application-specific snapshot format
    #[prost(uint32, tag = "2")]
    pub format: u32,
    /// Number of chunks in the snapshot
    #[prost(uint32, tag = "3")]
    pub chunks: u32,
    /// Arbitrary snapshot hash, equal only if identical
    #[prost(bytes = "bytes", tag = "4")]
    pub hash: ::prost::bytes::Bytes,
    /// Arbitrary application metadata
    #[prost(bytes = "bytes", tag = "5")]
    pub metadata: ::prost::bytes::Bytes,
}
```

What we can check for are height, format, chunks count, hash and metadata. I am already checking them all, metadata is something we need to define as a team and put some structure into it.  Once we have it, I would check that it follows the expected structure and values are correct.

crates/consensus/authority/src/comet_bft/abci.rs
rwlockbg: fixed

crates/consensus/authority/src/comet_bft/abci.rs

crates/storage/db/src/tables/mod.rs
rwlockbg: done, thanks.

crates/consensus/authority/src/comet_bft/abci.rs
rwlockbg: yep, I have added this condition now

crates/consensus/authority/src/snapshot_manager.rs
rwlockbg: Well, not sure - I believe this is something we need to discuss internally. I believe the state sync is usually for catching up with a couple of tausend blocks, whereas syncing from genesis smells like the node operator downloads and applies a snapshot using the CASS data. Sending gigabytes of data over comet would result in bad experience, I guarantee.  @0xBEEFCAF3 @SatwikPrabhu ideas ?

crates/consensus/authority/src/comet_bft/abci.rs
rwlockbg: I am running tests to find out how comet actually sends the data to us. Once I know better, will update here.

crates/consensus/authority/src/snapshot_manager.rs
rwlockbg: Sure. The idea behind the snapshot manager is that it receives over hte abci driver blocks in the order they are being added to the consensus by commet  meaning 1, 2, 3, ... N. I am fully relying on comet here and I believe that the moment it calls "commit" on the rpc side the commited block will be a sequential one, following the previous one. Correct me if I am mistaken here ! If that is the case the blocks will go in exactly the same order as they are commited as we send them over a channel from abci over to the snapshot manager. Also, as discussed, snapshot height = block num at which the snapshot was taken. Further down in the manager, if there is space and we add a block and all of its txs to the latest snapshot, the height of the snapshot goes up and matches the height of the sealed.block.number that was just passed over the channel.

crates/consensus/authority/src/comet_bft/abci.rs
rwlockbg: ITs not quite clear to me from the docs how it works. Building tests and int. tests to confirm the order the chunks are being passed. Nevertheless, putting a sync lock is not a bad idea as this is not a critical part of the protocol where speed and latency matter. I will add one for sure.

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -39,11 +43,43 @@ use tendermint_proto::{
...
 };
 
+/// Offer Snapshot Result
+enum SnapshotOfferResult {
+    UNKNOWN = 0,       // Unknown result, abort all snapshot restoration
+    ACCEPT = 1,        // Snapshot is accepted, start applying chunks.
+    ABORT = 2,         // Abort snapshot restoration, and don't try any other snapshots.
+    REJECT = 3,        // Reject this specific snapshot, try others.
+    REJECT_FORMAT = 4, // Reject all snapshots with this `format`, try others.
+    REJECT_SENDER = 5, // Reject all snapshots from all senders of this snapshot, try others.
```
SatwikPrabhu: Nit: Maybe `REJECT_SENDERS` is a better name here?

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -376,6 +442,287 @@ where
...
+            ),
+            Err(e) => {
+                error!("Error getting snapshots from db: {:?}", e);
+                return ResponseListSnapshots { snapshots: vec![] };
+            }
+        }
+    }
+
+    /// https://docs.cometbft.com/v0.38/spec/abci/abci++_methods#offersnapshot
+    fn offer_snapshot(&self, request: RequestOfferSnapshot) -> ResponseOfferSnapshot {
```
SatwikPrabhu, line 628: Do you think it would make sense to add info messages to distinguish between the different scenarios of the result `SnapshotOfferResult::REJECT`?

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -823,100 +1176,114 @@ where
...
-                            let mut storage = self.storage.inner.write().await;
-                            storage.aggregate_public_key = Some(edh.aggregated_public_key);
-                        }
+                        let mut storage = self.storage.inner.write().await;
+                        storage.aggregate_public_key = Some(edh.aggregated_public_key);
+                    }
+
+                    client.set_canonical_head(sealed_block_with_senders.header.clone());
+                    client.set_safe(sealed_block_with_senders.header.clone());
+                    client.set_finalized(sealed_block_with_senders.header.clone());
```
SatwikPrabhu: Can't we use `sealed_header` here instead?

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -823,100 +1176,114 @@ where
...
+                    )
+                    .await
+                    {
+                        Ok(_) => Ok(()),
+                        Err(err) => match err {
+                            EngineError => Err(EngineError),
+                            InvalidPayload => Err(InvalidPayload),
+                            _ => Err(RecvError),
+                        },
+                    };
```
SatwikPrabhu:
```suggestion
                    let _engine = engine_util::send_fork_choice_update_payload(
                        block_hash,
                        self.to_engine.clone(),
                    )
                    .await;
```

rwlockbg: Thanks, that code got copied from elsewhere. The error makes no sense as it is 1:1 mapped to the one thrown from the method. I have also kept the .await? as we dont want to fail when sending it fork choice update payload.

crates/consensus/authority/src/comet_bft/abci.rs
rwlockbg: This error struct was taken from the abci.rs library  - I have literally copy-pasted it as it was not exported. Just kept it 1:1 for compat reasons. I have no issues with renaming here, but lets maybe keep 1:1 with comet defs.

crates/consensus/authority/src/comet_bft/abci.rs
rwlockbg: Well, header is sealed header here. Its just the name that we get from reth :)

crates/consensus/authority/src/comet_bft/abci.rs
rwlockbg: I added both info! and warn! messages dep. on the serverity. Thanks for the idea!

crates/consensus/authority/src/comet_bft/abci.rs
rwlockbg: I have added a check for the block_id = snapshot_height now. We (as receivers) do not store at the point of confirming the snapshot data yet. I.e. comet would call the syncing logic if it decides we need to sync. There is a clear distinction between snapshot manager (Data and tables) whose snapshots we deliver to other users and our own data during syncing. We only start having some of that once the we agree to receiving a snapshot - there is only one table row during a sync. We cannot have multiple snapshot syncs taking place.

crates/consensus/authority/src/comet_bft/abci.rs
SatwikPrabhu: I meant that you have assigned this to the variable `sealed_header` earlier. Maybe we can just use that variable here or just remove the variable?

crates/storage/db-api/src/models/chunks.rs
```diff
@@ -0,0 +1,308 @@
...
+/// A snapshot hash is a keccak hash of a snapshot.
+pub type SnapshotChunkHash = B256;
+
+/// The storage of the a single chunk within a snapshot.
+#[derive(Debug, Default, Eq, PartialEq, Clone, Serialize, Deserialize, Compact)]
+#[cfg_attr(any(test, feature = "arbitrary"), derive(arbitrary::Arbitrary))]
+#[add_arbitrary_tests(compact)]
+pub struct BlockChunksRegister {
+    /// The block chunk ids
+    chunk_ids: Vec<u64>,
```
SatwikPrabhu: Maybe we can use Vec<ChunkId> here? I think this holds generally at various places in this file. Also, is there any reason we don't similar types for block id, block hash etc.

crates/consensus/authority/src/comet_bft/abci.rs
rwlockbg: oh ,got you. Yep, fixed that, thx.

crates/storage/db-api/src/models/chunks.rs
rwlockbg: Not following your first comment about using Vec ??? We need are separate id types eventhough they are the same since we are following the tables convention already defined by reth.

bin/test-suite/src/suite/consensus/sync/state_sync.rs
```diff
@@ -0,0 +1,162 @@
...
+        it_info_print!("#UNLOADING WALLET?", &wallet);
+        let _ = bitcoind_rpc.unload_wallet(Some(&wallet));
+    }
+    let create_res = bitcoind_rpc.create_wallet(BITCOIND_WALLET_NAME, None, None, None, None);
+    if create_res.is_err() {
+        // wallet already exists
+        // load wallet
+        let _ = bitcoind_rpc.load_wallet(BITCOIND_WALLET_NAME);
+    }
+    let _address =
```
arminsabouri: We can probably remove this

bin/test-suite/src/suite/consensus/sync/state_sync.rs
```diff
@@ -0,0 +1,162 @@
...
+    let _authority_signers = &suite
+        .local_context
+        .poa_nodes
+        .as_ref()
+        .unwrap()
+        .values()
+        .next()
+        .unwrap()
+        .authorities
+        .clone();
```
arminsabouri: Could remove this I think

bin/test-suite/src/suite/consensus/sync/state_sync.rs
```diff
@@ -0,0 +1,162 @@
...
+            .botanix_eth_client
+            .clone()
+            .expect("Botanix Client must be initialized");
+        let fed_member_pub_key = suite.local_context.authorities.get(*index as usize).unwrap();
+        let fed_member_ethereum_address = public_key_to_address(*fed_member_pub_key);
+        let fed_member_balance =
+            botanix_eth_client.get_botanix_balance(fed_member_ethereum_address).await.unwrap();
+        it_info_print!("Fed member balance", fed_member_balance);
+    }
+
```
arminsabouri: Is there more work to do for this test? To me it seems like it does some eoas then checks if everyone is on the same block height. This is not an easy test to write. What we probably want is a test util that will safely delete reset comet blocks and reth blocks. For comet we need to first ensure the node has stopped the abci server then we can run the dangerous reset command on that data dir. For Reth i wonder if we just have to run rm -rf on the db and static files. not sure if we can do better

crates/consensus/authority/src/comet_bft/abci.rs
arminsabouri: ok that makes sense. Thanks for clarifying

crates/consensus/data-parser/src/lib.rs
```diff
@@ -0,0 +1,345 @@
...
+    Json,
+}
+
+/// Traits required for a data type to be parseable
+pub trait DataParseable:
+    serde::Serialize + serde::de::DeserializeOwned + Clone + Send + Sync + Debug + std::marker::Sized
+{
+}
+
+impl<T: serde::Serialize + serde::de::DeserializeOwned + Clone + Send + Sync + Debug> DataParseable
```
arminsabouri: This is just for my rust knowledge. But why it this impl neccecary?

crates/storage/db-api/src/models/chunks.rs
SatwikPrabhu: Sorry. I meant `Vec<ChunkId>` instead of `Vec<u64>` since we already have that type defined.

rwlockbg: yes, but the mbdx macros wont allow that as they need a clearly defined type and not a type alias. I already tried that until I realized that they dont support that.

bin/test-suite/src/suite/consensus/sync/state_sync.rs
rwlockbg: The test is currently a copy -paste of the builder test. Please dont look at it yet as its far from ready. The difficulty I am facing is to dynamically generate a poa member after a few snapshots are generated by the remaining N-1 poa members. I am still re-thinking the concept here, and yes, it is a really hard test to write.

bin/test-suite/src/suite/consensus/sync/state_sync.rs
rwlockbg: please see below.

bin/test-suite/src/suite/consensus/sync/state_sync.rs
rwlockbg: please see below.

crates/consensus/authority/src/snapshot_manager.rs
scottmillner: makes sense, thanks!

arminsabouri:
@rwlockbg can we start reviewing this again?

rwlockbg:
> @rwlockbg can we start reviewing this again?

Yes

crates/consensus/authority/src/comet_bft/abci.rs
SatwikPrabhu: Sorry for being nitpicky here. I think you have just fixed this for `set_finalized()`. Same can be done for `set_canonical_head()` and `set_safe()` as well :)

bin/test-suite/src/suite/consensus/sync/test_state_sync.rs
```diff
@@ -0,0 +1,124 @@
...
+            );
+            let db_provider = suite
+                .local_context
+                .get_dbs()
+                .get(canon_state_notification.engine_index as usize)
+                .cloned()
+                .unwrap();
+            let snapshots = db_provider.get_snapshots().unwrap_or_default();
+            // NOTE: it means that the second snapshot is being created at the moment, hence prev.
+            // one must be ready
```
scottmillner: nit: add a short description of what this code block is doing? it's not hard to figure out but description could be nice

bin/test-suite/src/suite/consensus/sync/test_state_sync_dynamic.rs
```diff
@@ -0,0 +1,183 @@
...
+                // spawn the comet process
+                spawned_cometbft_processes.push(comet_node.spawn_service().unwrap());
+                // await initialization
+                comet_node.await_initialization().unwrap();
+            }
+        }
+    }
+
+    // get the syncing node db
+    let db_provider_syncing_member =
```
scottmillner: this test assumes there is just one syncing node correct? above in `saturating_sub(1)` as well
nit: small comment stating this

bin/test-suite/src/suite/consensus/sync/test_state_sync_dynamic.rs
```diff
@@ -0,0 +1,183 @@
...
+        common::{
+            comet_node::update_config_toml_with_trusted_height_and_hash,
+            events::{BITCOIND_WALLET_NAME, SEND_AMOUNT},
+        },
+        ConsensusIntegrationTestSuite,
+    },
+    utils::generate_blocks,
+};
+
+#[allow(clippy::too_many_lines)]
```
scottmillner: cool test!! i like it 😄

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -405,6 +590,393 @@ where
...
+                // filter out the snapshot that is the same as the current block as we might not be
+                // ready having all the chunks yet
+                let resp = snapshots
+                    .into_iter()
+                    .filter(|s| s.height() != snapshot_manager_state_lock.get_block_id())
+                    .fold(ResponseListSnapshots { snapshots: vec![] }, |mut acc, snapshot| {
+                        acc.snapshots.push(Snapshot {
+                            height: snapshot.height(),
+                            format: SNAPSHOT_MESSAGE_FORMAT,
+                            chunks: snapshot.chunk_ids().len() as u32,
```
scottmillner, line 607: nit: remove comment

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -405,6 +590,393 @@ where
...
+            warn!("Received snapshot height matches current block height, rejecting snapshot as it might not be ready yet");
+            return ResponseLoadSnapshotChunk::default();
+        }
+
+        match client.get_snapshot_id_by_block_id(request.height) {
+            Ok(Some(snapshot_id)) => {
+                // now take the entire snapshot data
+                match client.get_snapshot_by_id(snapshot_id) {
+                    Ok(Some(snapshot)) => {
+                        let mapped_request_chunk = request.chunk + 1;
```
scottmillner, line 764: little confused by this comment?
if cometbft starts at 0, why do the chunks start at 1? is this the way MDBX works?

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -405,6 +590,393 @@ where
...
+        let sealed_block_with_senders = compressor_task_rx
+            .blocking_recv()
+            .expect("to receive confirmation from snapshot manager");
+
+        let sealed_header = sealed_block_with_senders.header.clone();
+        let block_hash = sealed_header.hash();
+        info!("Restored Snapshot Hash {:?}", hex::encode(block_hash.to_vec().as_slice()));
+
+        // Update canonical chain
+        if let Err(e) =
```
scottmillner: just a heads up that in pr: https://github.com/botanix-labs/Macbeth/pull/545
the blockchain tree engine has been removed so can't use `insert_block` here but we can hop on a call to help rebase when the time comes

or we can merge this in first whatever makes the most sense
@0xBEEFCAF3 thoughts?

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -866,93 +1444,113 @@ where
...
+                        {
+                            error!("Error sending finalize message to driver: {:?}", e);
                         }
-
-                        tx.send(()).expect("to send");
-                    }
-                    ABCIDriverMessage::Exit => {
-                        break;
-                    }
+                    }));
```
scottmillner: im guessing this is commented out bc we don't want to prevent inserting blocks if creating a snapshot fails?
but if snapshot creation fails, do we have any way to retry creating it?

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -1069,12 +1669,14 @@ mod tests {
             nonce: 0,
         };
         let bitcoin_checkpoint: BitcoinCheckpoint =
-            Arc::new(RwLock::new(Some((bitcoin_header, 0))));
```
scottmillner, line 1070: i think `TokioRwLock` should be your new slack handle 😆

crates/consensus/authority/src/comet_bft/light_client.rs
```diff
@@ -49,7 +49,7 @@ impl LightCBFTClientBuilder {
             .header
             .hash();
 
-        let light_client = LightClientBuilder::prod(
```
scottmillner, line 52: are we using the light client anywhere or just an http provider?

crates/consensus/authority/src/snapshot_manager.rs
```diff
@@ -0,0 +1,579 @@
...
+        while let Some(abci_driver_message) = self.snapshot_manager_tx.recv().await {
+            debug!(target: "consensus::authority::snapshot_manager::run", "received abci driver message {:?}", abci_driver_message);
+
+            match abci_driver_message {
+                ABCIDriverMessage::CommitBlock((sealed_block_with_peg, cbft_hash, _tx)) => {
+                    // acknowledge the block
+                    //tx.send(()).expect("acknowledging received block send");
+
+                    let sealed_block = sealed_block_with_peg.block();
+
```
scottmillner, line 256: shouldn't this comment say the opposite?
we're serializing not deserialzing?

rwlockbg: yep, exactly. Thanks, I have changed it now.

crates/consensus/authority/src/comet_bft/light_client.rs
rwlockbg: We are not using it Scott. I am using the http provider instead due to the issues I mentioned to you.

crates/consensus/authority/src/comet_bft/abci.rs
rwlockbg: :) agreed

crates/consensus/authority/src/comet_bft/abci.rs
rwlockbg: Nah, this was commented out as we cannot have blocking here - the runtime crashes if we allow blocking_Recv a this point. I have removed it now.

crates/consensus/authority/src/comet_bft/abci.rs
rwlockbg: I will wait for 545 then.

crates/consensus/authority/src/comet_bft/abci.rs
rwlockbg: I have fixed the comment and made it more self-explanatory.

crates/consensus/authority/src/comet_bft/abci.rs
rwlockbg: done

bin/test-suite/src/suite/consensus/sync/test_state_sync_dynamic.rs
rwlockbg: thx

bin/test-suite/src/suite/consensus/sync/test_state_sync_dynamic.rs
rwlockbg: yes, in the setup in the Makefile we set syncing-nodes = 1 https://github.com/botanix-labs/Macbeth/blob/feat/rwlock/state-sync/Makefile#L559

bin/test-suite/src/suite/consensus/sync/test_state_sync.rs
rwlockbg: done

crates/consensus/authority/src/comet_bft/abci.rs

Makefile
```diff
@@ -639,6 +655,8 @@ start-poa-server-1:
 	--bitcoind.password "${BITCOIND_PWD}" \
 	--frost.min_signers 2 \
 	--frost.max_signers 2 \
+	--sync.max_snapshot_size_bytes 8388608 \
```
arminsabouri: Why this number?

Makefile
```diff
@@ -639,6 +655,8 @@ start-poa-server-1:
...
+	--sync.snapshot_keep_recent 3 \
```
arminsabouri: Keep recent what? chunks?

bin/test-suite/src/context/mod.rs
```diff
@@ -47,10 +50,13 @@ impl GlobalContext {
             dry_run: args.dry_run,
             fed_instances,
             rpc_instances: args.rpc_nodes,
+            syncing_instances: args.syncing_nodes,
```
arminsabouri, line 52: Are some of the federation nodes not able to serve syncing demands?

arminsabouri: Why do we need this distinction

bin/test-suite/src/suite/consensus/common/comet_node.rs
```diff
@@ -303,6 +306,44 @@ fn update_config_toml(cometbft_node: &CometBftNodeConfig) -> anyhow::Result<()>
...
+                    .collect::<Vec<String>>()
+                    .join(",");
+                *rpc_servers = toml::value::Value::String(rpc_state_sync_servers);
+            }
+            if let Some(trust_height) = rpc.get_mut("trust_height") {
+                *trust_height = toml::value::Value::Integer(1);
+            }
+            if let Some(trust_hash) = rpc.get_mut("trust_hash") {
+                *trust_hash = toml::value::Value::String(
+                    "7b3ca33b44aee2b296253fa69e1b2b74789655a8aacb7d53a29c397cc8a9b379".to_string(),
```
arminsabouri, line 337: Is this always the hash of the first block (height=1)?

bin/test-suite/src/suite/consensus/sync/test_state_sync.rs
```diff
@@ -0,0 +1,124 @@
...
+        let _ = bitcoind_rpc.unload_wallet(Some(&wallet));
+    }
+    let create_res = bitcoind_rpc.create_wallet(BITCOIND_WALLET_NAME, None, None, None, None);
+    if create_res.is_err() {
+        // wallet already exists
+        // load wallet
+        let _ = bitcoind_rpc.load_wallet(BITCOIND_WALLET_NAME);
+    }
+    let _address =
+        bitcoind_rpc.get_new_address(None, None).expect("get new address").assume_checked();
```
arminsabouri: Not used. please remove

bin/test-suite/src/suite/consensus/sync/test_state_sync.rs
```diff
@@ -0,0 +1,124 @@
...
+
+    let test_fed_members = suite
+        .local_context
+        .poa_nodes
+        .as_ref()
+        .expect("test federation member configurations")
+        .clone();
+    let mut rx = suite.local_context.poa_notification.as_ref().expect("poa notifs").subscribe();
+
+    // take the first member as the inturn member
```
arminsabouri: Inturn is not really a concept that we use.  I would remove this comment

bin/test-suite/src/suite/consensus/sync/test_state_sync.rs
```diff
@@ -0,0 +1,124 @@
...
+            let db_provider = suite
+                .local_context
+                .get_dbs()
+                .get(canon_state_notification.engine_index as usize)
+                .cloned()
+                .unwrap();
+            let snapshots = db_provider.get_snapshots().unwrap_or_default();
+            // NOTE: at these point we should have 2 snapshots in the db, the first one being
+            // finalized and the second one being in progress
+            if snapshots.len() == 2 {
```
arminsabouri: Should we check all members have the same snapshots at this point?

bin/test-suite/src/suite/consensus/sync/test_state_sync_dynamic.rs
```diff
@@ -0,0 +1,183 @@
...
+        let _ = bitcoind_rpc.unload_wallet(Some(&wallet));
+    }
+    let create_res = bitcoind_rpc.create_wallet(BITCOIND_WALLET_NAME, None, None, None, None);
+    if create_res.is_err() {
+        // wallet already exists
+        // load wallet
+        let _ = bitcoind_rpc.load_wallet(BITCOIND_WALLET_NAME);
+    }
+    let _address =
+        bitcoind_rpc.get_new_address(None, None).expect("get new address").assume_checked();
```
arminsabouri: Not used remove

bin/test-suite/src/suite/consensus/sync/test_state_sync_dynamic.rs
```diff
@@ -0,0 +1,183 @@
...
+    tokio::time::sleep(Duration::from_secs(5)).await;
+
+    let test_fed_members = suite
+        .local_context
+        .poa_nodes
+        .as_ref()
+        .expect("test federation member configurations")
+        .clone();
+
+    // take the first member as the inturn member
```
arminsabouri: same comment about the comment here

bin/test-suite/src/suite/consensus/sync/test_state_sync_dynamic.rs
```diff
@@ -0,0 +1,183 @@
...
+
+    // wait for the syncing node to catch up with expected_sync_height
+    loop {
+        let last_block_number = db_provider_syncing_member.last_block_number().unwrap();
+        it_info_print!("Syncing last block number {:?}", last_block_number);
+        if last_block_number >= expected_sync_height {
+            return Ok(());
+        }
+        tokio::time::sleep(Duration::from_secs(5)).await;
+    }
```
arminsabouri: Awesome test! Can we merge the two tests above into one?

bin/test-suite/src/suite/consensus/sync/test_state_sync_dynamic.rs
```diff
@@ -0,0 +1,183 @@
...
+                comet_node.await_initialization().unwrap();
+            }
+        }
+    }
+
+    // get the syncing node db
+    let db_provider_syncing_member =
+        suite.local_context.get_dbs().get(member_ids.len()).cloned().unwrap();
+
+    // wait for the syncing node to catch up with expected_sync_height
```
arminsabouri: Should we check that after the node has sync'd that they also have the same snapshots? Does it make sense to check this?

crates/btc-wallet/src/psbt.rs
```diff
@@ -246,9 +246,9 @@ pub enum PsbtToSigningPackageConversionError {
 }
 
 pub fn frost_id_from_bytes(b: &[u8]) -> Option<frost::Identifier> {
-    frost::Identifier::deserialize(&b).ok()
```
arminsabouri: This change will conflict with #552 . In general I dont think this PR should touch btc server

crates/consensus/authority/src/builder.rs
```diff
@@ -305,6 +324,12 @@ where
             frost_task = Some(task);
         }
 
+        let (snapshot_manager_tx, snapshot_manager_rx) =
+            tokio::sync::mpsc::channel::<ABCIDriverMessage>(100);
```
arminsabouri: I would pass canon notif reciever to the snapshot manager not the commit. Simply b/c there is a small chance that commit will error out and we will mistakenly save a block that was never committed

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -72,7 +112,77 @@ const VERIFY_REJECT: i32 = 2;
...
+        self.snapshot_chunks
+    }
+
+    /// Get snapshot format
+    pub fn get_snapshot_format(&self) -> u64 {
+        self.snapshot_format
+    }
+
+    /// checks if the snapshot is identical with the given props
+    pub fn is_identical(
```
arminsabouri: nit: can you impl PartialEq or Eq instead of this method?

arminsabouri: or even derive

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -883,93 +1466,110 @@ where
...
+
+                    let sealed_block_with_peg = sealed_block_with_peg.clone();
+                    let (snapshot_manager_announced_tx, _snapshot_manager_announced_rx) =
+                        tokio::sync::oneshot::channel::<()>();
+                    let snapshot_manager_tx = self.snapshot_manager_tx.clone();
+                    self.task_executor.spawn(Box::pin(async move {
+                        if let Err(e) = snapshot_manager_tx
+                            .send(ABCIDriverMessage::CommitBlock((
+                                sealed_block_with_peg,
+                                cbft_hash,
```
arminsabouri: Same comment here about using the canon subscriber instead of a new queue. The canon state notifications are fired off when commit is successful. Ideally the snapshot manager can just subscribe to the same queue

arminsabouri (commented):
I am 37 / 68 files deep. Leaving a review for now

Makefile
rwlockbg: This is equivalent to 8 Mbs in bytes. Its just a hard limit we set here, but we need ot check this in practice in terms of network performance.

crates/consensus/authority/src/comet_bft/abci.rs
rwlockbg: agree, but that is subject to merging Scotts pr first right ?

crates/btc-wallet/src/psbt.rs
rwlockbg: resolved.

bin/test-suite/src/suite/consensus/common/comet_node.rs
rwlockbg: Nope, this is a random hash which gets overwritten with the correct one during the test runtime, before the commet node is spawned.

rwlockbg: block = 1 has a diff hash on each run. Only height = 0 has the same, but we are not lallowed to have block h = 0 in the config.

crates/consensus/authority/src/builder.rs
rwlockbg: I agree, but lets add this after Scotts per gets merged as I fear there will be conflicts either way

bin/test-suite/src/suite/consensus/sync/test_state_sync.rs
rwlockbg: I have renamed it here and all the other tests having the same variable and description

bin/test-suite/src/suite/consensus/sync/test_state_sync_dynamic.rs
rwlockbg: removed

bin/test-suite/src/suite/consensus/sync/test_state_sync_dynamic.rs
rwlockbg: changed, see above

crates/consensus/authority/src/comet_bft/abci.rs

bin/test-suite/src/suite/consensus/sync/test_state_sync.rs
rwlockbg: removed

bin/test-suite/src/suite/consensus/common/comet_node.rs
rwlockbg: I have left a comment there for info.

Makefile
rwlockbg: In the comet docs they recommend up to 10 mbs.

bin/test-suite/src/context/mod.rs
rwlockbg: Because we need to know how many of them fed members will be syncing. Why ? Because the cometbt nodes are not being prepared and run until we decide so during the test execution  unlike the standard fed members which are being pre-created alongside with their cometbft nodes. I did not find a better way of doing it as we need to have this dynamic nature for the state sync.

rwlockbg: And yes, as I have mentioned before, if the block_height > 0 then the sync shall start. When all nodes start simultaneously as now there is no sync. The idea of the syncing)instances is to pre-create them and allow us to actually start them when we want state sync to be triggered.SO, in other words, these instances are not serving syncing demands when they are first being created but only rather when they are being run and onwards.

bin/test-suite/src/suite/consensus/sync/test_state_sync_dynamic.rs
rwlockbg: Not really, as the syncing node's commit rpc endpoint wont be called by the cometbft network during syncing (this is how comet works). Because of this the snapshot builder wont be receiving any messages during state sync and wont have any snapshots persisted. Once state sync is done and the cometbft mechanism kicks in (commit rpc being called, abci driver messages forwarding to snapshot manager) then we will start building snapshots.

Makefile
rwlockbg: keep recent snapshots. I have renamed it to `num_snapshots_to_keep` everywhere

bin/test-suite/src/suite/consensus/sync/test_state_sync_dynamic.rs
arminsabouri: @rwlockbg following up on this

bin/btc-server/src/database/error.rs
```diff
@@ -1,7 +1,7 @@
-use thiserror::Error;
+use bitcoin::psbt;
 use frost_secp256k1_tr as frost;
 use std::{array::TryFromSliceError, io};
-use bitcoin::psbt;
+use thiserror::Error;
```
arminsabouri: In the future please keep these cosmetic changes in a different commit or differnt PR entirely

crates/storage/provider/src/traits/snapshot.rs
```diff
@@ -0,0 +1,132 @@
...
+use reth_storage_errors::provider::ProviderResult;
+use std::ops::RangeInclusive;
+
+/// SnapshotReader
+#[auto_impl::auto_impl(&, Arc, Box)]
+pub trait SnapshotReader: Send + Sync {
+    /// Get snapshots
+    fn get_snapshots(&self) -> ProviderResult<Vec<Snapshot>>;
+
+    /// Get snapshot by id
```
arminsabouri, line 12: Wrong doc string for method

crates/storage/provider/src/traits/snapshot.rs
```diff
@@ -0,0 +1,132 @@
...
+    ) -> ProviderResult<Vec<(u64, Vec<u8>)>>;
+
+    /// Get snapshot by id
+    fn get_snapshot_by_id(&self, snapshot_id: SnapshotId) -> ProviderResult<Option<Snapshot>>;
+
+    /// Get last snapshot sync by id
+    fn get_last_snapshot_sync_id(&self) -> ProviderResult<Option<SnapshotSyncId>>;
+
+    /// Get snapshot sync by height
+    fn get_snapshot_sync_by_height(&self, height: u64) -> ProviderResult<Option<SnapshotSync>>;
```
arminsabouri, line 19: What is the difference beteween snapshot sync and snapshot itself?

crates/storage/provider/src/traits/snapshot.rs
```diff
@@ -0,0 +1,132 @@
...
+    fn get_snapshot_sync_by_id(&self, id: u64) -> ProviderResult<Option<SnapshotSync>>;
+
+    /// Get chunk by chunk id
+    fn get_chunk_by_id(
+        &self,
+        chunk_id: reth_db::models::ChunkId,
+    ) -> ProviderResult<Option<SnapshotChunk>>;
+
+    /// Get snapshot id by block id
+    fn get_snapshot_id_by_block_id(
```
arminsabouri, line 34: In someplaces in this trait we use height other places we use block number / block id.
I would just unify this terminology and use height. And rename this to `get_snapshot_id_by_height`

crates/storage/provider/src/traits/snapshot.rs
```diff
@@ -0,0 +1,132 @@
...
+    fn remove_snapshots(&self, range: RangeInclusive<SnapshotId>) -> ProviderResult<()>;
+
+    /// Removes oldest snapshot
+    fn remove_oldest_snapshot(&self) -> ProviderResult<()>;
+
+    /// Removes snapshots
+    fn remove_chunks(&self, range: RangeInclusive<ChunkId>) -> ProviderResult<()>;
+
+    /// Deletes chunks in blocks
+    fn delete_chunks_in_blocks(&self, range: RangeInclusive<ChunkId>) -> ProviderResult<()>;
```
arminsabouri, line 128: How are these two different?

crates/storage/provider/src/traits/snapshot.rs
```diff
@@ -0,0 +1,132 @@
...
+        format: u64,
+    ) -> ProviderResult<SnapshotId>;
+
+    /// Create new snapshot
+    fn create_new_snapshot(
+        &self,
+        block_id: BlockNumber,
+        block_hash: B256,
+        app_hash: &[u8],
+    ) -> ProviderResult<SnapshotId>;
```
arminsabouri: What exactly is the differnce between these? And is there a better name for snapshotsync that would be more descriptive on what we are actually creating. B/c when we say `create_new_snapshot` I can conceptualize what data is being saved or created on the db. But i cannot say the same thing for snapshot_sync

crates/consensus/authority/src/builder.rs
scottmillner: the way this works now:
- in driver `start` after inserting the block, the `CommitBlock` is sent to the snapshot manager. And the `CommitBlock` message has additional fields needed by the manager that the `canon notif receiver` will not have.

crates/consensus/authority/src/comet_bft/abci.rs
scottmillner: please see comment above

crates/consensus/authority/src/comet_bft/abci.rs
scottmillner: this can be any arbitrary hash and at the moment we only check that it's not default bytes
so i guess the value should be whatever might be useful in the future

we're setting it as the comet `app_hash` at the `snapshot.height` which is the final block height contained in the snapshot

maybe that's useful maybe not...i'd say leave a todo to revisit since we're not really using the field

crates/consensus/authority/src/comet_bft/abci.rs
scottmillner: no based on the docs: `The chunk index, starting from 0. CometBFT applies chunks sequentially.`

and have confirmed this when running int test

bin/test-suite/src/suite/consensus/sync/test_state_sync_dynamic.rs
scottmillner: done!

bin/btc-server/src/wallet/psbt.rs
```diff
@@ -248,11 +248,11 @@ pub enum PsbtToSigningPackageConversionError {
 }
 
 pub fn frost_id_from_bytes(b: &[u8]) -> Option<frost::Identifier> {
-    frost::Identifier::deserialize(&b).ok()
+    frost::Identifier::deserialize(b).ok()
 }
 
 pub fn signature_share_from_bytes(b: &[u8]) -> Option<frost::round2::SignatureShare> {
-    frost::round2::SignatureShare::deserialize(b.try_into().ok()?).ok()
+    frost::round2::SignatureShare::deserialize(b).ok()
```
arminsabouri, line 255: Is this fixing a warning? I generally agree with this change but why is it needed as a part of state sync?

bin/reth/src/commands/poa/mod.rs
```diff
@@ -553,11 +559,15 @@ impl<Ext: clap::Args + fmt::Debug> PoaNodeCommand<Ext> {
         };
 
         let (driver_tx, driver_rx) = tokio::sync::mpsc::channel(1);
+        let (snapshot_manager_tx, snapshot_manager_rx) =
+            tokio::sync::mpsc::channel::<ABCIDriverMessage>(100);
```
arminsabouri: I gave this feedback somewhere else in this PR. But why can't we just use the canon state notifications?

arminsabouri: A bit strange also that this queue can take up to 100 message but the one driver can only take 1? does that make sense?

bin/test-suite/src/suite/consensus/common/btc_server.rs
```diff
@@ -67,7 +67,11 @@ fn spawn_btc_server_process(
     let address = format!("0.0.0.0:{}", port);
     let http_port = (BTC_SERVER_HTTP_PORT + id).to_string();
 
-    let command = "./target/debug/btc-server";
+    let command = "target/debug/btc-server";
+    let binary_abs_path = working_directory.join(Path::new(command));
+    if !std::fs::exists(&binary_abs_path)? {
+        return Err(anyhow::anyhow!("btc-server binary not found at {}. Please compile it first before running the test-suite", binary_abs_path.display().to_string()));
+    }
```
arminsabouri, line 74: Ack this change but in the future I will be more anal about unrelated improvments being merged in main first.

bin/test-suite/src/suite/consensus/frost/test_utxo_sync.rs
```diff
@@ -84,7 +84,11 @@ pub async fn utxo_sync(
         .unwrap();
     it_info_print!("Eoa tx receipt hash: {:?}", tx_receipt.transaction_hash);
 
-    let poa_eth_clients = suite.local_context.poa_eth_providers.clone().unwrap();
+    let mut poa_eth_clients = suite.local_context.poa_eth_providers.clone().unwrap();
+    // remove the syncing clients
+    for _syncing_node in 0..suite.global_context.syncing_instances as usize {
+        poa_eth_clients.pop().unwrap();
+    }
```
arminsabouri, line 91: Nit: this operation can just be a method on `local_context()`.
```
suite.local_context.get_eth_providers()

```

This way each individual test doesnt need to remove syncing nodes

bin/test-suite/src/suite/consensus/sync/test_state_sync.rs
arminsabouri: @rwlockbg @scottmillner Following up on this

crates/consensus/authority/src/prost_parser.rs
```diff
@@ -0,0 +1,179 @@
+use bytes::Bytes;
+use displaydoc::Display as DisplayDoc;
+use thiserror::Error;
+
+/// Password hashing error types.
```
arminsabouri: Password Hashing?

crates/consensus/authority/src/prost_parser.rs
```diff
@@ -0,0 +1,179 @@
...
+    ProstEncode(#[from] prost::EncodeError),
+    /// serde prost decode error
+    ProstDecode(#[from] prost::DecodeError),
+}
+
+/// Prost Message Wrapper allowing serialization/deserialization
+#[allow(dead_code)]
+pub(crate) struct ProstMessageSerdelizer<T: prost::Message>(pub(crate) T);
+
+#[allow(dead_code)]
```
arminsabouri, line 18: Where do we expect to use this?

crates/primitives/src/botanix/peg_contract.rs
```diff
@@ -127,11 +132,13 @@ impl PeginData {
 }
 
 /// Pegin metadata structure
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize)]
```
arminsabouri: I dont understand why these `Deserialize, Serialize` changes are included? Is there a strong reason for state sync? If not please revert these changes.

crates/primitives/src/botanix/utils.rs
```diff
@@ -47,6 +51,48 @@ pub trait AmountExt: Copy + From<bitcoin::Amount> + Into<bitcoin::Amount> {
...
+    serializer.serialize_str(&address.to_string())
+}
+
+/// Custom deserializer for bitcoin::Address
+pub fn deserialize_bitcoin_address<'de, D>(deserializer: D) -> Result<bitcoin::Address, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let s = String::deserialize(deserializer)?;
+    Ok(bitcoin::Address::from_str(&s).map_err(serde::de::Error::custom)?.assume_checked())
```
arminsabouri: Dangerous use of `assumed_checked()` here. Can you justify why? 
Esp since I see this being used in peg_contract where we check for the validity of an address with respect to the bitcoin network

crates/storage/db/src/tables/mod.rs
```diff
@@ -307,6 +311,24 @@ tables! {
...
+    /// Stores block snapshots.
+    table BlockSnapshots<Key = BlockNumber, Value = SnapshotId>;
+
+    /// Stores the chunk to Block ids
+    table ChunkBlocks<Key = ChunkId, Value = BlockNumber>;
+
+    /// Stores the block number to chunk ids
+    table BlockChunks<Key = BlockNumber, Value = BlockChunksRegister>;
+
+    /// Table used when syncing snapshots.
```
arminsabouri, line 323: Nit: The rest of the tables are self-descriptive. Snapshot sync im not sure i understand. Can you provide some more detail what this table is responsible for

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -3688,6 +3693,312 @@ impl<TX: DbTxMut> FinalizedBlockWriter for DatabaseProvider<TX> {
...
+    fn get_snapshot_by_id(
+        &self,
+        snapshot_id: reth_db::models::SnapshotId,
+    ) -> ProviderResult<Option<reth_db::models::Snapshot>> {
+        Ok(self.tx.get::<tables::Snapshots>(snapshot_id)?)
+    }
+
+    fn get_snapshot_id_by_block_id(
+        &self,
+        block_id: BlockNumber,
```
arminsabouri, line 3740: Nit: block_id means a specific thing in the reth world. Lets rename to block_number

crates/storage/provider/src/traits/snapshot.rs
arminsabouri: remove_chunks & delete_chunks_in_blocks

crates/storage/db/src/tables/mod.rs
```diff
@@ -307,6 +311,24 @@ tables! {
     /// Canonical only Stores transaction receipts.
     table Receipts<Key = TxNumber, Value = Receipt>;
 
+    /// Store snapshot id to snapshot data.
+    table Snapshots<Key = SnapshotId, Value = Snapshot>;
+
+    /// Store chunk id to chunk data.
+    table Chunks<Key = ChunkId, Value = SnapshotChunk>;
+
+    /// Stores block snapshots.
```
arminsabouri:
```suggestion
    /// Association table linking block number to snapshot id 
```

crates/storage/db/src/tables/mod.rs
```diff
@@ -307,6 +311,24 @@ tables! {
...
+    table BlockSnapshots<Key = BlockNumber, Value = SnapshotId>;
+
+    /// Stores the chunk to Block ids
+    table ChunkBlocks<Key = ChunkId, Value = BlockNumber>;
+
+    /// Stores the block number to chunk ids
```
arminsabouri: is a `BlockChunksRegister` the same as a chunk id?

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -3688,6 +3693,312 @@ impl<TX: DbTxMut> FinalizedBlockWriter for DatabaseProvider<TX> {
...
+                    chunk_ids.first().cloned().unwrap_or_default()..=
+                        chunk_ids.last().cloned().unwrap_or_default(),
+                )?
+                .collect::<Result<HashMap<_, _>, _>>()?
+                .values()
+                .map(|value| value.size())
+                .sum()
+        };
+
+        Ok(snapshot_size + chunks_size)
```
arminsabouri, line 3807: Just double checking here. Isnt a snapshot consisting of a many chunks? So why are we not just returning `chunks_size`?

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -3688,6 +3693,312 @@ impl<TX: DbTxMut> FinalizedBlockWriter for DatabaseProvider<TX> {
...
+    fn assemble_snapshot_chunks_data(
+        &self,
+        snapshot_id: SnapshotId,
+    ) -> ProviderResult<Vec<(u64, Vec<u8>)>> {
+        let snapshot = self.get_snapshot_by_id(snapshot_id)?;
+        if snapshot.is_none() {
+            return Ok(Vec::new());
+        }
+
+        let block_ids = snapshot.unwrap().block_ids().to_vec();
```
arminsabouri:
```suggestion
        let block_ids = snapshot.expect("checked above").block_ids().to_vec();
```

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -3688,6 +3693,312 @@ impl<TX: DbTxMut> FinalizedBlockWriter for DatabaseProvider<TX> {
...
+                .values()
+                .map(|chunk| chunk.chunk_data().to_vec())
+                .collect::<Vec<_>>();
+
+            let mut concatenated = Vec::new();
+            for chunk_data in chunks_data {
+                concatenated.extend(chunk_data);
+            }
+
+            blocks_data.push((block, concatenated));
```
arminsabouri: Just want to understand? A block can be in many chunks? Is it possible for a many blocks to be in one chunk?

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -129,17 +249,21 @@ where
...
-    /// Start the ABCI server
+    /// Starts the abci client server
     pub async fn start_server<Pool: TransactionPool + Clone + 'static>(
         &self,
         task_executor: &impl TaskSpawner,
         validator: EthTransactionValidator<DB, EthPooledTransaction>,
         tx_pool: Pool,
         abci_host: String,
         abci_port: u16,
+        block_chain_provider_2: BlockchainProvider2<Arc<DatabaseEnv>>,
```
arminsabouri: Nit: `_2` is unneccecary

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -399,6 +584,404 @@ where
...
+                            height: snapshot.height(),
+                            format: SNAPSHOT_MESSAGE_FORMAT,
+                            chunks: snapshot.chunk_ids().len() as u32,
+                            hash: snapshot.get_app_hash().to_vec().into(),
+                            metadata: prost::bytes::Bytes::new(),
+                        });
+                        acc
+                    });
+                drop(snapshot_manager_state_lock);
+                info!("Returned snapshots {:?}", resp);
```
arminsabouri: Will this be a large print? If so can we just print the id and the number of chunks returned?

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -399,6 +584,404 @@ where
...
+
+        let executed_block = ExecutedBlock::new(
+            Arc::new(sealed_block_with_senders.block.clone()),
+            Arc::new(sealed_block_with_senders.senders.clone()),
+            Arc::new(sealed_block_with_context.exec_outcome.clone()),
+            Arc::new(hashed_state.clone()),
+            Arc::new(trie_updates.clone()),
+        );
+
+        let db_rw = self.provider_factory.provider_rw().unwrap();
```
arminsabouri: handle `unwrap()`

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -399,6 +584,404 @@ where
...
+                reject_senders: vec![],
+            };
+        }
+
+        // decompress and decode the snapshot chunk (= block) and apply it
+        let compressor = self.compressor.clone();
+        let (compressor_task_tx, compressor_task_rx) =
+            tokio::sync::oneshot::channel::<BlockWithContext>();
+        self.task_executor.spawn_blocking(Box::pin(async move {
+            let block_with_context: BlockWithContext = compressor
```
arminsabouri: Is each individual chunk a block with context? If so im not sure how this can be better than block sync. We would have to make a back and forth for each block in snapshot?

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -399,6 +584,404 @@ where
...
+        // and the methods used here are directly implemented on the block_provider_2 and not part
+        // of a trait
+        self.block_chain_provider_2.canonical_in_memory_state().update_chain(new_chain);
+
+        client.on_forkchoice_update_received(&ForkchoiceState::default());
+
+        // set canonical head
+        client.set_canonical_head(new_header.clone());
+        client.set_safe(new_header.clone());
+        client.set_finalized(new_header);
```
arminsabouri: There is a lot of duplicated code here. I will ask that we consolidate this as much as we can

crates/consensus/authority/src/comet_bft/abci.rs
arminsabouri: So is the lock redundant then?

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -1011,14 +1611,32 @@ where
                                 error!("Error notifying pegouts: {:?}", e);
                             }
                         }
-                        commit_tx.send(()).unwrap();
+                        commit_tx.send(()).expect("to send");
+
+                        // Send message to snapshot manager
+                        let (snapshot_manager_announced_tx, _snapshot_manager_announced_rx) =
```
arminsabouri: Why is the mpsc required? Im not seeing the rx side used?

arminsabouri (changes requested):
This is alot to review. The biggest things as of now that give me pause in order of urgency are
1. Seemingly unrelated consensus changes in peg_contract. There were some ser/deser utils added. If these utils are related lmk and I can review this more carefully other wise please remove this and consider a seperate PR with test coverage
1. Each chunk appears to be a block. The main benefit of state sync over block sync is that we can save round of communication however that is completly negated if we need to do 2 rounds of comms for each block. It would make sense to me if we can fit many blocks into a 10MB chunks. 
1. Duplicated code in abci.rs. We've spent a while in Jan removing techdebt and I dont think it makes sense for new experimental features to add tech debt 
1. I have some questions about the use of `BlockWithContext`. Generally it doesnt make alot of sense to me that each chunk is a rich block with excecution context. Perhaps we need this so we can `insert_block_with_data` b/c we need the historical reciepts. However,  each chunk being a block seems to negate some of the benefits of state sync. Mainly we need to do a full 2 rounds of comms for each block (as is the case with block sync). It would make alot of sense to concat blocks into many chunks until we hit the max chunk size. 
1. Lower priority but I'd like to see some docs around the new tables. How they relate to each other and how they intended to be used. 
1. I imagine in prod we are going to want to reset the snapshot manager / snapshot tables if we need to. Do we have a plan for this? 

I undestand this is a lot of review but this is a lot of code. Thanks for your persistance on this PR. I greatly appreciate the patience

crates/storage/provider/src/traits/snapshot.rs
scottmillner: `SnapshotSync` is a struct that is used while syncing and has metadata like `total_chunks: u64` and `last_applied_chunk_index: u64`

- this is used to enforce chunks are applied in sequential order

`Snapshot` is the final result after syncing all the chunks of a snapshot

crates/storage/provider/src/traits/snapshot.rs
scottmillner: these methods use two different tables

`Chunks` table which is a mapping of `ChunkId` and the chunk data itself
- `remove_chunks` updates this table

` ChunkBlocks` is a mapping of `ChunkId` to `BlockNumber`
- ` delete_chunks_in_blocks` updates this table

crates/storage/provider/src/traits/snapshot.rs
scottmillner: done in https://github.com/botanix-labs/Macbeth/pull/569

i looked at doing the same for methods like `get_snapshot_by_id` but `id` and `height`  are separate fields on `SnapShot`

crates/storage/provider/src/traits/snapshot.rs
scottmillner: fixed in https://github.com/botanix-labs/Macbeth/pull/569

crates/node/core/src/args/state_sync.rs
```diff
@@ -0,0 +1,34 @@
+use clap::Args;
+
+/// The default maximum size for a snapshot in bytes (8 MB).
+pub(crate) const DEFAULT_MAX_SNAPSHOT_SIZE_BYTES: usize = 8 * 1024 * 1024; // 8 Mbs max size
```
arminsabouri: Per the [cometBFT Docs](https://docs.cometbft.com/main/spec/abci/abci++_app_requirements#taking-snapshots)  the suggested size of chunks is 10MB. Where did 8MB come from for the snapshot size? This appears too low.

scottmillner: made it 10mb and non-configurable(const in snapshot_manager)

crates/consensus/authority/src/comet_bft/abci.rs
scottmillner: yes

PR #510, opened 2024-12-10
feat(btc-server): Add conflicting input when retyring pegouts

scottmillner:


bin/btc-server/src/coordinator.rs
```diff
@@ -97,15 +99,15 @@ where
...
+    pub(crate) async fn add_round1_signing(
         &self,
         signing_session_id: &[u8; 32],
         frost_id: frost::Identifier,
         psbt: &Psbt,
     ) -> Result<(), CoordinatorError> {
         let start = Instant::now();
         self.db.get_key_package()?.ok_or(CoordinatorError::MissingKeyPackage)?;
-        validate_psbt(psbt, ROUND1, self.min_signers, &self.db)?;
+        validate_psbt(psbt, ROUND1, self.min_signers, &self.db, &self.pegout_scheduler).await?;
```
arminsabouri: Would it make more sense to pass a ref to a list of pending txs? You could also get a iter() from the db which is already passed in. I would try to see if sled has a smart way of doing lookups rather than trying to read in all of tracked inputs

bin/btc-server/src/main.rs
```diff
@@ -1299,4 +1301,98 @@ mod test {
...
+        // Add the key packages
+        app.db.set_pubkey_package(pk_package.clone()).expect("set public key package");
+        app.db.set_key_package(key_package.clone()).expect("set key package");
+
+        // generate 36 random bytes
+        let mut rng = thread_rng();
+        let mut tx_id = [0u8; 32];
+        rng.fill(&mut tx_id);
+        let tx_idx = rng.gen_range(0..u32::MAX);
+        let pegout_id = PegoutId::new(tx_id, tx_idx);
```
arminsabouri: We have a test util to generate random pegoutIds. I would use that to save on code diff

bin/btc-server/src/signer.rs
```diff
@@ -217,11 +219,7 @@ where
         // }
 
         // Validate PSBT
-        // TODO lets re-enable this one we can check for conflicting inputs
-        // Right now singers can get into a state where they are tracking txs that the
-        // coordinator is not. This will not be an issue if the signers just check for one
-        // conflicting input until then signers can just trust the coordinator
-        // validate_psbt(psbt, ROUND1, self.min_signers, &self.db)?;
+        validate_psbt(psbt, ROUND1, self.min_signers, &self.db, &self.pegout_scheduler).await?;
```
arminsabouri: We should think about feature flags here. 
if we feature flag this we can start to write int. tests for this code and ship it turned off. 
Otherwise We cant ship it

bin/btc-server/src/util.rs
```diff
@@ -1185,4 +1226,75 @@ mod util_tests {
...
+
+        // Add the key packages
+        app.db.set_pubkey_package(pk_package.clone()).expect("set public key package");
+        app.db.set_key_package(key_package.clone()).expect("set key package");
+
+        // generate 36 random bytes
+        let mut rng = thread_rng();
+        let mut tx_id = [0u8; 32];
+        rng.fill(&mut tx_id);
+        let tx_idx = rng.gen_range(0..u32::MAX);
```
arminsabouri: Same pegoutId comment here

arminsabouri (commented):
Just some comments at first glance. Overall this seems to be moving in the right direction!

bin/btc-server/src/coordinator.rs
scottmillner: yeah i def shouldve approached this a different way, not sure why i went this route

i need two things: the `tracked inputs` and the `pegout request ids` both of which i get from the `tracked_txs` 
im gonna see if i can do smart lookups with the db since it's already passed in

thanks for suggesting that!

scottmillner: refactored so no additional args are passed to `validate_psbt`, will just use the db
but see my comment about `has_conflicting_input`, not sure it's really much better than reading all the tracked_txs into memory

bin/btc-server/src/main.rs
scottmillner: done, thanks!

bin/btc-server/src/util.rs
scottmillner: done!

bin/btc-server/src/database.rs
```diff
@@ -559,6 +560,37 @@ impl Db {
         Ok(())
     }
 
+    /// Checks a PSBT has a conflicting input if it contains a tracked PegoutId.
```
scottmillner: @0xBEEFCAF3 
im not a big fan of what im doing here...its O(n2) since we have two for loops
was trying not to read all `tracked_txs` into memory first and this is the only way to really do it
sled doesn't offer any smarter way since I have to deserialize `Tx` first and then check it's fields

bin/btc-server/src/signer.rs
scottmillner: added feature flags
will start on integration test now

scottmillner: integration test added

bin/btc-server/src/util.rs
```diff
@@ -206,6 +209,11 @@ pub fn validate_psbt(
     // Sanity check for # of inputs and outputs
     if psbt.inputs.is_empty() {
         return Err(ValidatePSBTError::NoInputs);
+    } else {
```
arminsabouri: I agree that this if else makes sense but since we return err in the first if can we just call `db.has_conflicting_input(psbt)?;` without the else. Imo its a bit easier to read

bin/test-suite/src/suite/consensus/frost/test_conflicting_input.rs
```diff
@@ -0,0 +1,173 @@
...
+        .expect("btc server rpc clients to be defined");
+
+    // Getting public key should fail for all clients
+    for client in &mut clients {
+        let pk = client.get_public_key(tonic::Request::new(client::Empty {})).await;
+        assert!(pk.is_err());
+        let err = pk.err().ok_or_else(|| anyhow::anyhow!("missing key package"))?;
+        assert_eq!(err.code(), tonic::Code::Internal);
+        assert!(err.message().contains("missing key package"));
+    }
```
arminsabouri: We have this specific assertion in a different test. I would just remove this from the conflicting inputs test.

bin/test-suite/src/suite/consensus/frost/test_conflicting_input.rs
```diff
@@ -0,0 +1,173 @@
...
+        send_pegout_notification(c, amount.to_sat(), 1, pegout_id, spk.clone()).await?;
+    }
+
+    // Create a new psbt that honors the pending pegout that is already being tracked in the
+    // previous psbt.
+
+    // FAILURE CASES:
+    // 1) Signers will reject the psbt if it contains no conflicting input.
+    // 2) The network will reject the broadcast psbt if it contains a conflicting input with error
+    //    "txn-mempool-conflict" since the first psbt is still in the mempool.
```
arminsabouri:
```suggestion
    //    "txn-mempool-conflict" since the first tx is still in the mempool.
```

bin/test-suite/src/suite/consensus/frost/test_conflicting_input.rs
```diff
@@ -0,0 +1,173 @@
...
+    // resend the same pegout notification so it is a pending pegout
+    for c in clients.iter_mut() {
+        send_pegout_notification(c, amount.to_sat(), 1, pegout_id, spk.clone()).await?;
+    }
+
+    // Create a new psbt that honors the pending pegout that is already being tracked in the
+    // previous psbt.
+
+    // FAILURE CASES:
+    // 1) Signers will reject the psbt if it contains no conflicting input.
```
arminsabouri:
```suggestion
    // 1) Signers will reject the psbt if it contains no conflicting input. This does not happen b/c the coordinator will include a conflicting input and signers will validate against this
```

bin/test-suite/src/suite/consensus/frost/test_conflicting_input.rs
```diff
@@ -0,0 +1,173 @@
...
+    // 2) The network will reject the broadcast psbt if it contains a conflicting input with error
+    //    "txn-mempool-conflict" since the first psbt is still in the mempool.
+    // We want failure case 2.
+    if let Err(e) = do_signing(&mut clients, &bitcoind, &[2u8; 32]).await {
+        assert!(e.to_string().contains("txn-mempool-conflict"));
+        return Ok(())
+    }
+
+    Err(anyhow::anyhow!("expected txn-mempool-conflict error"))
```
arminsabouri, line 164: Another testing case that is interesting is to do round 1 FROST but never do round 2. And then in a new signing session do round 1 & round 2 and broadcast.

bin/btc-server/src/util.rs
```diff
@@ -206,6 +209,11 @@ pub fn validate_psbt(
...
+        // Validate psbt contains conflicting input if retrying a pegout
+        if cfg!(feature = "conflicting_input") {
+            db.has_conflicting_input(psbt)?;
```
arminsabouri: Any opinions on creating a the `has_conflicting_input` method in `util.rs` instead of db? And passing the db down? Just thinking domain seperation

bin/btc-server/src/signer.rs
```diff
@@ -336,18 +335,17 @@ where
             psbt_in.set_partial_signature(self.identifier, &sigs);
         }
 
-        let _tx = psbt.clone().extract_tx()?;
-
```
arminsabouri: Do we want to remove this? Eventho _tx is never used extract_tx() performs several fee sanity checks.
maybe we need a comment to justify the use here.

bin/btc-server/src/database.rs
arminsabouri: is there a way to do contains with sled?
im thinking 
```
for pegout_id in pegouts_ids {
   if self.tracked_txs.contains(pegout_id) -> look_for_conflicting_input()
}

**Review by arminsabouri — CHANGES_REQUESTED** [review]
Thanks for adding those feature flags!  I was thinking env vars but I like your approach much better. just leaving some ideas around code structure and testing approach. And found one extract_tx() call that I think was removed b.c it wasn't commented well.

File: bin/btc-server/src/database.rs (hunk 5 above)
Comment (SatwikPrabhu) [reply bin/btc-server/src/database.rs]: Isn't this technically O(n3) because we are also using any(). I think that maybe it is possible to do O(n2). I'll get back if what I have in mind makes sense

Comment (SatwikPrabhu) [reply bin/btc-server/src/database.rs]: For example, maybe the following (untested) is  valid and also having better complexity:
`


    let psbt_inputs: HashSet<_> = psbt.unsigned_tx.input
        .iter()
        .map(|input| input.previous_output)
        .collect();

    let pegout_ids: HashSet<_> = psbt
        .pegout_ids()
        .iter()
        .map(|id| PegoutId::from_bytes(id).expect("valid pegout id"))
        .collect();

    for res in self.tracked_txs.iter() {
        let (_k, v) = res?;
        let tx: pegout_scheduler::Tx =
            ciborium::de::from_reader(v.as_ref()).expect("corrupt db: tracked tx");
        
        let has_matching_pegout = tx.pegout_requests
            .iter()
            .any(|request| pegout_ids.contains(&request.id));
            
        if has_matching_pegout {
            let has_shared_input = tx.inputs()
                .any(|input| psbt_inputs.contains(&input));
                
            if has_shared_input {
                return Ok(());
            } else {
                return Err(Error::NoConflictingInput);
            }
        }
    }`

File: bin/btc-server/src/main.rs [hunk 13]
```diff
@@ -1298,4 +1298,162 @@ mod test {
 
         assert_eq!(response.err().expect("error exists").to_string(), "invalid change output",);
     }
+
+    #[tokio::test]
+    // this test ensures a conflicting input in included in the psbt when retyring the same
```
SatwikPrabhu: Nit:
```suggestion
    // this test ensures a conflicting input is included in the psbt when retrying the same
```

bin/btc-server/src/signer.rs
```diff
@@ -336,18 +335,17 @@ where
...
 
-        // TODO(armins) we will need to re-enable this once we have conflicting inputs check
-        // self.add_tracked_tx(tx.clone(), &pending_pegouts, SystemTime::now()).await?;
-        // self.db.remove_pending_pegout(&pending_pegout_ids)?;
-        // self.db.flush()?;
+        if cfg!(feature = "conflicting_input") {
+            self.add_tracked_tx(tx.clone(), &pending_pegouts, SystemTime::now()).await?;
+            self.db.remove_pending_pegout(&pending_pegout_ids)?;
+            self.db.flush()?;
+        }
```
SatwikPrabhu: Maybe the assignment of the variable `pending_pegout_ids` should be moved inside the if as well?

bin/btc-server/src/database.rs
scottmillner: > is there a way to do contains with sled? im thinking
> 
> ```
> for pegout_id in pegouts_ids {
>    if self.tracked_txs.contains(pegout_id) -> look_for_conflicting_input()
> }
> ```

unfortunately there isn't a way to do that
you can only do `contains_key` which is `txid` in this case and isn't useful

sled doesn't have a concept of a complex data structure at rest...values are just arbitrary byte arrays
would be easy with a relational database but we have to deserialize `Tx` and then check it's fields

scottmillner: > For example, maybe the following (untested) is valid and also having better complexity: `
> 
> ```
> let psbt_inputs: HashSet<_> = psbt.unsigned_tx.input
>     .iter()
>     .map(|input| input.previous_output)
>     .collect();
> 
> let pegout_ids: HashSet<_> = psbt
>     .pegout_ids()
>     .iter()
>     .map(|id| PegoutId::from_bytes(id).expect("valid pegout id"))
>     .collect();
> 
> for res in self.tracked_txs.iter() {
>     let (_k, v) = res?;
>     let tx: pegout_scheduler::Tx =
>         ciborium::de::from_reader(v.as_ref()).expect("corrupt db: tracked tx");
>     
>     let has_matching_pegout = tx.pegout_requests
>         .iter()
>         .any(|request| pegout_ids.contains(&request.id));
>         
>     if has_matching_pegout {
>         let has_shared_input = tx.inputs()
>             .any(|input| psbt_inputs.contains(&input));
>             
>         if has_shared_input {
>             return Ok(());
>         } else {
>             return Err(Error::NoConflictingInput);
>         }
>     }
> }`
> ```

thanks, ill take a look!

bin/btc-server/src/signer.rs
scottmillner: done, thanks!

bin/btc-server/src/signer.rs
scottmillner: added back, sorry i missed that!

bin/btc-server/src/util.rs
scottmillner: yeah was thinking to do that and forgot so thanks!

bin/btc-server/src/util.rs
scottmillner: done!

bin/test-suite/src/suite/consensus/frost/test_conflicting_input.rs
scottmillner: done!

bin/test-suite/src/suite/consensus/frost/test_conflicting_input.rs
scottmillner: not quite following:
- a tx is tracked in `finalize_signing` when it's broadcast (after round 2 is finished)
- so there would be no conflicting input needed in this scenario

what am i missing?

bin/btc-server/src/database.rs
scottmillner: @SatwikPrabhu went with a slightly refactored version of what you suggested
thanks, it's def better now!

bin/btc-server/src/signer.rs
SatwikPrabhu: Maybe my comment was not totally clear (sorry abou that), because I meant moving the line with `let pending_pegout_ids = ` inside the if block that checks for the feature. I thought that this is better because we don't use the variable if the feature is not enabled.

bin/test-suite/src/suite/consensus/frost/test_conflicting_input.rs
arminsabouri: txs are tracked as soon as signers sign. if they start tracking only after witness has been added then a coordinator could replay round 1 spending to the same set of outputs with different set of inputs and signers will double spend. This is why its crutial that signers start tracking post round1.

bin/btc-server/src/util.rs
```diff
@@ -408,17 +422,53 @@ pub(crate) fn validate_outputs(psbt: &Psbt, db: &database::Db) -> Result<(), Val
...
+
+/// Checks a PSBT has a conflicting input if it contains a tracked PegoutId.
+pub fn has_conflicting_input(db: &Db, psbt: &Psbt) -> Result<(), ConflictingInputError> {
+    let psbt_inputs: HashSet<_> =
+        psbt.unsigned_tx.input.iter().map(|input| input.previous_output).collect();
+
+    let pegout_ids = psbt
+        .pegout_ids()
+        .iter()
+        .map(|id| PegoutId::from_bytes(id).expect("valid pegout id"))
```
arminsabouri: should we error out here instead of expect?

bin/btc-server/src/util.rs
```diff
@@ -1185,4 +1235,38 @@ mod util_tests {
...
+        db.store_tracked_tx(&tracked_tx).unwrap();
+        db.flush().unwrap();
+
+        // create a psbt with the pegout request from the tracked tx
+        let mut psbt = create_psbt(1, 1, None);
+        psbt.outputs[0].set_pegout_id(pegout_id.as_bytes());
+
+        let res = has_conflicting_input(&db, &psbt);
+        // should be an error since the psbt does not have a conflicting input
+        assert_eq!(res.unwrap_err().to_string(), "no conflicting input");
```
arminsabouri, line 1334: Can we enumerate over a few more test cases?
- conflicting inputs not needed
- conflicting inputs across two different txs. Lets say we are retrying tx A with outputs A_o and B with B_o. Tx C will be a batch that spends another outputs but also retries A_o and B_o. So we need conflicting inputs from both

bin/btc-server/src/util.rs
```diff
@@ -1185,4 +1235,38 @@ mod util_tests {
...
+            pegout_idxs: vec![0],
+            pegout_requests,
+            created: SystemTime::now(),
+        };
+        db.store_tracked_tx(&tracked_tx).unwrap();
+        db.flush().unwrap();
+
+        // create a psbt with the pegout request from the tracked tx
+        let mut psbt = create_psbt(1, 1, None);
+        psbt.outputs[0].set_pegout_id(pegout_id.as_bytes());
```
arminsabouri, line 1330: its interesting that this just setting the pegout id is all it takes for a conflicting inputs check. In someway I think the pegout id needs to commit to the output. im not sure yet. What do you think?

arminsabouri (commented):
Just a few more comments. And after reviewing this code I think I found some missing checks in out psbt validation method. So i'll be fixing those. Particularly the case for duplicate outputs

bin/btc-server/src/signer.rs
scottmillner: i understood and made that change 😃

bin/test-suite/src/suite/consensus/frost/test_conflicting_input.rs
scottmillner: ah gotchya, at the moment the tx is only being tracked in `get_round2_signing_package` and `finalize_signing`

but we need to track the tx before `get_round2_signing_package` as well and write an int test for the case above

arminsabouri: Sorry I mispoke. the tx is being tracked in `get_round2_signing_package` and that is the correct functionality. The only reason we also track in finalize is incase he coordinator didnt sign (which in our current state machine would never be the case).

So the test case is really: Ensure that signers don't need conflicting inputs if they never got to round 2.

bin/btc-server/src/util.rs
scottmillner: maybe we can whiteboard this a bit

my thoughts:
- a `PegoutId` is composed of `txid` and `idx` on the L2 which is a unique and discreet event that will never happen again
- an `Output` is composed of `value` and `script_pubkey` and could occur multiple times (ie, the user pegouts at a different time for the same amount to the same address)

so what does commiting to the `Output` give us? 
im probably missing something here

scottmillner: we discussed offline
- problem: how can we know if btc-server is given malicious payloads to sign 
- a possible approach is adding additional keys for authentication (like Liquid) but needs further research and is outside the scope of this PR

bin/btc-server/src/util.rs
scottmillner: we should and done!

bin/btc-server/src/util.rs
scottmillner: done!

bin/test-suite/src/suite/consensus/frost/test_conflicting_input.rs
scottmillner: just a follow up on this one: `validate_psbt` which isn't called in round 1 is the only way to trigger the `has_conflicting_input` check

bin/btc-server/src/coordinator.rs
```diff
@@ -263,18 +314,25 @@ where
...
             )
             .map_err(CoordinatorError::CoinSelection)?;
 
-        let selected = selection
+        // include conflicting utxos when selecting from available utxos
+        // this is done after the coin selection result above to prevent duplicate utxos
+        conflicting_utxos.into_iter().for_each(|(op, u)| {
+            available_utxos.insert(op, u);
+        });
+        let mut selected = selection
```
SatwikPrabhu: Why is this `mut`?

scottmillner: ah thanks, it shouldn't be, it was from a previous refactor that i changed
done!

bin/btc-server/src/coordinator.rs
```diff
@@ -230,13 +234,60 @@ where
...
             .collect::<HashMap<_, _>>();
 
+        // if we are retrying pegouts, we need to add a conflicting input for each tracked tx
+        // that honors each pegout
+        let mut conflicting_inputs: Result<Vec<Utxo>, CoordinatorError> = Ok(vec![]);
+        let mut conflicting_utxos: HashMap<OutPoint, Utxo> = HashMap::new();
+        if cfg!(feature = "conflicting_input") {
+            let tracked_pegout_request_ids = tracked_txs
+                .iter()
+                .flat_map(|tx| tx.pegout_requests.clone().into_iter().map(|p| p.id))
```
arminsabouri: Nit: do we need to clone here?

bin/test-suite/src/suite/consensus/frost/test_conflicting_input.rs
scottmillner: i added a test to cover this called `test_round1_then_new_signing_session`

bin/btc-server/src/coordinator.rs
```diff
@@ -255,17 +319,24 @@ where
...
                 target_amount.to_sat(),
                 &change_script, // drain_script
             )
             .map_err(CoordinatorError::CoinSelection)?;
 
+        // include conflicting utxos when selecting from available utxos
+        // this is done after the coin selection result above to prevent duplicate utxos
+        conflicting_utxos.into_iter().for_each(|(op, u)| {
+            available_utxos.insert(op, u);
+        });
```
arminsabouri, line 295: Just a note that https://github.com/botanix-labs/Macbeth/pull/358 will be extracting coin selection into its own mod. Once that happens we can write granular unit tests for this

bin/btc-server/src/coordinator.rs
```diff
@@ -226,8 +232,66 @@ where
...
+                .filter_map(|(_, pegout_id)| {
+                    pegout_id.as_ref().filter(|id| tracked_pegout_request_ids.contains(id))
+                })
+                .collect();
+
+            // get a tracked input for each matching pegout
+            let matching_tracked_inputs: Result<Vec<OutPoint>, CoordinatorError> = tracked_txs
+                .iter()
+                .filter(|tx| tx.pegout_requests.iter().any(|p| matching_pegouts.contains(&&p.id)))
+                .map(|tx| tx.inputs().next().ok_or_else(|| CoordinatorError::NoConflictingInputs))
```
arminsabouri, line 268: Aplogies if I am just misunderstanding this line. But why is this map needed? Are we just checking after we filter the tracked txs for matching pegout ids that the inputs are non empty?

bin/btc-server/src/coordinator.rs
```diff
@@ -226,8 +232,66 @@ where
...
+            conflicting_inputs = matching_tracked_inputs
+                .iter()
+                .map(|op| {
+                    utxos
+                        .get(&op)
+                        .ok_or_else(|| CoordinatorError::MissingUtxoForConflictingInput)
+                        .and_then(|u: &Utxo| {
+                            // Conflicting utxos will be added to available utxos before finishing
+                            // coin selection
+                            conflicting_utxos.insert(op.clone(), u.clone());
```
arminsabouri, line 282: Do we need to include every UTXO from the prev tracked tx? is sufficient just to add one. I worry that requiring all these utxos could lead to unnatural utxo selection.

bin/btc-server/src/signer.rs
```diff
@@ -344,18 +343,21 @@ where
...
         if let Some(telemetry) = self.telemetry.as_ref() {
             telemetry.update_pending_pegouts(pending_pegouts.len() as i64);
         }
-        let _pending_pegout_ids = pending_pegouts.iter().map(|p| p.id).collect::<Vec<PegoutId>>();
 
-        // TODO(armins) we will need to re-enable this once we have conflicting inputs check
-        // self.add_tracked_tx(tx.clone(), &pending_pegouts, SystemTime::now()).await?;
-        // self.db.remove_pending_pegout(&pending_pegout_ids)?;
-        // self.db.flush()?;
+        if cfg!(feature = "conflicting_input") {
```
arminsabouri, line 354: One other note that came to mind. Do we need to update the tracked txs if we are including a conflicting input? 
Bc we will have a set of pegouts we are tracking and now there will be a another tx with a subset of the same pegouts. It seems like we can track both txs but once one is confirmed the other MUST be deleted. And perhaps we should also have a retention period for how long we track the pegout that is not in a block (mempool tracking).

This can get complicated. Lets hop on a quick call to talk the rest of this out

arminsabouri (commented):
This PR is pretty much ready to go but I would like to hop on a quick call to make sure we have covered all of our edge cases.

bin/btc-server/src/signer.rs
SatwikPrabhu: Im generally available for a call after an hour. Let me know when is a good time.

arminsabouri:
@scottmillner im assigning my self to this PR to rebase on the latest in main. I've discovered a bug in the fee calculation

bin/btc-server/src/coordinator.rs
scottmillner: this is what im thinking which may be wrong:
- im finding every tracked tx that was honoring a pegout that is being retried in this psbt
- then getting an input from each of those tracked txs
- so there is a conflicting input from each of those tracked txs

bc:
- if i don't include a conflicting input from each tracked tx that has a pegout being retried
- that tracked tx and the current psbt about to be broadcast could both be confirmed and we have double spent the same output

bin/btc-server/src/coordinator.rs
scottmillner: nope, it was bc i was using `into_iter()` when i didn't need to

using `iter()` now and removed `clone()`, thanks!

bin/btc-server/src/coordinator.rs
scottmillner: see my answer above which describes my thinking

bin/btc-server/src/signer.rs
scottmillner: yes i think we do
based on our follow-up call, made new ticket: https://github.com/botanix-labs/botanix/issues/858

bin/btc-server/src/coordinator.rs
arminsabouri: Actually my question was regarding the .inputs().next() semantics you used rather than the check it self.

scottmillner: ah my bad, `.inputs().next()` returns the first element by the iterator...so i'm grabbing the first input to use as a conflicting input

PR #512, opened 2024-12-11
feat(repo): Pending pegouts controller

rwlockbg:
Related issue: https://github.com/botanix-labs/botanix/issues/757

### DIFF
--- a/bin/btc-server/src/server.rs
+++ b/bin/btc-server/src/server.rs
@@ -1,10 +1,5 @@
-use std::time::{Duration, SystemTime};
-
 use base64::{engine::general_purpose, Engine as _};
-use bitcoin::{
-    absolute::LockTime, hashes::Hash, psbt::Psbt, transaction::Version, Amount, BlockHash,
-    OutPoint, ScriptBuf, Sequence, Transaction, TxIn, TxOut, Txid, Witness,
-};
+use bitcoin::{hashes::Hash, psbt::Psbt, Amount, BlockHash, ScriptBuf, TxOut};
 use bitcoincore_rpc::{json::EstimateMode, RpcApi};
 use frost_secp256k1_tr as frost;
 use reth_btc_wallet::util::VerifyingKeyExt;
@@ -17,7 +12,8 @@ use crate::{
     pegout_id::PegoutId,
     pegout_scheduler::{PegoutRequest, Tx},
     rpc::{self},
-    util, App,
+    util::{self, get_available_utxos},
+    App,
 };
 
 const JWT_HEADER_KEY: &str = "trace-proto-bin";
@@ -224,6 +220,17 @@ where
             value: Amount::from_sat(req.amount),
             botanix_height: req.height,
         };
+
+        // do not accept any pending pegouts when there are not available UTXOs and tracked inputs
+        let (available_utxos, tracked_inputs) =
+            get_available_utxos(&self.db).await.map_err(|e| {
+                error!("Failed to get all available utxos. error: {}", e);
+                internal!("Failed to get all available utxos. error: {}", e)
+            })?;
+        if available_utxos.is_empty() && tracked_inputs.is_empty() {
+            error!("Received a pegout request when there are no utxos or pending transactions");
+            return Ok(tonic::Response::new(rpc::Empty {}));
+        }
 
         self.db.store_pending_pegout(&pegout_req)?;
         self.db.flush()?;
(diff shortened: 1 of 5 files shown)


scottmillner (dismissed):
thanks for doing this which will prevent that annoying error log now 😃 !

looks like `test_notify_pegout` is failing

bin/btc-server/src/server.rs
```diff
@@ -225,6 +221,16 @@ where
             botanix_height: req.height,
         };
 
+        // do not accept any pending pegouts when there are not available UTXOs
+        let available_utxos =
+            get_available_utxos(&self.pegout_scheduler, &self.db).await.map_err(|e| {
```
arminsabouri: its possible that we may pre-maturely reject a pegout even if there are tracked tx i.e change coming into the spendable utxo set. I believe the check should be if there are no available utxos and no tracked txs then there is no possible way there can be pending tx.

arminsabouri (changes requested):

bin/btc-server/src/server.rs
scottmillner: yeah that makes sense bc the `tracked_tx` could get deeply confirmed and the change returned to the utxo set before the epoch ends and the pegouts are honored

rwlockbg: yep, totally! I made the change now.

scottmillner (dismissed):

bin/btc-server/src/server.rs
```diff
@@ -225,6 +221,16 @@ where
...
         };
 
+        // do not accept any pending pegouts when there are not available UTXOs and tracked inputs
+        let (available_utxos, tracked_inputs) =
+            get_available_utxos(&self.pegout_scheduler, &self.db).await.map_err(|e| {
+                error!("Failed to get all available utxos. error: {}", e);
+                internal!("Failed to get all available utxos. error: {}", e)
+            })?;
+        if available_utxos.is_empty() && tracked_inputs.is_empty() {
+            return Ok(tonic::Response::new(rpc::Empty {}));
```
arminsabouri, line 232: Sorry i should have picked up on this earlier. can we log an error here.
"Recieved a pegout request when there are no utxos or pending transactions"

bin/btc-server/src/util.rs
```diff
@@ -408,6 +411,24 @@ pub(crate) fn validate_outputs(psbt: &Psbt, db: &database::Db) -> Result<(), Val
...
+    db: &database::Db,
+) -> Result<(HashMap<OutPoint, Utxo>, HashSet<OutPoint>), database::Error> {
+    let utxos: HashMap<OutPoint, Utxo> =
+        db.iter_utxos().try_fold(HashMap::new(), |mut map, r| {
+            let utxo = r?;
+            map.insert(utxo.outpoint, utxo);
+            Ok::<HashMap<bitcoin::OutPoint, Utxo>, database::Error>(map)
+        })?;
+    // Filter the ones that are still pending and conflict with pending txs.
+    let tracked_inputs = pegout_scheduler.lock().await.tracked_inputs();
```
arminsabouri: I think you can just pull the tracked inputs from the db that way you dont need to lock

bin/btc-server/src/server.rs
rwlockbg: done

bin/btc-server/src/util.rs
rwlockbg: done

arminsabouri (approved):

scottmillner (dismissed):

arminsabouri (changes requested):
Looks like some unit tests in btc server are failing for legit reasons

rwlockbg:
> Looks like some unit tests in btc server are failing for legit reasons

Yes, the notify pegouts is failing for legit reasons. Maybe we should think about whether this code application is always correct. In the case of pegouts there are 0 tracked txs in the db hence its failing. Suggestions ?

arminsabouri:
> > Looks like some unit tests in btc server are failing for legit reasons
> 
> Yes, the notify pegouts is failing for legit reasons. Maybe we should think about whether this code application is always correct. In the case of pegouts there are 0 tracked txs in the db hence its failing. Suggestions ?

@rwlockbg to get around this in the test, I would just add some random utxos to the db before we notify the pegout. And I would also add a test condition to the `test_notify_pegouts` that ensures no pending pegouts are added when there are no utxos available.

arminsabouri (approved):

PR #516, opened 2024-12-16
refactor(workflows): update image push to build off tags

birozuru:


.github/workflows/deploy-testnet-v1.yml
```diff
@@ -31,15 +30,21 @@ jobs:
         id: sha
         run: echo "short_sha=$(git rev-parse --short HEAD)" >> $GITHUB_OUTPUT
```
arminsabouri, line 31: Can we delete this step above as its always writing to `$GITHUB_OUTPUT`

arminsabouri (approved):
nice!!

.github/workflows/deploy-testnet-v1.yml
birozuru: yeah we can. 
The output can also be gotten from 'steps.sha.outputs.short_sha'

scottmillner (approved):

arminsabouri (approved):

PR #518, opened 2024-12-16
Set reth version to 1.0.0

SatwikPrabhu:


bin/reth/Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "reth"
 description = "Reth node implementation"
 default-run = "reth"
-version = { workspace = true }
+version = "1.0.0"
```
arminsabouri, line 5: Will this version reflect in the rpc call? https://www.quicknode.com/docs/ethereum/web3_clientVersion

arminsabouri (changes requested):
Do you mind also changing the hardcoded app version in abci.rs? Ideally we can just pull in one version and not hardcode it in multiple places?

https://github.com/botanix-labs/Macbeth/blob/4147460070a646bfbc6379ea1ab8bf47084803d0/crates/consensus/authority/src/comet_bft/abci.rs#L380

bin/reth/Cargo.toml
SatwikPrabhu: It should be fixed now.

arminsabouri (approved):

scottmillner (approved):

PR #520, opened 2024-12-17
Sat/test/mock btc server client

SatwikPrabhu:


SatwikPrabhu:
Let me know what you guys think of this mock. I've left a few functionalities unimplemented since they seem to be unimportant for unit testing dkg and signing state machines. 

I can use these mocks for the those unit tests if they look fine to you guys.

bin/btc-server/src/extended_client.rs
```diff
@@ -74,14 +74,49 @@ macro_rules! generate_method {
...
+    async fn abort_signing(&mut self, request: Empty) -> Result<Empty, GrpcClientError>;
+    async fn get_signing_status(&mut self, request: GetSigningStatusRequest) -> Result<GetSigningStatusResponse, GrpcClientError>;
+    async fn get_session_ids(&mut self, request: GetSessionIdsRequest) -> Result<GetSessionIdsResponse, GrpcClientError>;
+    async fn health_check(&mut self, request: Empty) -> Result<Empty, GrpcClientError>;
+    async fn tx_index_new_checkpoint(&mut self, request: SyncTxIndexRequest) -> Result<Empty, GrpcClientError>;
+    async fn reset_all_utxos(&mut self, request: ResetAllUtxosRequest) -> Result<Empty, GrpcClientError>;
+    async fn get_all_utxos(&mut self, request: Empty) -> Result<GetAllUtxosResponse, GrpcClientError>;
+    async fn get_tracked_txs(&mut self, request: Empty) -> Result<GetTrackedTxsResponse, GrpcClientError>;
+    async fn get_pending_pegouts(&mut self, request: Empty) -> Result<GetPendingPegoutsResponse, GrpcClientError>;
+    async fn reset_wallet_state(&mut self, request: ResetWalletStateRequest) -> Result<Empty, GrpcClientError>;
```
arminsabouri: This is a lot to mock. There is alternative approach here where we only mock one method and all other methods just end up calling that one. take a look [RPCApi](https://github.com/rust-bitcoin/rust-bitcoincore-rpc/blob/master/client/src/client.rs#L221). All the methods in that trait end up calling `call` so that ends up being the only method that really needs to get mocked.

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -76,7 +76,7 @@ pub struct ABCIClientBuilder<EF, BF, DB> {
     storage: Storage<EF, BF, DB>,
     bitcoin_checkpoint: BitcoinCheckpoint,
     network_handle: NetworkHandle,
-    btc_server: Option<BtcServerExtendedClient>,
+    btc_server: Option<BtcServerExtendedClientImpl>,
```
arminsabouri: Do we not want to use the trait type here? That way we can pass in the mock

bin/btc-server/src/extended_client.rs
```diff
@@ -74,14 +74,49 @@ macro_rules! generate_method {
...
+/// Bitcoin Server Client implementation with extended authentication credentials
 #[derive(Clone, Debug)]
-pub struct BtcServerExtendedClient {
+pub struct BtcServerExtendedClientImpl {
     client: BtcServerClient<tonic::transport::channel::Channel>,
     jwt_secret: Option<JwtSecret>,
 }
 
-impl BtcServerExtendedClient {
+impl BtcServerExtendedClientImpl {
```
arminsabouri: Nit: maybe a better name could be `BtcServerClient`

arminsabouri (commented):
I havent reviewed the "meat" of this PR which is in test_utils yet. I wanted to leave my initial thoughts in case it did change the code in test_utils. awesome stuff :fire:

bin/btc-server/src/extended_client.rs
SatwikPrabhu: That is indeed interesting and new to me actually. I'll try it out to see if its feasible here

crates/consensus/authority/src/comet_bft/abci.rs
SatwikPrabhu: Hmm. I did try to pass the trait type here, but it eventually got a bit messy at the `AuthorityConsensusBuilder`. I can try again to see if I missed something.

bin/btc-server/src/extended_client.rs
SatwikPrabhu: I agree. Thanks for the suggestion. Although now that I look closer, there seems to be another struct with this name. So not sure now

crates/consensus/authority/src/comet_bft/abci.rs
scottmillner: let me know if you need any help using the trait and getting it to compile
we need to use the trait so we can mock

SatwikPrabhu: Gotcha. I'll give it a try and let you know if I get stuck.

arminsabouri:
@SatwikPrabhu one more thing do you mind squashing your commits into one and following the guidelines here https://github.com/botanix-labs/botanix/blob/master/CONTRIBUTING.MD

crates/consensus/authority/src/test_utils.rs
```diff
@@ -0,0 +1,582 @@
...
+            })
+            .collect();
+
+        Self {
+            peer_id,
+            test_vectors,
+            added_round1_packages: Arc::new(RwLock::new(BTreeMap::new())),
+            added_round2_packages: Arc::new(RwLock::new(BTreeMap::new())),
+            added_round1_signing_packages: Arc::new(RwLock::new(BTreeMap::new())), // Initialize new field
+            added_round2_signing_packages: Arc::new(RwLock::new(BTreeMap::new())), // Initialize new field
```
arminsabouri: Nit: comments unneeded

crates/consensus/authority/src/test_utils.rs
```diff
@@ -0,0 +1,582 @@
...
+            public_key_package,
+            state: Arc::new(RwLock::new(BTreeMap::new())),
+            dkg_state: Arc::new(RwLock::new(DKGState::Initial)),
+            coordinator_map: Arc::new(RwLock::new(BTreeMap::new())),
+        }
+    }
+
+    // Helper method to generate round1 signing packages
```
arminsabouri: Use `///` rust docs

crates/consensus/authority/src/test_utils.rs
```diff
@@ -0,0 +1,582 @@
...
+        let (nonces, commitments) = generate_signing_test_vectors(
+            &test_data.key_package,
+        )?;
+
+        test_data.signing_round1_nonces = Some(nonces);
+        test_data.signing_round1_commitments = Some(commitments);
+        Ok(())
+    }
+
+    // Helper method to generate round2 signing packages
```
arminsabouri: Use `///` rust docs

crates/consensus/authority/src/test_utils.rs
```diff
@@ -0,0 +1,582 @@
...
+
+/// Mock BTC server client for testing
+#[derive(Clone)]
+pub struct MockBtcServerClient {
+    pub peer_id: String,
+    pub test_vectors: Vec<DkgTestData>,
+    pub added_round1_packages: Arc<RwLock<BTreeMap<Vec<u8>, Vec<u8>>>>,
+    pub added_round2_packages: Arc<RwLock<BTreeMap<Vec<u8>, Vec<u8>>>>,
+    pub added_round1_signing_packages: Arc<RwLock<BTreeMap<Vec<u8>, Vec<u8>>>>, // Add new field
+    pub added_round2_signing_packages: Arc<RwLock<BTreeMap<Vec<u8>, Vec<u8>>>>, // Add new field
```
arminsabouri: unneeded comments

crates/consensus/authority/src/test_utils.rs
```diff
@@ -0,0 +1,582 @@
...
+    pub identifier: Identifier,
+    pub round1_package: round1::Package,
+    pub round1_secret: round1::SecretPackage,
+    pub round2_packages: BTreeMap<Identifier, round2::Package>,
+    pub round2_secret: round2::SecretPackage,
+    pub key_package: KeyPackage,
+    pub public_key_package: PublicKeyPackage,
+    pub signing_round1_nonces: Option<SigningNonces>,
+    pub signing_round1_commitments: Option<SigningCommitments>,
+    pub signing_round2_share: Option<SignatureShare>,
```
arminsabouri, line 39: Should signing stuff be in a struct called DKG*
maybe we rename the struct to just TestData or put this somewhere else

crates/consensus/authority/src/test_utils.rs
```diff
@@ -0,0 +1,582 @@
...
+        });
+    }
+
+    test_data
+}
+
+/// Generate signing test vectors for a participant
+pub fn generate_signing_test_vectors(
+    key_package: &KeyPackage,
+) -> Result<(SigningNonces, SigningCommitments), String> {
```
arminsabouri: Result doesnt seem needed here

crates/consensus/authority/src/test_utils.rs
```diff
@@ -0,0 +1,582 @@
...
+    ))
+}
+
+/// Generate round2 signing package for a participant
+pub fn generate_signing_round2_package(
+    key_package: &KeyPackage,
+    round1_nonces: &SigningNonces,
+    received_round1_packages: &BTreeMap<Identifier, SigningCommitments>,
+    signing_session_id: [u8; 32],
+) -> Result<SignatureShare, String> {
```
arminsabouri: Maybe use AnyHow::Error here instead of String?
That way you can do `sign(...)?` and the error will just get propogated to the caller

crates/consensus/authority/src/test_utils.rs
```diff
@@ -0,0 +1,582 @@
...
+    pub added_round2_signing_packages: Arc<RwLock<BTreeMap<Vec<u8>, Vec<u8>>>>, // Add new field
+    pub public_key_package: PublicKeyPackage,
+    pub state: Arc<RwLock<BTreeMap<[u8; 32], SessionState>>>,
+    pub dkg_state: Arc<RwLock<DKGState>>,
+    pub coordinator_map: Arc<RwLock<BTreeMap<[u8; 32], u64>>>, // Track coordinator for each session
+}
+
+impl MockBtcServerClient {
+    pub fn new(min_signers: u16, peer_id: String) -> Self {
+        let test_vectors = generate_test_vectors(3, min_signers);
```
arminsabouri: Any reason we shouldnt pull max_signers also from new params?

crates/consensus/authority/src/test_utils.rs
```diff
@@ -0,0 +1,582 @@
...
+        let peer_idx = peer_id.parse::<usize>().unwrap_or(0);
+        let public_key_package = test_vectors[peer_idx].public_key_package.clone();
+        
+        // Initialize test vectors with None for signing fields
+        let test_vectors = test_vectors.into_iter()
+            .map(|mut tv| {
+                tv.signing_round1_nonces = None;
+                tv.signing_round1_commitments = None;
+                tv.signing_round2_share = None;
```
arminsabouri: Is this neccecary? I think they are also None after calling generate_test_vectors

crates/consensus/authority/src/test_utils.rs
```diff
@@ -0,0 +1,582 @@
...
+            added_round2_signing_packages: Arc::new(RwLock::new(BTreeMap::new())), // Initialize new field
+            public_key_package,
+            state: Arc::new(RwLock::new(BTreeMap::new())),
+            dkg_state: Arc::new(RwLock::new(DKGState::Initial)),
+            coordinator_map: Arc::new(RwLock::new(BTreeMap::new())),
+        }
+    }
+
+    // Helper method to generate round1 signing packages
+    pub fn generate_round1_signing(&mut self, session_id: [u8; 32], psbt: &[u8]) -> Result<(), String> {
```
arminsabouri: Same comment about AnyHow::Error here. Generally I would recommend using AnyHow for test related Results<>

arminsabouri (commented):
OK I took a look at test_utils.rs and this is awesome. Very impressive design! I have a few ideas.
I still think we should have a one or two required methods in the trait. For example 
```rust
pub fn call<RES, REQ>(&self, method_name, req: REQ): Result<RES, GrpcError>
```
and then call other methods just proxy to `self::call(...)`. The main reason i think this is a better design is b/c we can create different implementations of the grpc methods to introduce faulty behavior which then we can catch in our tests.  However, we still need the utils that you have written for DKG and signing. So ideally we have those as a standalone stateful helper struct which a mock would use.
e.g
```rust
pub struct HelperStruct {
    pub identifier: Identifier,
    pub round1_package: round1::Package,
    pub round1_secret: round1::SecretPackage,
    ...
}

pub struct MockBitcoinClient {
   pub state: HelperStruct
}

impl MockBitcoin for MockBitcoinClient {
   pub fn call<RES, REQ>(&self, method_name, req: REQ, req): Result<RES, GrpcError> 
   {
      // do nothing, no one should be calling call directly in the tests
   }

   pub fn get_public_key(...) {                       
        self.state.test_vectors.public_key
    }
}

pub struct MockBitcoinClientWhereGettingPublicKeyDoesntWork {
   pub state: HelperStruct
}

// reallly bad name for a struct but just to demonstrate the point
impl MockBitcoin for MockBitcoinClientWhereGettingPublicKeyDoesntWork {
 pub fn call<RES, REQ>(&self, method_name, req: REQ, req): Result<RES, GrpcError> 
   {
      // do nothing, no one should be calling call directly in the tests
   }

   pub fn get_public_key(...) {                       
        // This should cause a failure in the state machine and affect whats in storage
        let rando_bytes = rng.fillRandom(...);
        Ok(GetPublicKeyResponse {
            publickey: 
           hex::encode(rando_bytes),
        })
    }
}

```

I hope this gets my point across. Also happy to hop on a short call to discuss more

bin/btc-server/src/extended_client.rs
SatwikPrabhu: I tried doing this. But it was harder than I expected. Basically we need to use serde serialization/deserialization for some of bitcoinrpc vars. Not sure if it is worth trying it.

crates/consensus/authority/src/comet_bft/abci.rs
SatwikPrabhu: Done thanks!

crates/consensus/authority/src/test_utils.rs
SatwikPrabhu: Done!

crates/consensus/authority/src/test_utils.rs
SatwikPrabhu: Done!

crates/consensus/authority/src/test_utils.rs
SatwikPrabhu: Done!

crates/consensus/authority/src/test_utils.rs
SatwikPrabhu: Done!

crates/consensus/authority/src/test_utils.rs
SatwikPrabhu: Done!

crates/consensus/authority/src/test_utils.rs
SatwikPrabhu: Done!

crates/consensus/authority/src/test_utils.rs
SatwikPrabhu: Done!

crates/consensus/authority/src/test_utils.rs
SatwikPrabhu: Done!

crates/consensus/authority/src/test_utils.rs
SatwikPrabhu: Done!

arminsabouri:
@SatwikPrabhu three things.
1. There are some merge conflicts
2. The commits need to be squashed and follow the contribuiting guide https://github.com/botanix-labs/botanix/blob/master/CONTRIBUTING.MD
3. the last commit here is unverified. [9cf5c7b](https://github.com/botanix-labs/Macbeth/pull/520/commits/9cf5c7bac415d6dcce321ebcdb6caddc9766dcac) but the others are verified

bin/btc-server/src/extended_client.rs
arminsabouri: Why is bitcoinrpc involved here?

SatwikPrabhu:
> @SatwikPrabhu three things.
> 
> 1. There are some merge conflicts
> 2. The commits need to be squashed and follow the contribuiting guide https://github.com/botanix-labs/botanix/blob/master/CONTRIBUTING.MD
> 3. the last commit here is unverified. [9cf5c7b](https://github.com/botanix-labs/Macbeth/pull/520/commits/9cf5c7bac415d6dcce321ebcdb6caddc9766dcac) but the others are verified

My bad. I'll fix them right away.

bin/btc-server/src/extended_client.rs
SatwikPrabhu: All the trait methods have different request and response types. I don't think there's a straightfoward way to do this. I think using serde json is the way here. In that case, we have to change add the serde serialize/deserialize for all these types and I was running into some issues while doing this. So, I thought that maybe its not worth doing this in this PR.

My bad though. Its not the bitcoinrpc, its in 
https://github.com/botanix-labs/Macbeth/blob/main/bin/btc-server/client/src/btc_server.rs

arminsabouri: I imagined it would look something like this 
```rust 
pub trait BtcServerGrpcClient {
    fn get_client(&mut self) -> &mut BtcServerClient<tonic::transport::channel::Channel>;
    async fn call_method<Req, Resp>(&mut self, method_name: &str, request: tonic::Request<Req>) -> Result<Resp, GrpcClientError>;

    async fn health_check(&mut self) -> Result<(), GrpcClientError> {
        self.call_method("health_check", tonic::Request::new(Empty {})).await?;
        Ok(())
    }

}


pub struct BtcServerGrpcClientImpl {
    client: BtcServerClient<tonic::transport::channel::Channel>,
    jwt_secret: Option<JwtSecret>,
}

impl BtcServerGrpcClient for BtcServerGrpcClientImpl {
    fn get_client(&mut self) -> &mut BtcServerClient<tonic::transport::channel::Channel> {
        &mut self.client
    }
    
    
    async fn call_method<Req, Resp>(&mut self, method_name: &str, request: tonic::Request<Req>) -> Result<Resp, GrpcClientError> {
        let grpc_call = generate_method!(method_name, Req, Resp);
        grpc_call(&mut self, request).await
    }
}


```

arminsabouri: would something like this not work?

SatwikPrabhu: Yep this was what I was trying as well. The problem was that request and response types were different and I don't think its possible to do this when that's the case. Maybe there's a way

SatwikPrabhu: Do you want to hop on to a call maybe in 5 mins?

crates/consensus/authority/src/builder.rs
```diff
@@ -204,15 +201,19 @@ where
     /// Builds and returns the necessary components for the authority consensus, including the
     /// consensus itself, the client used to interact with the consensus, and the block
     /// production task.
-    pub async fn build(
+    pub async fn build<BtcClient>(
```
arminsabouri: Nit: BtcClient here could mean two things. 
Lets specify BtcServerClient

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -72,19 +73,19 @@ const VERIFY_ACCEPTED: i32 = 1;
 const VERIFY_REJECT: i32 = 2;
 
 #[derive(Clone, Debug)]
-pub struct ABCIClientBuilder<EF, BF, DB> {
+pub struct ABCIClientBuilder<EF, BF, DB, BtcClient> {
```
arminsabouri: Same comment here

bin/btc-server/src/extended_client.rs
```diff
@@ -46,42 +47,72 @@ impl GrpcClientError {
     }
 }
 
-/// Macro for generating grpc methods
+pub trait BtcServerGrpcClient {}
```
arminsabouri: is this used? if not lets rm

crates/consensus/authority/src/test_utils.rs
```diff
@@ -0,0 +1,606 @@
...
+
+#[allow(unused_variables)]
+impl BtcServerExtendedClient for MockBtcServerClient {
+    fn update_jwt_secret(&mut self, _jwt_secret: JwtSecret) {}
+
+    fn generate_jwt_token(&mut self) -> Option<String> {
+        None
+    }
+
+    fn get_round1_dkg_package<'a>(&'a mut self, _: Empty) -> Pin<Box<dyn Future<Output = Result<DkgPayload, GrpcClientError>> + Send + 'a>> {
```
arminsabouri, line 255: Nit:

can we use the BoxedFuture `Pin<Box<dyn Future<Output = Result<DkgPayload, GrpcClientError>> + Send + 'a>>` type?

crates/consensus/authority/src/test_utils.rs
```diff
@@ -0,0 +1,606 @@
...
+        &'a mut self,
+        request: ToSignRequest,
+    ) -> Pin<Box<dyn Future<Output = Result<SigningPackage, GrpcClientError>> + Send + 'a>> {
+        Box::pin(async move{
+            let peer_idx = self.peer_id.parse::<usize>().unwrap_or(0);
+            let test_data = &self.test_vectors[peer_idx];
+            
+            // Check if we have enough round1 signing packages
+            let added_round1 = self.added_round1_signing_packages.read().await;
+            if added_round1.len() < 2 {
```
arminsabouri: Does this need to be dynamic?

crates/consensus/authority/src/test_utils.rs
```diff
@@ -0,0 +1,606 @@
...
+        })
+    }
+
+    fn finalize_signing<'a>(
+        &'a mut self,
+        request: FinalizeSigningRequest,
+    ) -> Pin<Box<dyn Future<Output = Result<FinalizeSigningResponse, GrpcClientError>> + Send + 'a>>{
+        Box::pin(async move{
+            let added_round2 = self.added_round2_signing_packages.read().await;
+            if added_round2.len() < 2 {
```
arminsabouri: same thing here

crates/consensus/authority/src/test_utils.rs
```diff
@@ -0,0 +1,606 @@
...
+
+            // Create signing package and aggregate signature
+            let signing_package = frost_secp256k1_tr::SigningPackage::new(round1_commitments, &session_id);
+            let _signature = frost_secp256k1_tr::aggregate(
+                &signing_package,
+                &signature_shares,
+                &test_data.public_key_package,
+            ).map_err(|e| GrpcClientError::Call(tonic::Status::internal(e.to_string())))?;
+
+            // We don't need to verify the signature in tests since we're mocking
```
arminsabouri: Note that aggregate will verify each parital signature against the message. Im not sure if this will change anything in your test utils

crates/consensus/authority/src/test_utils.rs
```diff
@@ -0,0 +1,606 @@
...
+    fn notify_pegins<'a>(&'a mut self, _: NotifyPeginsRequest) -> Pin<Box<dyn Future<Output = Result<Empty, GrpcClientError>> + Send + 'a>> {
+        Box::pin(async move {
+            unimplemented!("Not required for core DKG/signing tests")
+        })
+    }
+
+    fn notify_pegout<'a>(&'a mut self, _: NotifyPegoutRequest) -> Pin<Box<dyn Future<Output = Result<Empty, GrpcClientError>> + Send + 'a>> {
+        Box::pin(async move {
+            unimplemented!("Not required for core DKG/signing tests")
+        })
```
arminsabouri: Nit: group all the unimplemented methods together

crates/consensus/authority/src/test_utils.rs
```diff
@@ -0,0 +1,606 @@
...
+            let mut packages = self.added_round2_packages.write().await;
+            packages.insert(request.identifier, request.payload);
+            Ok(Empty {})
+        })
+    }
+
+    fn get_public_key<'a>(&'a mut self, _: Empty) -> Pin<Box<dyn Future<Output = Result<GetPublicKeyResponse, GrpcClientError>> + Send + 'a>> {
+        Box::pin(async move {
+            let added_round2 = self.added_round2_packages.read().await;
+            if added_round2.len() < 2 { // Minimum required signers
```
arminsabouri: for DKG it would be N not 2.

crates/consensus/authority/src/test_utils.rs
```diff
@@ -0,0 +1,606 @@
...
+    }
+
+    fn get_round2_dkg_package<'a>(&'a mut self, _: Empty) -> Pin<Box<dyn Future<Output = Result<DkgPayload, GrpcClientError>> + Send + 'a>> {
+        Box::pin(async move {
+            let added_round1 = self.added_round1_packages.read().await;
+            let peer_idx = self.peer_id.parse::<usize>().unwrap_or(0);
+            let test_data = &self.test_vectors[peer_idx];
+
+            // Need packages from all participants before returning round2
+            if added_round1.len() < 2 {
```
arminsabouri: Should this be min_signers. Might be best to create two new props for this struct. 
min and max signers.

crates/consensus/authority/src/test_utils.rs
```diff
@@ -0,0 +1,606 @@
...
+    pub public_key_package: PublicKeyPackage,
+    pub state: Arc<RwLock<BTreeMap<[u8; 32], SessionState>>>,
+    pub dkg_state: Arc<RwLock<DKGState>>,
+    pub coordinator_map: Arc<RwLock<BTreeMap<[u8; 32], u64>>>, // Track coordinator for each session
+}
+
+impl MockBtcServerClient {
+    pub fn new(min_signers: u16, max_signers: u16, peer_id: String) -> Self {
+        let test_vectors = generate_test_vectors(max_signers, min_signers);
+        let peer_idx = peer_id.parse::<usize>().unwrap_or(0);
```
arminsabouri: I would just unwrap here as it would mean a test is using a invalid peer id param

crates/consensus/authority/src/test_utils.rs
```diff
@@ -0,0 +1,606 @@
...
+            // Check if we have enough round1 signing packages
+            let added_round1 = self.added_round1_signing_packages.read().await;
+            if added_round1.len() < 2 {
+                return Err(GrpcClientError::Call(tonic::Status::internal("Not enough round1 signing packages")));
+            }
+            
+            Ok(SigningPackage {
+                identifier: test_data.identifier.serialize().to_vec(),
+                signing_session_id: request.signing_session_id,
+                psbt: added_round1.values().next().unwrap().clone(),
```
arminsabouri: can you explain this? can we just return empty bytes here since signers are signing the session id?

crates/consensus/authority/src/test_utils.rs
```diff
@@ -0,0 +1,606 @@
...
+/// Mock BTC server client for testing
+#[derive(Clone)]
+pub struct MockBtcServerClient {
+    pub peer_id: String,
+    pub test_vectors: Vec<TestData>,
+    pub added_round1_packages: Arc<RwLock<BTreeMap<Vec<u8>, Vec<u8>>>>,
+    pub added_round2_packages: Arc<RwLock<BTreeMap<Vec<u8>, Vec<u8>>>>,
+    pub added_round1_signing_packages: Arc<RwLock<BTreeMap<Vec<u8>, Vec<u8>>>>,
+    pub added_round2_signing_packages: Arc<RwLock<BTreeMap<Vec<u8>, Vec<u8>>>>,
+    pub public_key_package: PublicKeyPackage,
```
arminsabouri: Should this be optional? In the case where we want to emulate DKG has not been done yet

crates/consensus/authority/src/test_utils.rs
```diff
@@ -0,0 +1,606 @@
...
+    pub state: Arc<RwLock<BTreeMap<[u8; 32], SessionState>>>,
+    pub dkg_state: Arc<RwLock<DKGState>>,
+    pub coordinator_map: Arc<RwLock<BTreeMap<[u8; 32], u64>>>, // Track coordinator for each session
+}
+
+impl MockBtcServerClient {
+    pub fn new(min_signers: u16, max_signers: u16, peer_id: String) -> Self {
+        let test_vectors = generate_test_vectors(max_signers, min_signers);
```
arminsabouri, line 151: Perhaps the test vectors should be passed in. That way we can emulate different signing server / application behaviors. E.g a case one signer refuses to sign or a signers signs an entirely different message. Having the test vectors inside new seems to pre-define the mock for one exact usecase.

arminsabouri (commented):
All just nits. Over all this is progressing nicely. To emulate different application behavior I suppose instead of rewriting a specific grpc response we can feed in different test data? What do you think? 

For example the case where we want to error out of `get_public_key`. With this paradigm its very cumbersome to write a new mock but we can create a version of MockBtcServerClient that has `None` for the publickey package. Then the god mock will just error out. Something like that could work

SatwikPrabhu:
> All just nits. Over all this is progressing nicely. To emulate different application behavior I suppose instead of rewriting a specific grpc response we can feed in different test data? What do you think?
> 
> For example the case where we want to error out of `get_public_key`. With this paradigm its very cumbersome to write a new mock but we can create a version of MockBtcServerClient that has `None` for the publickey package. Then the god mock will just error out. Something like that could work

That's what I had in mind as well. Thanks for elaborating it out. Basically, we can add parameters to `generate_test_vectors` to customize what kind of error scenario we want to test. Its just a question of the number and type of scenarios we want to test. I'm fairly confident that we can tweak `generate_test_vectors` to cover the things we have in mind. I feel that this can be done as part of the PR's that actually add unit tests. What do you feel?

crates/consensus/authority/src/test_utils.rs
SatwikPrabhu: That is indeed a way. I was having the idea that we can further parameterize `generate_test_vectors` as needed. Not sure which would be more cumbersome.

bin/btc-server/src/extended_client.rs
rwlockbg: @SatwikPrabhu I would rather rename the BtcServerExtendedClient to BtcServerExtendedApi and keep the BtcServerClient (Rather than using BtcServerExtendedClientImpl) - this is the rusty way!

crates/consensus/authority/src/test_utils.rs
```diff
@@ -0,0 +1,565 @@
...
+    pub min_signers: u16,
+    pub test_vectors: Vec<TestData>,
+    pub added_round1_packages: Arc<RwLock<BTreeMap<Vec<u8>, Vec<u8>>>>,
+    pub added_round2_packages: Arc<RwLock<BTreeMap<Vec<u8>, Vec<u8>>>>,
+    pub added_round1_signing_packages: Arc<RwLock<BTreeMap<Vec<u8>, Vec<u8>>>>,
+    pub added_round2_signing_packages: Arc<RwLock<BTreeMap<Vec<u8>, Vec<u8>>>>,
+    pub public_key_package: Option<PublicKeyPackage>,
+    pub state: Arc<RwLock<BTreeMap<[u8; 32], SessionState>>>,
+    pub dkg_state: Arc<RwLock<DKGState>>,
+    pub coordinator_map: Arc<RwLock<BTreeMap<[u8; 32], u64>>>, // Track coordinator for each session
```
rwlockbg: I wonder if we need to have all these mutexes here ? What if we rather keep the client linear and instead whenever used in tests have `Arc<RwLock<Client>>`, this would be much easier in terms of interface usage and practicability. The client itself is most likely meant to be used in state machine contexts where actions are applied one at a time. I see little need of having all these RwLocks for every member tbh.

crates/consensus/authority/src/test_utils.rs
rwlockbg: What is the idea here of returning a pinned future ? This is normally the case of a streaming api such as the grpc2 stream one. All these here are unary calls. Or are you rather just implementing the current grpc interface of the server ?

rwlockbg:
> All just nits. Over all this is progressing nicely. To emulate different application behavior I suppose instead of rewriting a specific grpc response we can feed in different test data? What do you think?
> 
> For example the case where we want to error out of `get_public_key`. With this paradigm its very cumbersome to write a new mock but we can create a version of MockBtcServerClient that has `None` for the publickey package. Then the god mock will just error out. Something like that could work

I had the exact same idea and observation. Armins idea is indeed what we might need overall - mocking the client is hard and it might not always work well.

crates/consensus/authority/src/test_utils.rs
```diff
@@ -0,0 +1,565 @@
...
+            &test_data.key_package,
+        );
+
+        test_data.signing_round1_nonces = Some(nonces);
+        test_data.signing_round1_commitments = Some(commitments);
+        Ok(())
+    }
+
+    /// Helper method to generate round2 signing packages
+    pub fn generate_round2_signing(
```
rwlockbg, line 186: Since you are returning result with an Error, lets remove all sorts of expects and unwraps and bubble up errors to the caller ?

bin/btc-server/src/extended_client.rs
rwlockbg: You can just use this: https://crates.io/crates/grpcmock Its already done for you

SatwikPrabhu: Will check this out. Thanks!

bin/btc-server/src/extended_client.rs
SatwikPrabhu: Done!

crates/consensus/authority/src/builder.rs
SatwikPrabhu: Done!

crates/consensus/authority/src/comet_bft/abci.rs
SatwikPrabhu: Done!

bin/btc-server/src/extended_client.rs
SatwikPrabhu: Yep. It was not used. Removed now.

crates/consensus/authority/src/test_utils.rs
SatwikPrabhu: The reason was so that we can use the trait directly everywhere that we previously used the struct. In some cases, this was not possible with all the async functions in the trait and thus I had to remove the async and do the pinned future way

crates/consensus/authority/src/test_utils.rs
SatwikPrabhu: Changed it now.

crates/consensus/authority/src/test_utils.rs
SatwikPrabhu: Updated.

crates/consensus/authority/src/test_utils.rs
SatwikPrabhu: I've removed it from the mock, since we do not actually need to sign for the mock I guess

crates/consensus/authority/src/test_utils.rs
SatwikPrabhu: Done!

crates/consensus/authority/src/test_utils.rs
SatwikPrabhu: Updated.

crates/consensus/authority/src/test_utils.rs
SatwikPrabhu: Yep. I updated according to the suggestion. Thanks!

crates/consensus/authority/src/test_utils.rs
SatwikPrabhu: Done!

crates/consensus/authority/src/test_utils.rs
SatwikPrabhu: Yeah. I meant to simplify it earlier. Ive updated it now.

crates/consensus/authority/src/test_utils.rs
SatwikPrabhu: That makes sense. Will update it.

crates/consensus/authority/src/test_utils.rs
SatwikPrabhu: Updated. Thanks!

crates/consensus/authority/src/test_utils.rs
SatwikPrabhu: Updated now.

crates/consensus/authority/src/test_utils.rs
arminsabouri: Yeah ideally we pass in some closure or have a way of affecting the behavior of finalize signing

arminsabouri (approved):

rwlockbg (approved):

PR #524, opened 2024-12-27
Feature/btc server track mempool drops

scottmillner:
ready for review

I've added an int test but not to the `test_runner` file. We need to refactor the test-suite a bit to get the functionality I need in the test.

Here is the issue that describes this: https://github.com/botanix-labs/botanix/issues/877

### DIFF
--- a/bin/btc-server/src/pegout_scheduler.rs
+++ b/bin/btc-server/src/pegout_scheduler.rs
@@ -819,7 +819,7 @@ mod tests {
         frost_id,
         test_utils::test_utils::{
             create_block, create_random_pegout_id, create_tx, pegout_requests_from_tx,
-            random_p2wpkh_script, setup_db, trusted_dealer_setup,
+            random_p2wpkh_script, setup_db, trusted_dealer_setup, MockBitcoind,
         },
     };
 
@@ -830,7 +830,7 @@ mod tests {
 
     #[test]
     fn tracked_tx_utils() {
-        let tx = create_tx(0, 0, None);
+        let tx = create_tx(0, 0, None, false);
         let tx = Tx {
             txid: tx.txid(),
             tx,
@@ -845,7 +845,7 @@ mod tests {
         assert_eq!(tx.change().count(), 0);
 
         // 5 inputs, 2 outputs
-        let dummy_tx = create_tx(5, 2, None);
+        let dummy_tx = create_tx(5, 2, None, false);
         assert_eq!(dummy_tx.input.len(), 5);
         assert_eq!(dummy_tx.output.len(), 2);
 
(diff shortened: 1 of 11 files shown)


bin/btc-server/src/pegout_scheduler.rs
```diff
@@ -599,6 +599,62 @@ impl PegoutScheduler {
...
+                    // we've already checked if the tx is in the finalized outputs list
+                    // so it has been dropped or there was a reorg
+                    warn!("Tx {} not in the mempool or finalized outputs list: {}", txid, e);
+                    // 4) untrack the tx
+                    self.un_track_tx(&txid)?;
+                    // 5) add the pegout back to the pending pegout set
+                    self.add_tx_back_to_pending(&tx)?;
+                }
+            }
+        }
```
SatwikPrabhu, line 653: Maybe we can make the call of `add_tx_back_to_pending` before the match block. Since it is done in both ok and err cases.

bin/btc-server/src/test_utils.rs
```diff
@@ -168,6 +199,15 @@ pub mod test_utils {
...
+    pub fn deterministic_p2wpkh_script() -> ScriptBuf {
+        let sk =
+            bitcoin::PrivateKey::from_wif("cV4G8b983VToX9qL5u82qKVNkVEMs3F3gSdf3s1ormG5S5vi2Gi6")
+                .unwrap();
+        let pk = sk.public_key(SECP256K1);
+        let spk = Address::p2wpkh(&pk, NETWORK).unwrap().script_pubkey();
+
+        spk
+    }
     pub fn setup() -> App<MockBitcoind> {
```
SatwikPrabhu, line 211:
```suggestion
    pub fn deterministic_p2wpkh_script() -> ScriptBuf {
        let sk =
            bitcoin::PrivateKey::from_wif("cV4G8b983VToX9qL5u82qKVNkVEMs3F3gSdf3s1ormG5S5vi2Gi6")
                .unwrap();
        let pk = sk.public_key(SECP256K1);
        let spk = Address::p2wpkh(&pk, NETWORK).unwrap().script_pubkey();

        spk
    }
    
    pub fn setup() -> App<MockBitcoind> {
```
Nit

bin/test-suite/src/suite/consensus/frost/test_track_mempool.rs
```diff
@@ -0,0 +1,217 @@
...
+        PegoutId::from_bytes(&pegout_id_bytes).map_err(|_| anyhow::anyhow!("invalid pegout id"))?;
+
+    let secp = bitcoin::secp256k1::Secp256k1::new();
+    let sk = bitcoin::PrivateKey::generate(bitcoin::Network::Regtest);
+    let pk = sk.public_key(&secp);
+
+    let spk = bitcoin::Address::p2wpkh(&pk, bitcoin::Network::Regtest)?.script_pubkey();
+
+    // Notify there is a pending pegout
+    let amount = bitcoin::Amount::from_sat(100_000);
```
SatwikPrabhu, line 148: We can reuse an earlier variable `amount_to_send` since the value is the same.

scottmillner:
closing and will open new pr with changes: easier to do this than rebase

PR #527, opened 2025-01-02
feature(btc-server): notifying pending pegouts should be a list

arminsabouri:
https://github.com/botanix-labs/botanix/issues/620


bin/btc-server/proto/btc_server.proto
```diff
@@ -79,6 +79,10 @@ message NotifyPegoutRequest {
     uint64 height = 4;
```
SatwikPrabhu: We probably need to remove the above `NotifyPegoutRequest` now.

SatwikPrabhu (approved):

PR #537, opened 2025-01-07
773 fixnet frost messages should authenticate coord based on peer id not id passed in payload

arminsabouri:
I was actually mistaken about the original desc. of this ticket. https://github.com/botanix-labs/botanix/issues/773
The peer message is provided in `PeerMessage` by the frost protocol manager. The original peer id is provided in `in_coming` which comes from the network manager. meaning a faulty federation member cannot just inject invalid peer ids. HOWEVER, they can provide whatever frost ids they want. 

I would want to remove the frost id as something the peer can provide in a message and calculate it on the fly.
We have a mapping of peer_id -> auth_pk (sorted list). So we should be able to get the index of the authority pk and calculate the frost id from the index. 

@rwlockbg @scottmillner  Does this make sense to you guys?

### DIFF
--- a/bin/btc-server/README.md
+++ b/bin/btc-server/README.md
@@ -2,18 +2,18 @@
 
 Bitcoin signer service that interacts with a database and performs transaction signing and related operations. Below is an overview of the main components and functionality provided by this code.
 
-### Building and Running
+## Building and Running
 
-    Ensure you have Rust and Cargo installed on your system.
+Ensure you have Rust and Cargo installed on your system.
 
-    Run `make start-btc-server-1` from the project root directory
+Run `make start-btc-server-1` from the project root directory
 
 
-    To just generate the client. run `cargo build`.
+To just generate the client. run `cargo build --features conflicting_input`.
 
-    Take a look at the integration test suite in `bin/test-suite` for examples of using the client.
+Take a look at the integration test suite in `bin/test-suite` for examples of using the client.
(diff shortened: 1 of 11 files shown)


rwlockbg (approved):
Nice! Makes total sense to me to use the peer_id ins. of the frost_id. Nevertheless, I dont see  a scenario where one could connect with us who is not a fed member provided we have blocked all outgoing connections and are preloading the fed members static list. But the more checks, obv. the better.

scottmillner:
> I was actually mistaken about the original desc. of this ticket. [botanix-labs/botanix#773](https://github.com/botanix-labs/botanix/issues/773) The peer message is provided in `PeerMessage` by the frost protocol manager. The original peer id is provided in `in_coming` which comes from the network manager. meaning a faulty federation member cannot just inject invalid peer ids. HOWEVER, they can provide whatever frost ids they want.
> 
> I would want to remove the frost id as something the peer can provide in a message and calculate it on the fly. We have a mapping of peer_id -> auth_pk (sorted list). So we should be able to get the index of the authority pk and calculate the frost id from the index.
> 
> @rwlockbg @scottmillner Does this make sense to you guys?

ack that: calculate frost id on the fly instead of a peer passing it in a message

scottmillner (approved):

arminsabouri:
ALL GREEN!

PR #538, opened 2025-01-07
Chore: Appropriate error for botanix consensus pkg

SatwikPrabhu:
Previously, we did not return an appropriate error when we ran `botanix_consensus_package`


crates/blockchain-tree-api/src/error.rs
```diff
@@ -482,7 +488,7 @@ impl InsertBlockErrorKind {
                         true
                     }
                     // these are internal errors, not caused by an invalid block
-                    BlockExecutionError::Internal(_) => false,
+                    BlockExecutionError::Internal(_) | BlockExecutionError::BotanixConsensusPkgError(_) => false,
```
arminsabouri: "Returns true if the error is caused by an invalid block"
Shouldn't this be true?

crates/evm/execution-errors/src/lib.rs
```diff
@@ -190,6 +191,8 @@ pub enum BlockExecutionError {
     Consensus(ConsensusError),
     /// Internal, i.e. non consensus or validation related Block Executor Errors
     Internal(InternalBlockExecutionError),
+    /// Botanix Consensus Package Error, transparently wrapping [`BotanixConsensusPackageError`]
+    BotanixConsensusPkgError(BotanixConsensusPackageError),
```
arminsabouri: Does this need a root level error? It might make more sense to add a new variant of `BlockValidationError` to include BotanixConsensusPackageError

arminsabouri (changes requested):

arminsabouri:
@SatwikPrabhu take a look at #539. I add a new error variant to block validation here. It would look something like this

crates/blockchain-tree-api/src/error.rs
SatwikPrabhu: I thought that this was just a package error. Was actually not sure whether this had any implications wrt to the validity of a block.

crates/evm/execution-errors/src/lib.rs
SatwikPrabhu: Done. Thanks for the suggestion

crates/evm/execution-errors/src/lib.rs
```diff
@@ -162,6 +162,10 @@ pub enum BlockValidationError {
     /// [EIP-6110]: https://eips.ethereum.org/EIPS/eip-6110
     #[display("failed to decode deposit requests from receipts: {_0}")]
     DepositRequestDecode(String),
+
+    /// Error when creating a Botanix consensus package
+    #[display("Failed to construct Botanix Consensus Pkg")]
+    BotanixConsensusPkgError(String),
```
arminsabouri: Nit: can we pass down the actual error? not just the string representation?

scottmillner (dismissed):

arminsabouri:
@SatwikPrabhu I also notcied this error variant https://github.com/botanix-labs/botanix/issues/870
Is this used anywhere if not can we remove it in this pr

SatwikPrabhu:
> @SatwikPrabhu I also notcied this error variant https://github.com/botanix-labs/botanix/issues/870
Is this used anywhere if not can we remove it in this pr

Yes. Its actually not used anywhere, Can remove it.

crates/evm/execution-errors/src/lib.rs
SatwikPrabhu: Made the change. Please let me know if it looks fine.

arminsabouri:
@SatwikPrabhu it seems like the crates check job is failing. Can you please take a look?

arminsabouri:
@SatwikPrabhu this needs a rebase off main

SatwikPrabhu:
@0xBEEFCAF3 I have rebased and fixed the check crates error. I guess its good to go now.

crates/primitives/src/header_ext.rs
```diff
@@ -63,6 +63,22 @@ pub enum BotanixConsensusPackageError {
     FailedToRetrieveBitcoinCheckpointHeight(BitcoindError),
 }
```
scottmillner, line 65: why do we need to implement Clone, PartialEq, and Eq for an error?

SatwikPrabhu: I had to do it because `BlockValidationError` derives Clone, PartialEq and Eq. And deriving these for `BotanixConsensusPackageError` is not possible as these are not implemented for the error types included with its variants. 
Maybe you have a cleaner solution for this?

scottmillner (approved):

PR #540, opened 2025-01-08
ci(tests): fix ci pipeline and tests

scottmillner:


bin/test-suite/src/suite/consensus/common/poa_node.rs
```diff
@@ -299,7 +299,13 @@ impl FederationMemberTestConfig {
 
         // prepare run arguments
         let command = "./target/debug/reth";
+        // TODO(scott): make features flag dynamic
+        // it can be passed in via cli and does default to the feature below but
+        // needs to be passed down to this function
         let args = vec![
+            "--features",
```
rwlockbg: Should not this flag be passed further down, after we have the -- ?

arminsabouri: Im not sure we need this flag to begin with

scottmillner: yep, passing flag in integration workflow to enable the feature

and the flag needs to be passed before "--" since it's a build arg and not used by the binary

rwlockbg (dismissed):

bin/btc-server/src/coordinator.rs
```diff
@@ -175,18 +175,15 @@ where
     ) -> Result<(PublicKey, PublicKey, Address), CoordinatorError> {
         // try to get pk package from db in case we already did dkg round 3
         if let Some(pk_package) = self.db.get_public_key_package()? {
-            let agg_key = pk_package
-                .verifying_key()
-                .to_secp_pk()
-                .map_err(CoordinatorError::FailedToConvertVerifyingKeyToSecpPk)?;
```
SatwikPrabhu, line 181: Maybe we can remove this error variant since its not being used anymore?

arminsabouri: This is a fair comment. If its ok with you I might fix this in a follow up PR with some further enhancments

arminsabouri (approved):

PR #544, opened 2025-01-13
fix(abci): handle errors when configuring payload builder

scottmillner:
If we have an error when building the payload builder arguments, I'm returning an error and `prepare_proposal` will be return the default response

Why?
- `PayloadAttributes.parent_beacon_block_root` is needed for EIP-4788 and needs to be Some(value). I'd rather pull it from the parent block then pass an arbitrary value since i don't know the potential impact of this
- And errors would occur when the provider can't read from the db so i just want to return an empty response (default response) to basically short circuit


arminsabouri (approved):
CodeACK!

rwlockbg (approved):

PR #545, opened 2025-01-16
fix(abci): remove insert_block() and FCU

scottmillner:
~~After our discussions and tracing the code here is my approach and why:~~

~~In the abci driver when can remove `insert_block` and the `FCU` update.~~

~~Why?~~  
~~- `insert_block` is needed for fork logic which we don't need. And the validation it runs has already been done in `process_proposal` by `build_and_execute`~~  
~~- FCU makes the block canonical but also does fork choice logic checks that we don't need to run.  The functionality we need from the FCU is `make_canonical()` and updating the `canon chain tracker` which is used by RPC endpoints~~

~~What are the changes?~~  
~~- remove `insert_block`~~  
~~- remove `FCU` message send~~  
~~- add `make_canonical`~~  
~~- keep the already existing canon chain tracker updates (this was redundant before since FCU did this)~~

~~Additional Notes:~~  
~~- I think we should still use the BTE `make_canonical()` instead of creating out own bespoke method~~

~~Why?~~  
~~- Less changes and unforeseen impacts~~  
~~- The DB passed to the ABCI driver is the `BlockchainProvider` which already has the BTE trait implemented so it's not like we had to do extra work to add that functionality~~

**What was actually done:**
ABCI driver was refactored to remove components we don't need and this also resolves the bitcoind call issue since we don't make that call any more when the driver commits a block

- Blockchain Tree Engine removed
- Beacon Engine removed
- The functionality provided by the above that we need is now done manually (ie we insert a block and update the in-memory canon state cache)


crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -880,16 +886,10 @@ where
...
-                            sealed_block_with_senders.clone(),
-                            BlockValidationKind::Exhaustive,
-                        ) {
-                            Ok(_) => {}
-                            Err(e) => {
-                                error!(target: "consensus::authority", ?e, "Failed to insert block");
-                                return Err(Box::new(e));
-                            }
+                        // Update the canonical chain
+                        if let Err(e) = client.make_canonical(block_hash) {
```
arminsabouri: More of a question than review. Do we not need to insert the block somehow first before making it canonical. I'm not sure how the DB will be updated here

arminsabouri (commented):
Thanks for updating the docs and excellent pr description. Just had one question

crates/consensus/authority/src/comet_bft/abci.rs
scottmillner: good question: the block is inserted in the db during `make_canonical()`:

where?

`blockchain_tree.make_canonical()` calls `commit_canonical_to_database()` which inserts the block into the db

arminsabouri: But do we need a candidate block ready to be made canonnical? Which is what I think `insert_block` does

arminsabouri: I think that is also whats being reported from the int. tests results
https://github.com/botanix-labs/Macbeth/actions/runs/12799812200/job/35709035995?pr=545#step:10:6552

```
Error: Critical task `ABCI Driver` panicked: `ABCI Driver encountered a error: BlockchainTree(BlockHashNotFoundInChain { block_hash: 0x21a6be5e998fd1b748176a2fac4c22d27f3f31d89756498a7509f92220d8b416 })`
```

arminsabouri:
We need to be using the UnifiedStorageWriter here as static files will not be updating.

Makefile
```diff
@@ -591,7 +591,7 @@ start-btc-server-2:
 	--bitcoind-pass "${BITCOIND_PWD}" \
 	--btc-signing-server-jwt-secret "${NODE_2_DIR}/bjwt.hex" \
 	--fall-back-fee-rate-sat-per-vbyte 5 \
-	--metrics-port 7001
+	--metrics-port 7000
```
arminsabouri, line 587: Do we want to commit this? If there are conflicting ports I think we should

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -653,10 +632,21 @@ where
...
                 }
 
+                let block_with_context = BlockWithContext {
+                    sealed_block_with_peg,
+                    exec_outcome: ExecutionOutcome::new(
+                        exec_results.state,
+                        exec_results.receipts.into(),
+                        block_number,
+                        // TODO: pull requests from the block exec outcome
+                        vec![],
```
arminsabouri: Can this be left as a TODO. I think so b/c withdrawals should always be empty vec

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -780,141 +780,152 @@ where
...
-                        client.set_finalized(sealed_block_with_senders.header.clone());
+                    ABCIDriverMessage::CommitBlock(sealed_block_with_context) => {
+                        let consistent_db_view =
+                            ConsistentDbView::new_with_latest_tip(self.database_provider.clone())?;
+                        let hashed_state = sealed_block_with_context.exec_outcome.hash_state_slow();
+                        let (_, trie_updates) =
+                            ParallelStateRoot::new(consistent_db_view, hashed_state.clone())
+                                .incremental_root_with_updates()
+                                .map(|(root, updates)| (root, Some(updates)))
+                                .map_err(ProviderError::from)?;
```
arminsabouri: My understanding is that this is a computationally expensive operation. Is there anyway to pass this down from `build_and_execute()` ?

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -780,141 +780,152 @@ where
...
+                            new: vec![executed_block],
+                        };
+                        self.blockchain_provider_2
+                            .canonical_in_memory_state()
+                            .update_chain(new_chain);
+
+                        self.blockchain_provider_2
+                            .on_forkchoice_update_received(&ForkchoiceState::default());
+
+                        // print block number
```
arminsabouri: Nit: remove comment

Makefile
scottmillner: no, that's just for my machine...i have something running on port 7000 so that's why i make that change when running the tests but accidentally pushed it

unless others are having an issue with this?

crates/consensus/authority/src/comet_bft/abci.rs
scottmillner: i remember our conversation about this but was incorrect when i thought `validate_and_execute` was being called during `build_and_execute()` in the abci client

`validate_and_execute` is part of the `AppendableChain` trait which BTE impls and was ultimately being called as a result of an FCU

so although the block header does calculate the `HashedPostState`, the `trie_updates` have not been calculated since nothing has been appended to the chain yet so we have to do this here

crates/consensus/authority/src/comet_bft/abci.rs
scottmillner: yeah and these requests are intended for the CL and since we use CometBFT has no meaning in our context

removed TODO and added comment

crates/consensus/authority/src/comet_bft/abci.rs

rwlockbg (dismissed):
Nice cleanup, great job!

crates/net/network/src/import.rs
```diff
@@ -104,9 +104,9 @@ impl ProofOfAuthorityBlockImport {
     /// NOTE: this is not used currently but if its needed again lets add validation
     fn validate(&self, new_block: &SealedBlock) -> Result<(), ConsensusError> {
         let _header = new_block.header.clone();
-        //validation::validate_header_with_total_difficulty(&header, header.difficulty)?; 
+        //validation::validate_header_with_total_difficulty(&header, header.difficulty)?;
         // check after reabse validation::validate_block_standalone(new_block,
-        // &self.chain_spec)?; 
+        // &self.chain_spec)?;
```
SatwikPrabhu: Should we remove these comments now?

rwlockbg (dismissed):

crates/net/network/src/import.rs
scottmillner: thanks for the comment
im going to leave as is since this code isn't used anywhere

this is related to block import which we no longer use...at some point we could remove this and related code

SatwikPrabhu (dismissed):

crates/consensus/authority/src/comet_bft/abci.rs
arminsabouri: It still makes sense to me front load this work to process/finalize as there are error to be handled. Right now it looks like errrors will be silent. So commit wil be handled but the block is never appended. Is that correct?

Makefile
arminsabouri: works for me

crates/consensus/authority/src/comet_bft/abci.rs
scottmillner: the error wouldn't be silent: `start()` would exit with the error returned from `map_err`

but yes, `commit` would be successful but the block would never be appended

ill move this up and add trie updates to the cache

arminsabouri (approved):
[tACK](6beb59f)
Function tests passing :heavy_check_mark: .

rwlockbg (approved):

PR #554, opened 2025-01-28
fix(abci): log error when notifying pegins/pegouts fails

scottmillner:
This is just a band-aid so failed btc-server calls don't cause a chain halt

There's more work coming around the btc-server syncing on a comet checkpoint (app hash) so node can loop through blocks and update pegins/pegouts accordingly.

### DIFF
--- a/crates/consensus/authority/src/comet_bft/abci.rs
+++ b/crates/consensus/authority/src/comet_bft/abci.rs
@@ -309,7 +309,9 @@ where
         ))
     }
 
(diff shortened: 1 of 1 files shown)


scottmillner:
@birozuru can you set up a regex match on the error logs in this pr and pipe those to slack so we are aware this is happening? pretty please 😄 🙏 

we'll eventually add code so the node can self heal but that might take a while

arminsabouri (approved):

rwlockbg (approved):

PR #558, opened 2025-01-30
Fix(evm): Check block address before execution

SatwikPrabhu:
Fix for https://github.com/botanix-labs/botanix/issues/872

### DIFF
--- a/crates/consensus/authority/src/comet_bft/abci.rs
+++ b/crates/consensus/authority/src/comet_bft/abci.rs
@@ -22,8 +22,7 @@ use reth_evm::execute::BlockExecutorProvider;
 
 use reth_payload_builder::EthPayloadBuilderAttributes;
 use reth_primitives::{
-    botanix::block_with_peg::SealedBlockWithPeg, header_ext::HeaderExt, Address, BlockHash,
-    SealedBlock, TransactionSigned,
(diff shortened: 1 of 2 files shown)


scottmillner (dismissed):

crates/consensus/authority/src/excecution_utils.rs
```diff
@@ -58,6 +59,17 @@ pub(crate) mod authority_execution_utils {
...
+        if let Ok(edh_producer) = block.header.block_producer_address() {
+            if edh_producer != *block_builder_address {
+                warn!("Block producer address mismatch. EDH: {:?}, Builder: {:?}", edh_producer, block_builder_address);
+                return Err(BlockExecutionError::Validation(BlockValidationError::BlockAddressMismatch { 
+                    cometbft_address: *block_builder_address, 
+                    extra_data_address: edh_producer 
+                }));
+            }
+        }
+
```
arminsabouri: The check here is correct but the placment is not. Here you are checking if the block you created has the block builder address you intended. Not if the edh block builder address matches what comet says it should be

arminsabouri (changes requested):

crates/consensus/authority/src/excecution_utils.rs
rwlockbg: agree

arminsabouri:
Apologies this seems to be a non-issue. The block producer is determinstic data we get from cbft. If one block producer decided to change the block producer they would just fork off the network.

PR #559, opened 2025-01-31
Refactor(btc-server): Make telemetry optional

SatwikPrabhu:
Fix for https://github.com/botanix-labs/botanix/issues/898

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -565,8 +565,7 @@ start-btc-server-1:
 	--bitcoind-user "${BITCOIND_USER}" \
 	--bitcoind-pass "${BITCOIND_PWD}" \
 	--btc-signing-server-jwt-secret "${NODE_1_DIR}/bjwt.hex" \
-	--fall-back-fee-rate-sat-per-vbyte 5 \
-	--metrics-port 7000
+	--fall-back-fee-rate-sat-per-vbyte 5
 
 start-btc-server-2:
 	cd ./bin/btc-server && \
@@ -583,8 +582,7 @@ start-btc-server-2:
 	--bitcoind-user "${BITCOIND_USER}" \
 	--bitcoind-pass "${BITCOIND_PWD}" \
 	--btc-signing-server-jwt-secret "${NODE_2_DIR}/bjwt.hex" \
-	--fall-back-fee-rate-sat-per-vbyte 5 \
-	--metrics-port 7000
+	--fall-back-fee-rate-sat-per-vbyte 5
 
 start-btc-server-3:
 	cd ./bin/btc-server && \
@@ -601,8 +599,7 @@ start-btc-server-3:
 	--bitcoind-user "${BITCOIND_USER}" \
 	--bitcoind-pass "${BITCOIND_PWD}" \
 	--btc-signing-server-jwt-secret "${NODE_3_DIR}/bjwt.hex" \
-	--fall-back-fee-rate-sat-per-vbyte 5 \
-	--metrics-port 7002
+	--fall-back-fee-rate-sat-per-vbyte 5
 
 start-poa-server-1:
 	cd ./bin/reth && \
--- a/bin/btc-server/src/bin/main.rs
+++ b/bin/btc-server/src/bin/main.rs
@@ -1425,10 +1425,15 @@ async fn main() -> anyhow::Result<(), Box<dyn std::error::Error>> {
         .filter_module("bitcoincore_rpc::", log::LevelFilter::Trace)
         .init();
 
-    let telemetry = Telemetry::new().await?;
-    telemetry.start().await?;
-
     let config = btcserverlib::config::load_config()?;
+
+    let telemetry = if config.metrics_port.is_some() {
+        let telemetry = Telemetry::new().await?;
+        telemetry.start().await?;
+        Some(telemetry)
+    } else {
+        None
+    };
 
     // setup the grpc server
     let bitcoind_client = bitcoincore_rpc::Client::new(
@@ -1437,7 +1442,7 @@ async fn main() -> anyhow::Result<(), Box<dyn std::error::Error>> {
     )
     .expect("bitcoind client");
     let btc_server: App<bitcoincore_rpc::Client> =
-        App::new(config.clone(), bitcoind_client, Some(telemetry.clone()))?;
+        App::new(config.clone(), bitcoind_client, telemetry.clone())?;
 
     // run grpc server in the background
     let grpc_stop_tx = match btc_server.serve_async().await {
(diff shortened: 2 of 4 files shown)


scottmillner (dismissed):

Makefile
```diff
@@ -568,6 +568,23 @@ start-btc-server-1:
 	--fall-back-fee-rate-sat-per-vbyte 5 \
 	--metrics-port 7000
 
+start-btc-server-1-local:
```
arminsabouri: Do we need new commands in our make file? I would just remove telemetry from the existing commands

bin/btc-server/src/bin/main.rs
```diff
@@ -1424,19 +1406,24 @@ async fn main() -> anyhow::Result<(), Box<dyn std::error::Error>> {
         .filter_module("bitcoincore_rpc::", log::LevelFilter::Trace)
         .init();
 
-    let telemetry = Telemetry::new().await?;
-    telemetry.start().await?;
-
     let config = btcserverlib::config::load_config()?;
 
+    let telemetry = if config.enable_metrics {
```
arminsabouri: Instead of adding a new enable metrics flag how about we make the metrics port flag optional. And that can be the indicator for wether we start telemetry or not

bin/btc-server/src/bin/main.rs
```diff
@@ -16,28 +16,10 @@ use bitcoin_hashes::Hash;
 use bitcoincore_rpc::{Auth, RpcApi};
 use btc_server::btc_server_server::{BtcServer, BtcServerServer};
 use btcserverlib::{
-    config::{Config, Error as ConfigError},
-    coordinator,
```
arminsabouri: Are these formatting changes a results of running `make fmt`? seem unrelated to the changes in this PR

arminsabouri (changes requested):

Makefile
SatwikPrabhu: I just thought that it could be useful to have both versions. Will remove it. Thanks

bin/btc-server/src/bin/main.rs
SatwikPrabhu: I initially tried that. But the metrics port is needed for initializing socket addresses etc. So it can't be optional. Maybe I can explore further if there is still a way to make it optional.

bin/btc-server/src/bin/main.rs
SatwikPrabhu: Sorry about that. Yeah these changes are unrelated. Will fix them.

Makefile
rwlockbg: No new commands, pleasssse

bin/btc-server/src/bin/main.rs
rwlockbg: Why dont you try making Option<u16> for the telemetry port in hte cli args. If its Some, then you start, if not, you dont.

rwlockbg:
@SatwikPrabhu Whatever change you make to the input args, pls check the int.test suite and in particular the poa_node file and the rpc_nod files in there. There is a place where we are passing the args -- to the spawned process. These args should also be upated there.

bin/btc-server/src/bin/main.rs
SatwikPrabhu: I think I managed to do this. Lemme know if this looks good

SatwikPrabhu:
> @SatwikPrabhu Whatever change you make to the input args, pls check the int.test suite and in particular the poa_node file and the rpc_nod files in there. There is a place where we are passing the args -- to the spawned process. These args should also be upated there.

I dont think poa_node and rpc_node files have metrics port as input. So no changes there. But I changed the part for the btc server int tests.

bin/test-suite/src/context/mod.rs
```diff
@@ -27,6 +27,7 @@ pub struct GlobalContext {
     pub bitcoind_pass: String,
     pub botanix_fee_recipient: String,
     pub features: String, // space delimited String
+    pub enable_metrics: bool,
```
arminsabouri: int. tests generally dont need metrics enables. So we can always exlucde `--metrics X` from the btc server start commands

bin/btc-server/src/bin/main.rs
```diff
@@ -1457,8 +1458,12 @@ async fn main() -> anyhow::Result<(), Box<dyn std::error::Error>> {
     // create and spin up the http server
     let state = ServerState::new(telemetry.clone()).await;
     // create the actix webserver
+    let port = match config.metrics_port {
```
arminsabouri: Nit: can we use `unwrap_or()` here?

bin/btc-server/src/http/state.rs
```diff
@@ -14,13 +14,13 @@ pub struct HealthResponse {
 
 #[derive(Clone)]
 pub struct ServerState {
-    pub telemetry: Arc<Telemetry>,
+    pub telemetry: Option<Arc<Telemetry>>,
```
arminsabouri: IIRC serverstate is only being used in conjunction with telemetry. So I think it makes sense for this not to be optional here as an instance of  Server state should not exist without telemetry

bin/btc-server/src/telemetry/mod.rs
```diff
@@ -14,10 +14,14 @@ pub struct Telemetry {
 }
 
 impl Telemetry {
-    pub async fn new() -> anyhow::Result<Arc<Self>> {
+    pub async fn new(enable_metrics: bool) -> anyhow::Result<Arc<Self>> {
```
arminsabouri: I think the act of calling Telemetry::new() indicated that we do want to enable metrics. So do we need this boolean flag?

arminsabouri (changes requested):

bin/btc-server/src/http/state.rs
SatwikPrabhu: So, I guess it then makes sense to have an optional `ServerState` in `create_web_server`?

arminsabouri: I dont think any of the metrics or internal telemetry related datastructures should have any optional fields. The fact that they were instantiated means that the user opt'd in to enabling metrics.

SatwikPrabhu: Apologies if my previous comment was unclear. I mean currently `create_web_server` has a non-optional `ServerState` parameter which has the telemetry field. If the telemetry field will be non-optional, then the parameter `state` of `create_web_server` should be made optional right?

arminsabouri: I would say lets keep the top level structs optional not the internal ones. So here it would be `ServerState` that would optional

bin/btc-server/src/bin/main.rs
SatwikPrabhu: Done!

bin/test-suite/src/context/mod.rs
SatwikPrabhu: Done!

bin/btc-server/src/bin/main.rs
SatwikPrabhu: Done!

bin/btc-server/src/http/state.rs
SatwikPrabhu: Done!

bin/btc-server/src/telemetry/mod.rs
SatwikPrabhu: Done!

bin/btc-server/src/bin/main.rs
SatwikPrabhu: Done!

bin/test-suite/src/suite/consensus/common/btc_server.rs
```diff
@@ -68,7 +68,7 @@ fn spawn_btc_server_process(
     let http_port = (BTC_SERVER_HTTP_PORT + id).to_string();
 
     let command = "./target/debug/btc-server";
-    let args = vec![
+    let mut args = vec![
```
scottmillner: nit: doesn't need to be mut

SatwikPrabhu: Done. Thanks!

scottmillner (approved):

bin/btc-server/src/bin/main.rs
```diff
@@ -1455,31 +1460,40 @@ async fn main() -> anyhow::Result<(), Box<dyn std::error::Error>> {
...
+        // spawn the server in the background
+        tokio::spawn(async move {
+            if let Err(err) = grpc_server.await {
+                error!("Actix Web server error: {:?}", err);
+            }
+        });
+        info!("Grpc server started.");
+        Some(server_handle)
+    } else {
+        info!("Telemetry is disabled. Not starting the http server.");
```
rwlockbg, line 1482: @SatwikPrabhu Sorry, what is the logic here - why not start the http server when there are no metrics. There is a healthcheck that we still need for the health status on  k8s and that is part of the server state where the metrics also go in. What I would do is rather  pass optional telemetry via `let state = ServerState::new(telemetry.clone()).await;` and inside the btc-server logic just have some macro that would print metrics if the telemetry is Some. Something like if_metrics!( .. and ehre you call whatever method on the telemetry you want .. ), I feel this might be cleaner, but if the new design is what you have already discussed with the rest of the team, let it be that way too.

SatwikPrabhu: We had a discussion earlier and I think the preference to not have telemetry optional in `ServerState` and I wrote this piece of code based on that. If we really need the health check to happen without telemetry then I think we need to pass some default instance of telemetry when we initialize the state.

bin/btc-server/src/bin/main.rs
```diff
@@ -1455,31 +1460,40 @@ async fn main() -> anyhow::Result<(), Box<dyn std::error::Error>> {
...
-    });
-    info!("Grpc server started.");
+    let server_handle = if let Some(telemetry) = telemetry {
+        // create and spin up the http server
+        let state = ServerState::new(telemetry.clone()).await;
+        // create the actix webserver
+        let port = config.metrics_port.unwrap_or(7000);
+        let grpc_server_addr =
+            SocketAddr::new(IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0)), port);
+        let grpc_server = create_web_server(state, grpc_server_addr)?;
```
arminsabouri, line 1470: Nit: Should be called `metrics_server_addr`

arminsabouri (approved):

bin/btc-server/src/bin/main.rs
rwlockbg: fine, then lets go with it

rwlockbg (approved):

arminsabouri (approved):

PR #560, opened 2025-01-31
fix(repo): Reworked message passing logic with serde traits, fixes to…

rwlockbg:
… chunk apply


bin/test-suite/src/suite/consensus/sync/mod.rs
```diff
@@ -1,5 +1,5 @@
 pub mod error;
 
 // tests
-pub mod test_state_sync;
```
scottmillner: why is this test commented out?

rwlockbg: because the file name got changed and the use statement was failing. I thought you wanted to fix smthg in the test. I will switch it back on.

scottmillner: ah not sure how that happened
ill change the name back and enable the test

SatwikPrabhu (approved):

scottmillner (approved):

PR #566, opened 2025-02-05
fix(repo): Updated crate versions

rwlockbg:


arminsabouri (changes requested):
ApproachNack. Please break this work into two seperate well described commits:
for example `chore: bump thiserror to 2.01`

arminsabouri:
@SatwikPrabhu please review this PR 
seems related to #571

SatwikPrabhu:
@rwlockbg Looks good to me. Like the previous comment by @0xBEEFCAF3 , I think it still needs to be split into separate comments.

rwlockbg:
@SatwikPrabhu @0xBEEFCAF3 added the descriptions to the commit

arminsabouri (approved):
@rwlockbg in the future it would be great to batch related upgrades together in seperate commits instead of a bulk upgrade. The reason being if we need to revert a certain upgrade we can do that without affecting the rest of them. Since this upgrade was clean and had no code changes required I would like to merge it and keep the pipeline moving. Thanks for doing this @rwlockbg

PR #569, opened 2025-02-05
feature(state-sync): add models, tables, and provider to support stat…

scottmillner:
…e sync

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8620,6 +8620,7 @@ dependencies = [
  "reth-storage-errors",
  "reth-trie-common",
  "serde",
+ "sha2 0.10.8",
  "test-fuzz",
 ]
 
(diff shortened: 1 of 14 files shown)


arminsabouri:
e2e passing!

rwlockbg:
This is already contained in the state sync PR. This PR on its own makes no sense, it is either the complete feature or not. The tables and the models merged this way create confusion to whoever is looking at the tree history.

scottmillner:
> This is already contained in the state sync PR. This PR on its own makes no sense, it is either the complete feature or not. The tables and the models merged this way create confusion to whoever is looking at the tree history.

We can all hop on a call tomorrow if we want

But the thinking is...a feature as large as state-sync can be done in incremental prs (this could also have been done on a feature branch then merged into main). This makes it easier to review and reduces the likelihood of bugs being introduced.

One fear is that a pr with so many changes might not be as isolated from critical code as we think

scottmillner:
closing: changes will be made in original pr https://github.com/botanix-labs/Macbeth/pull/506

PR #572, opened 2025-02-05
pull trie updates from exec outcome

arminsabouri:
related ticket: https://github.com/botanix-labs/botanix/issues/896

There is a couple wins here:
* Trie updates is non-optional
* We no longer need to re-calculate the trie root 
* concolidate creation of block with context in exec utils


crates/consensus/authority/src/excecution_utils.rs
```diff
@@ -232,15 +264,11 @@ pub(crate) mod authority_execution_utils {
...
-        let state_root = client
-            .latest()
-            .map_err(|_| {
-                BlockExecutionError::Internal(InternalBlockExecutionError::LatestBlock(
-                    ProviderError::StateForHashNotFound(header.hash_slow()),
-                ))
-            })?
-            .state_root(&block_exec_result.state)
-            .unwrap();
+        let provider = database_provider.provider()?.disable_long_read_transaction_safety();
```
scottmillner: in regards to `disable_long_read_transaction_safety`

what's the benefit of disabling these guarantees?

scottmillner:
🔥 
approach ack

arminsabouri:
e2e passing :heart:

crates/consensus/authority/src/excecution_utils.rs
arminsabouri: Yeah thanks for pointing that. Thats the WIP part of this WIP PR. I dont intend to leave this in. I was looking at what it would take to not get those long read db txs. This didnt work

scottmillner (approved):

PR #573, opened 2025-02-06
Arm/rework chunks

arminsabouri:
The testsuite doesnt compile rn. Given that I have removed assemble_chunks method. But I wanted to break out this work so it can be clearly reviewed and doesnt get lost in the big state sync PR

### DIFF
--- a/crates/consensus/authority/src/snapshot_manager.rs
+++ b/crates/consensus/authority/src/snapshot_manager.rs
@@ -12,7 +12,7 @@ use reth_evm::execute::BlockExecutorProvider;
 use reth_node_core::args::StateSyncArgs;
 use reth_primitives::{BlockNumber, SealedBlockWithSenders};
 use reth_provider::{
-    BlockReaderIdExt, ProviderError, ProviderFactory, SnapshotReader, SnapshotWriter,
+    BlockReaderIdExt, CanonStateNotification, CanonStateSubscriptions, ProviderError, ProviderFactory, SnapshotReader, SnapshotWriter
 };
 use tracing::{debug, error, info, trace, warn};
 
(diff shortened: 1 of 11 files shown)


crates/consensus/authority/src/snapshot_manager.rs
```diff
@@ -5,18 +5,20 @@ use crate::{comet_bft::abci::ABCIDriverMessage, Storage};
...
 use reth_node_core::args::StateSyncArgs;
 use reth_primitives::{BlockNumber, SealedBlockWithSenders};
 use reth_provider::{
-    BlockReaderIdExt, ProviderError, ProviderFactory, SnapshotReader, SnapshotWriter,
+    BlockReaderIdExt, CanonStateNotification, CanonStateSubscriptions, ProviderError,
+    ProviderFactory, SnapshotReader, SnapshotWriter,
 };
 use tracing::{debug, error, info, trace, warn};
 
-const MAX_SNAPSHOT_SIZE_BYTES: usize = 10 * 1024 * 1024; // 10 Mbs
```
scottmillner, line 19: is 500MB from the docs or we have to choose something that makes sense?

crates/storage/db-api/src/models/chunks.rs
```diff
@@ -154,19 +136,9 @@ impl Snapshot {
         self.block_hash = block_hash;
     }
 
-    /// Sets the snapshot hash.
-    pub fn set_hash(&mut self) {
```
scottmillner, line 158: ah yep this makes more sense
don't need to set hash on our custom Snapshot struct, just need to calculate on the fly when creating comet Snapshot for peers

thanks!

scottmillner (approved):

crates/consensus/authority/src/snapshot_manager.rs
```diff
@@ -263,22 +276,41 @@ where
...
-                    let chunk_id = self.create_new_chunk(
-                        last_snapshot_id,
-                        sealed_block.number,
-                        serialized_block_with_context,
-                    )?;
+                    let snapshot =
+                        self.get_snapshot_by_id(last_snapshot_id)?.expect("checked above");
+                    let chunk_id = match snapshot.get_latest_chunk_id() {
+                        Some(chunk_id) => {
+                            // Existing chunk lets append to it
```
scottmillner, line 286: should we check here that appending chunk is within `MAX_SNAPSHOT_CHUNK_SIZE_BYTES`?

PR #574, opened 2025-02-07
Batch insert blocks during state sync

scottmillner:


crates/consensus/authority/src/snapshot_manager.rs
```diff
@@ -283,6 +283,19 @@ where
                         self.get_snapshot_by_id(last_snapshot_id)?.expect("checked above");
                     let chunk_id = match snapshot.get_latest_chunk_id() {
                         Some(chunk_id) => {
+                            // Check if there is enough space in the latest chunk
+                            let latest_chunk_size =
+                                self.provider_factory.provider()?.get_chunk_size(chunk_id)?;
+                            if latest_chunk_size + serialized_block.len() >
+                                MAX_SNAPSHOT_CHUNK_SIZE_BYTES
+                            {
+                                self.create_new_chunk(
```
arminsabouri: Good catch. But I think we need to return after creating a new chunk. Right now we are creating and then appending to the same block

arminsabouri (changes requested):

crates/consensus/authority/src/snapshot_manager.rs
arminsabouri: I'll fix this

scottmillner: yep you're right, thanks!

crates/consensus/authority/src/excecution_utils.rs
```diff
@@ -285,6 +285,40 @@ pub(crate) mod authority_execution_utils {
         Ok(header)
     }
 
+    pub(crate) fn batch_execute<DB, EF>(
```
SatwikPrabhu, line 288: Nit: Would it be better for this method to be named `batch_and_execute()`?. Since we are batching and calling another method to execute

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -904,81 +900,56 @@ where
...
+            }
+        };
 
-        // set canonical head
-        client.set_canonical_head(new_header.clone());
-        client.set_safe(new_header.clone());
-        client.set_finalized(new_header);
+        // TODO: Provide static file provider
+        let mut writer = UnifiedStorageWriter::new(&provider, None);
+        match writer.write_to_storage(exec_outcome, OriginalValuesKnown::Yes) {
```
scottmillner: is there an additional step to move the canonical chain forward?
we're persisting the batched `exec_outcome` but does that update the canonical chain as well?

crates/consensus/authority/src/excecution_utils.rs
arminsabouri: I'm not sure if batch can be used as verb here. Like we're not "batching" blocks. We're are executing a batch of blocks.

crates/consensus/authority/src/comet_bft/abci.rs
arminsabouri: It should... but I wouldnt be suprised if we also need to update some canon fields on the in_memory chain data. That should be obvious if the int. test fails right?

scottmillner: yeah int tests should surface this loudly i'd think

arminsabouri (approved):

PR #575, opened 2025-02-07
feat(abci): Added validate block check to the finalize endpoint

rwlockbg:


SatwikPrabhu (dismissed):

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -737,6 +737,19 @@ where
             }
         };
 
+        if matches!(
+            self.validate_block(&block_with_context.sealed_block_with_peg.block()),
+            ResponseProcessProposal { status: VERIFY_REJECT }
+        ) {
```
scottmillner, line 1361: thanks for adding this!

just wondering if we'd rather panic here?
since we also panic above in the method if `transactions_signed_from_bytes` fails

arminsabouri: The best option is not to panic

arminsabouri: Oh sorry I misunderstood your comment. 
yes we need to figure out these panics at somepoint. Rn its super trivial to crash anyones node via an invalid block. There needs to be a graceful error somewhere before Finalized during block sync

arminsabouri: but yes to follow our current paradigm we need to panic here instead of returning a default response

arminsabouri (changes requested):

crates/consensus/authority/src/comet_bft/abci.rs
rwlockbg: agreed, added.

scottmillner (approved):

arminsabouri (approved):

PR #581, opened 2025-02-11
Clean up bitcoind resources after testrun

arminsabouri:


scottmillner (dismissed):

SatwikPrabhu (dismissed):

arminsabouri:
@SatwikPrabhu and @scottmillner do you guys mind taking another look. Had to push some more

scottmillner:
> @SatwikPrabhu and @scottmillner do you guys mind taking another look. Had to push some more

looks like `Received SIGCHLD ...` is ultimately causing the test-suite to be killed during setup

SatwikPrabhu (dismissed):

arminsabouri:
> > @SatwikPrabhu and @scottmillner do you guys mind taking another look. Had to push some more
> 
> looks like `Received SIGCHLD ...` is ultimately causing the test-suite to be killed during setup

Agh! Looks like we spawn a bunch of child processes to get cometbft information

rwlockbg (commented):
has this been tested @0xBEEFCAF3 ?

arminsabouri:
> has this been tested @0xBEEFCAF3 ?

yes and the test doesnt pass. There SIGCHLD is not the actual solution here for a couple reasons. The best thing to do is actually to `try_wait` on each child process that actually matters and sending the shutdown signal if a child prematurely exits

bin/test-suite/src/suite/consensus/common/bitcoind_node.rs
```diff
@@ -94,48 +102,26 @@ impl BitcoindNodeConfig {
...
-            &bitcoind_user,
-            &bitcoind_pwd,
-            "generatetoaddress",
-            "50",
-            &wallet_address,
-        ];
-        let (_, _, _) =
-            spawn_await_child_process(Scope::Bitcoind, command, args, &self.working_directory)
-                .await?;
-
```
rwlockbg, line 138: why was this removed ? This is the start code for the bitcoind node

bin/test-suite/src/suite/consensus/common/bitcoind_node.rs
```diff
@@ -94,48 +102,26 @@ impl BitcoindNodeConfig {
...
+            // Load the wallet if it already exists
+            Err(e) => {
+                if e.to_string().contains("wallet already exists") {
+                    bitcoin_client.load_wallet(BITCOIND_WALLET_NAME)?;
+                } else {
+                    return Err(anyhow::anyhow!("Failed to create wallet: {}", e));
+                }
+            }
+            Ok(_) => {}
+        }
```
rwlockbg, line 122: Why not just: if let Err(e) =  bitcoin_client.create_wallet(BITCOIND_WALLET_NAME, None, None, None, None) { ... } ?

bin/test-suite/src/suite/consensus/common/bitcoind_node.rs
arminsabouri: this was replaced with calls to the rpc instead of spawning new processes.

bin/test-suite/src/suite/consensus/common/bitcoind_node.rs
arminsabouri: I had a slight prefrence over match here. Just thought it was more readable

bin/test-suite/src/suite/consensus/common/bitcoind_node.rs
rwlockbg: got it.

bin/test-suite/src/suite/consensus/common/bitcoind_node.rs
rwlockbg: fine by me :)

rwlockbg (approved):

PR #582, opened 2025-02-11
Fix/db segfault

arminsabouri:
Hopefully closing https://github.com/botanix-labs/botanix/issues/889
Everytime a provider()? is created a db read tx is created. In poa/mod.rs we never return so this db read tx was left open indefinetly. We would see this in forms of [warnings](https://github.com/botanix-labs/botanix/issues/604) and [segfaults](https://github.com/botanix-labs/botanix/issues/889).

Please review this commit by commit.


scottmillner (dismissed):

lamafab:
This seems odd, the transactions are created internally and are short-lived:

```
// > crates/storage/db/src/mdbx.rs

/// Opens up an existing database or creates a new one at the specified path. Creates tables if
/// necessary. Read/Write mode.
pub fn init_db<P: AsRef<Path>>(path: P, args: DatabaseArguments) -> eyre::Result<DatabaseEnv> {
    let client_version = args.client_version().clone();
    let db = create_db(path, args)?;
    db.create_tables()?;
    db.record_client_version(client_version)?;
    Ok(db)
}
```

The `Transaction` type (`crates/storage/libmdbx-rs/src/transaction.rs`) does implement a drop guard.

scottmillner (approved):

lamafab (approved):

arminsabouri:
> This seems odd, the transactions are created internally and are short-lived:
> 
> ```
> // > crates/storage/db/src/mdbx.rs
> 
> /// Opens up an existing database or creates a new one at the specified path. Creates tables if
> /// necessary. Read/Write mode.
> pub fn init_db<P: AsRef<Path>>(path: P, args: DatabaseArguments) -> eyre::Result<DatabaseEnv> {
>     let client_version = args.client_version().clone();
>     let db = create_db(path, args)?;
>     db.create_tables()?;
>     db.record_client_version(client_version)?;
>     Ok(db)
> }
> ```
> 
> The `Transaction` type (`crates/storage/libmdbx-rs/src/transaction.rs`) does implement a drop guard.


This area of the code is not my expertise but I believe they are intended to be short-lived however its up to the caller to ensure they are cleaned up after being used. In most places in our code we get a db read tx and drop it at the end of a function call. Poa/mod main is a exception bc it never returns. 

>This seems odd

How do you mean?

lamafab:
> Poa/mod main is a exception bc it never returns.

Indeed, I was looking at the wrong part; `state_provider` is created like this:

```rust
/// State provider for latest block
#[track_caller]
pub fn latest(&self) -> ProviderResult<StateProviderBox> {
    trace!(target: "providers::db", "Returning latest state provider");
    Ok(Box::new(LatestStateProvider::new(self.db.tx()?, self.static_file_provider())))
}
```

It starts the transaction (`self.db.tx()?`), but `state_provider` is never dropped :+1:

bin/reth/src/commands/poa/mod.rs
```diff
@@ -579,6 +579,7 @@ impl<Ext: clap::Args + fmt::Debug> PoaNodeCommand<Ext> {
             error!(target: "reth::cli", "{}", e);
             panic!("{}", e);
         }
+        drop(state_provider);
```
rwlockbg, line 582: That should be dropped by itself further down. I dont understand why it would affect the db?

rwlockbg (approved):

bin/reth/src/commands/poa/mod.rs
lamafab:
```rust
        let (tx, rx) = oneshot::channel();
        executor.spawn_critical(
            "abci driver",
            Box::pin(async move {
                let res = abci_driver.start().await;
                let _ = tx.send(res);
            }),
        );

        match rx.await? {
            Ok(()) => info!("ABCIDriver exited successfully"),
            Err(error) => {
                error!(target: "reth::cli", %error, "ABCIDriver exited with an error")
            }
        };

        Ok(())
```

The `PoaNodeCommand::execute` method only exits if the ABCIDriver exits, which triggers the channel send. The `ABCIDriver::start` is blocking and does not spawn off.

PR #584, opened 2025-02-12
feat(bitcoin-server): Added utxo versioning

rwlockbg:


bin/btc-server/src/database/version.rs
```diff
@@ -0,0 +1,26 @@
+use super::error::Error;
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Serialize, Deserialize, Default, Clone, Copy, PartialEq)]
+pub enum UtxoVersion {
+    /// Original version with basic spending conditions
+    #[default]
+    V1 = 1,
+    /// Enhanced version with additional spending conditions ?
```
scottmillner: nit: why the ?

bin/btc-server/proto/btc_server.proto
```diff
@@ -131,9 +131,10 @@ message OutPoint {
 }
 
 message Utxo {
-    OutPoint outpoint = 1; 
+    OutPoint outpoint = 1;
     TxOut output = 2;
     string eth_address = 3;
+    uint32 version = 4;
```
arminsabouri: Version should not be set by the gRPC caller. For this PR its enough to just set the version field to a hardcoded value. In the future if we ever add a tapleaf spending condition for example we can use the version to decide how this UTXO should be spent

bin/btc-server/src/database/version.rs
arminsabouri: If its not used, lets not create a v2.

bin/btc-server/src/wallet/psbt.rs
```diff
@@ -33,6 +34,13 @@ lazy_static::lazy_static! {
...
         key: Vec::new(),
     };
+
+
+    pub static ref VERSION_TYPE_KEY: ProprietaryKey = ProprietaryKey {
+        prefix: PROP_KEY_PREFIX.to_vec(),
+        subtype: VERSION_TYPE,
+        key: Vec::new(),
+    };
 }
```
arminsabouri, line 45: Lets move the PSBT related code from database.rs to here. I

arminsabouri: Adding the utxo version to the psbt is a great idea but this should be set at `make_tx` not during update psbt. i.e the psbt creator should inform signers how they should be signing for specific psbt

arminsabouri (changes requested):

bin/btc-server/src/database/version.rs
rwlockbg: removed v2 and the ? along with that.

rwlockbg: removed

bin/btc-server/src/wallet/psbt.rs
rwlockbg: done

bin/test-suite/src/suite/consensus/frost/test_dkg.rs
```diff
@@ -1,7 +1,7 @@
 use super::error::Error;
 use crate::{it_error_print, suite::consensus::ConsensusIntegrationTestSuite};
 use bitcoin::{consensus::Encodable, Address, Amount};
-use btcserverlib::{frost_id, pegout_id::PegoutId};
+use btcserverlib::{database::version::UtxoVersion, frost_id, pegout_id::PegoutId};
```
arminsabouri: Is this used?

crates/consensus/authority/src/prost_parser.rs
```diff
@@ -38,6 +38,7 @@ where
 mod test {
     use crate::prost_parser::ProstMessageSerdelizer;
     use bitcoin::{hashes::Hash, Txid};
+    use btcserverlib::database::version::UtxoVersion;
```
arminsabouri: Seems unused

bin/btc-server/src/database/mod.rs
```diff
@@ -128,6 +148,33 @@ impl Db {
...
+    pub fn get_version_from_psbt_input(input: &bitcoin::psbt::Input) -> Option<UtxoVersion> {
+        input.proprietary.get(&VERSION_TYPE_KEY).and_then(|bytes| {
+            if bytes.len() == 4 {
+                let version = u32::from_le_bytes(bytes.as_slice().try_into().ok()?);
+                UtxoVersion::try_from(version).ok()
+            } else {
+                None
+            }
+        })
+    }
```
arminsabouri: These utils are great they just need to live in psbt.rs under the psbt trait ext

bin/btc-server/src/database/mod.rs
```diff
@@ -19,8 +19,21 @@ use serde::{Deserialize, Serialize};
...
 pub mod error;
-
+pub mod version;
 pub use error::Error;
+use version::UtxoVersion;
+
+const VERSION_TYPE: u8 = 4;
+
+lazy_static::lazy_static! {
+    static ref PROP_KEY_PREFIX: &'static [u8] = b"btx";
```
arminsabouri: This is duplicate code i believe. Once psbt operation is moved to psbt.rs this whole lazy static block can be removed

bin/btc-server/src/database/error.rs
```diff
@@ -29,6 +29,10 @@ pub enum Error {
     OutputNotFound(usize),
     #[error("hash engine error {0}")]
     HashEngine(#[from] std::io::Error),
+    #[error("wrong version number {0}")]
+    InvalidVersion(u32),
```
arminsabouri: I'd be a bit more specific here.
```suggestion
    #[error("Invalid UTXO version number {0}")]
    InvalidUTXOVersion(u32),
```

bin/btc-server/src/database/error.rs
```diff
@@ -29,6 +29,10 @@ pub enum Error {
...
+    #[error("extract tx error {0}")]
+    ExtractTx(#[from] ExtractTxError),
```
arminsabouri: Does this get used anywhere in the DB crate? I don't think db should be extracting txs?

bin/btc-server/src/database/mod.rs
arminsabouri: Even better after coin selection we make call to `create_psbt` to that method you can pass the UTXO version along with the utxo. 
Update InputDTO to include utxo version u32.

arminsabouri (changes requested):

scottmillner (dismissed):

bin/btc-server/src/database/error.rs
rwlockbg: removed

bin/btc-server/src/database/error.rs
rwlockbg: fixed

bin/btc-server/src/database/mod.rs
rwlockbg: done

bin/btc-server/src/database/mod.rs
rwlockbg: done

crates/consensus/authority/src/prost_parser.rs
rwlockbg: removed

bin/test-suite/src/suite/consensus/frost/test_dkg.rs
rwlockbg: cleaned up

bin/btc-server/src/pegout_scheduler/mod.rs
```diff
@@ -527,10 +527,18 @@ impl PegoutScheduler {
...
+                    .get_utxo(outpoint)
+                    .ok()
+                    .flatten()
+                    .map(|utxo| utxo.version)
+                    .unwrap_or_default();
                 change_utxos.push(database::Utxo {
                     outpoint,
                     output: output.clone(),
                     eth_address: None,
+                    version: utxo_version,
```
arminsabouri, line 541: More of a note rather than a request for change. This works for now. But when we upgrade the UTXO version we may need to look for change spks created with each different version.

arminsabouri (approved):
Looks good! @rwlockbg do we need to do any db migrations? What happens if we deserialize a non-versioned UTXO with the new structure?

scottmillner (approved):

bin/btc-server/src/pegout_scheduler/mod.rs
rwlockbg: I agree with both points made above. I believe we should make a research ticket on migrations/upgrading utxo versions and stash it for later.

PR #585, opened 2025-02-12
feat(node): Added state sync enabled boolean

rwlockbg:


scottmillner (approved):

crates/consensus/authority/src/snapshot_manager.rs
```diff
@@ -248,6 +251,10 @@ where
         + 'static,
 {
     async fn run(&mut self) -> Result<(), SnapshotManagerError> {
+        if !self.enable_state_sync {
+            tracing::info!("Snapshot manager is disabled. Exiting...");
+            return Ok(());
+        }
```
arminsabouri, line 257: Nit: ideally this conditional would be at the top level. i.e in poa/mod.rs. 1 so we dont have to pass additional flags all the way down. 2 its a bit odd that snapshot_manager determines its own existence. It should just exist if the caller determines it should

arminsabouri (approved):

PR #587, opened 2025-02-13
Arm/deprecate healthcheck task

arminsabouri:
And clean up dependencies that should be dev dependencies


crates/consensus/authority/src/lib.rs
```diff
@@ -49,7 +49,6 @@ pub use comet_bft::light_client::LightCBFTClientBuilder;
 mod dkg;
 mod excecution_utils;
 mod frost_task;
-mod healthcheck_task;
```
lamafab, line 52: Also remove `crates/consensus/authority/src/healthcheck_task.rs` itself?

lamafab (dismissed):

crates/consensus/authority/src/lib.rs
lamafab: This also means `FrostCommand::SendHealtcheckToPeers` should be deprecated; I can do this as part of my PR.

arminsabouri: >This also means FrostCommand::SendHealtcheckToPeers should be deprecated; I can do this as part of my PR.

Yes please.

SatwikPrabhu (dismissed):

crates/consensus/authority/src/lib.rs
arminsabouri: Removed!

rwlockbg (approved):

PR #592, opened 2025-02-14
test(contract): update mint amount

scottmillner:


arminsabouri (approved):
Weird this needed to be updated. But our int. tests pass? Odd.

scottmillner:
> Weird this needed to be updated. But our int. tests pass? Odd.

yeah pretty sure what happened is:

- the workflow downloads and builds the latest version of forge each time
- forge probably accounted for an eip that impacted how much gas is assigned each `op_code` and/or `calldata`
- this resulted in a `200 wei`difference in the final minted amount than we expected in the test
- bc the test only runs the contract call and nothing else so it sets up the evm itself

PR #593, opened 2025-02-14
Feature/btc server track mempool drops

scottmillner:
https://github.com/botanix-labs/botanix/issues/790

### DIFF
--- a/bin/btc-server/Cargo.toml
+++ b/bin/btc-server/Cargo.toml
@@ -79,3 +79,4 @@ tonic-build = "0.12.3"
 
 [dev-dependencies]
 test-log = "0.2.12"
+once_cell = { workspace = true }
--- a/bin/btc-server/src/pegout_scheduler/mod.rs
+++ b/bin/btc-server/src/pegout_scheduler/mod.rs
@@ -28,6 +28,9 @@ use thiserror::Error;
 
 use crate::{database, rpc};
 
+pub const TX_NOT_FOUND_BITCOIND_ERROR: &str = "no such mempool or blockchain transaction";
+pub const TX_NOT_IN_MEMPOOL_BITCOIND_ERROR: &str = "transaction not in mempool";
+
 macro_rules! print_safe {
     ($e:expr) => {
         $e.map(|v| v.to_string()).unwrap_or("ERR".to_owned())
@@ -516,7 +519,7 @@ impl PegoutScheduler {
             .filter(|tx| tx.created < cp_time)
             .map(|tx| tx.txid)
             .collect::<Vec<_>>();
-        info!(
+        debug!(
             "Txids older than checkpoint: {:?}",
             maybe_dropped_txs.iter().map(|t| t.to_string()).collect::<Vec<_>>()
         );
@@ -536,6 +539,12 @@ impl PegoutScheduler {
                     continue;
                 }
                 Err(e) => {
+                    // check error message to confirm the tx is not in the mempool
+                    if !e.to_string().to_lowercase().contains(TX_NOT_IN_MEMPOOL_BITCOIND_ERROR) {
+                        warn!("Error checking mempool for tx {}: {}", &txid, e);
+                        continue;
+                    }
+
                     // the tx has been dropped or there was a reorg
                     warn!("Tx {} is not in the mempool and not included in a block: {}", &txid, e);
                     self.un_track_tx(&txid)?;
--- a/bin/btc-server/src/pegout_scheduler/mod.rs
+++ b/bin/btc-server/src/pegout_scheduler/mod.rs
@@ -12,6 +12,7 @@ use std::{
 };
 
 use crate::{
+    badarg,
     pegout_id::PegoutId,
     telemetry::Telemetry,
     update_telemetry_error,
@@ -20,7 +21,11 @@ use crate::{
         util::{VerifyingKeyExt, VerifyingKeyExtError},
     },
 };
-use bitcoin::{Amount, Block, BlockHash, OutPoint, ScriptBuf, Transaction, TxOut, Txid};
+use bitcoin::{
+    absolute::LockTime, transaction::Version, Amount, Block, BlockHash, OutPoint, ScriptBuf,
+    Sequence, Transaction, TxIn, TxOut, Txid, Witness,
+};
+use bitcoin_hashes::Hash;
 use bitcoincore_rpc::RpcApi;
 use log::{debug, error, info, trace, warn};
 use serde::{Deserialize, Serialize};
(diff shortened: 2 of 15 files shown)


bin/btc-server/src/test_utils/mod.rs
```diff
@@ -185,8 +230,17 @@ pub fn trusted_dealer_setup(
 }
 
 // Util function to create a btc tx with random inputs and outputs as defined by fn params
-pub fn create_tx(num_inputs: usize, num_outputs: usize, change: Option<TxOut>) -> Transaction {
-    let txid = random_compute_txid();
+pub fn create_tx(
+    num_inputs: usize,
+    num_outputs: usize,
+    change: Option<TxOut>,
+    deterministic: bool, /* sets specific txid and p2wpkh_script so tx.txid is
```
arminsabouri: instead of a deterministic flag can we create and refrence a pub const transaction?
```
const test_transaction: Transaction = Transaction {
 inputs: ...
...,
}

**Review by arminsabouri — COMMENTED** [review]
I'll do a more deep dive review later on but just had one initial thought on the changes to a test utils

**Review by rwlockbg — DISMISSED** [review]

File: bin/btc-server/src/pegout_scheduler/mod.rs [hunk 2]
```diff
@@ -603,6 +493,55 @@ impl PegoutScheduler {
...
+        bitcoind: &impl RpcApi,
+        checkpoint: bitcoincore_rpc::json::GetBlockHeaderResult,
+    ) -> Result<(), SyncError> {
+        // Determine the timestamp of the checkpoint block
+        let cp_time = checkpoint.block_time();
+        // Get txs older than timestamp
+        let maybe_dropped_txs = self
+            .txs
+            .values()
+            .filter(|tx| tx.created < cp_time)
```
arminsabouri, line 519: Should we consider a some buffer? I.e `tx.created + 1 hour < cp_time`

bin/btc-server/src/pegout_scheduler/mod.rs
```diff
@@ -603,6 +493,55 @@ impl PegoutScheduler {
...
+                Ok(_) => {
+                    warn!("Tx {} still in the mempool", txid);
+                }
+                Err(e) => {
+                    // the tx has been dropped or there was a reorg
+                    warn!("Tx {} is not in the mempool and not included in a block: {}", txid, e);
+                    self.un_track_tx(&txid)?;
+                }
+            }
+
```
arminsabouri, line 553: Should we also check if the tx is not onchain? Perhaps the tx.create is less than the cp and its currently on its way to be deeply confirmed?

bin/btc-server/src/pegout_scheduler/mod.rs
```diff
@@ -603,6 +493,55 @@ impl PegoutScheduler {
...
+        for txid in maybe_dropped_txs {
+            // TODO(scott): first check if tx is in the `finalized outputs` if we implement this
+            // see Updates section in `https://github.com/botanix-labs/botanix/issues/701`
+
+            // a tx that is in a deeply confirmed block should have been handled already
+            // so check if still in mempool
+            let tx = self.txs.get(&txid).expect("tx should exist").clone();
+            match bitcoind.get_mempool_entry(&txid) {
+                Ok(_) => {
+                    warn!("Tx {} still in the mempool", txid);
```
arminsabouri: Do we need to continue here? Otherwise we seem to leave the match block and add the tx back to the pending queue

arminsabouri (changes requested):
Awesome work with the tests. Some initial thoughts on the btc server portion of this work

bin/btc-server/src/test_utils/mod.rs
scottmillner: ah yep def better approach
can't to a const but created static variable

done!

bin/btc-server/src/pegout_scheduler/mod.rs
scottmillner: what would the buffer achieve?

my thinking is the buffer is arbitrary anyway (could be plus 5 min or an hour) but we have the logic of conflicting inputs to prevent double spending an output in case the tx isn't actually older than the checkpoint

bin/btc-server/src/pegout_scheduler/mod.rs
scottmillner: we do want to add the tx back to the pending pegouts since it's still in the mempool
- we want to add it back to the pending pegouts no matter what
 -- if it's still in the mempool, let's try with another psbt and conflicting inputs got us covered
 -- if it's not in the mempool or in a block we need to retry
- but we only want to `un_track` if it doesn't exist at all (isn't in the mempool) since there's nothing to track

i left the `Ok()` arm so we could log when the tx is still in the mempool

bin/btc-server/src/pegout_scheduler/mod.rs
scottmillner: `track_mempool` is only called inside `sync_until` after it has finalized the latest deeply confirmed block so i don't think we'll gain anything by checking onchain here

bin/btc-server/src/pegout_scheduler/mod.rs
arminsabouri: >we want to add it back to the pending pegouts no matter what

if its in the mempool won't it fail as a double spend? Unless you abide by rbf rules.

scottmillner: if we add it back to pending pegouts:
- it'll be honored in a new psbt
- which will include a conflicting input
- and this won't be rejected since it's an entirely new transaction

so we'd have two transactions: both honoring the same output but the second one has a conflicting input so only one will be included in a block

and the one that is not included in a block and rejected, will get noticed by `track_mempool` and those will be added back to pending pegouts

that logic make sense? i feel like im taking a btc quiz 😆

bin/btc-server/src/pegout_scheduler/mod.rs
```diff
@@ -1224,4 +1193,58 @@ mod tests {
...
+        let tx = create_tx(1, 2, None);
+        let pegouts = pegout_requests_from_tx(&tx, &[0]);
+        pegout_scheduler.add_tx(tx.clone(), &pegouts, SystemTime::now());
+
+        let mock_bitcoind = MockBitcoind::new();
+        let mut checkpoint = mock_bitcoind
+            .get_block_header_info(&bitcoin::BlockHash::all_zeros())
+            .expect("valid checkpoint");
+        // increase time for checkpoint block so tracked tx is older
+        checkpoint.time = checkpoint.time + 5;
```
lamafab: nitpick: you can just do this :)

```rust
       checkpoint.time += 5;
```

bin/btc-server/src/pegout_scheduler/mod.rs
```diff
@@ -1224,4 +1193,58 @@ mod tests {
...
+
+        let result = pegout_scheduler.track_mempool(&mock_bitcoind, checkpoint);
+        assert!(result.is_ok());
+
+        // assert the pegout was added to pending pegouts
+        let pending_pegouts = db.get_pending_pegouts().expect("pending pegouts exist");
+        assert_eq!(pending_pegouts.len(), 1);
```
lamafab, line 1276: nitpick: for consistency:

```rust
        assert_eq!(pending_pegouts[0], pegouts[0]);
```

bin/btc-server/src/test_utils/mod.rs
```diff
@@ -102,6 +116,21 @@ impl bitcoincore_rpc::RpcApi for MockBitcoind {
                     &format!("{{\"hash\": \"{block_hash}\", \"confirmations\": 1, \"height\": 1, \"version\": 1, \"version_hex\": \"01000000\", \"merkleroot\": \"{block_hash}\", \"time\": {current_time}, \"mediantime\": {current_time}, \"nonce\": 1, \"bits\": \"1d00ffff\", \"difficulty\": 1, \"chainwork\": \"0000000000000000000000000000000000000000000000000000000000000001\", \"nTx\": 1, \"previousblockhash\": \"{block_hash}\", \"nextblockhash\": \"{block_hash}\"}}",),
                 ).unwrap());
         }
+        if method == "getmempoolentry" {
+            // error case is triggered by a specific txid
```
lamafab, line 123: I'd reference `track_mempool_should_untrack_and_add_back_pegout_when_not_in_mempool`

bin/test-suite/src/suite/consensus/common/bitcoind_node.rs
```diff
@@ -48,6 +52,27 @@ impl SpawnedBitcoindProcess {
...
+    pub async fn stop(&mut self, bitcoind_user: &str, bitcoind_password: &str) {
+        let datadir =
+            format!("-datadir={}", self.working_directory.join("data").display().to_string());
+
+        let status = Command::new("bitcoin-cli")
+            .arg("-regtest")
+            .arg(format!("-rpcuser={}", bitcoind_user))
+            .arg(format!("-rpcpassword={}", bitcoind_password))
+            .arg("-rpcport=18443")
+            .arg(datadir)
```
lamafab: Since we're connecting via RPC, this datadir is probably not necessary here.

bin/test-suite/src/suite/consensus/common/bitcoind_node.rs
```diff
@@ -105,6 +131,34 @@ impl BitcoindNodeConfig {
             working_directory: self.working_directory.clone(),
         })
     }
+
+    // this re-starts an existing stopped bitcoind instance and updates the local context
+    pub async fn re_start(&mut self, suite: &mut ConsensusIntegrationTestSuite) {
+        match self.spawn_service() {
+            Ok(bitcoind_process) => {
+                tokio::time::sleep(Duration::from_secs(6)).await;
```
lamafab, line 135: As mentioned in internal comms, I really don't like stuff like that 😅 But probably a straight forward option for now.

bin/btc-server/Cargo.toml
```diff
@@ -46,6 +46,7 @@ hex = "0.4.3"
 lazy_static = "1.4.0"
 log = "0.4.18"
 miniscript = "12.3.0"
+once_cell = { workspace = true }
```
arminsabouri: Used only in tests. Please move to dev dependencies

bin/btc-server/src/pegout_scheduler/mod.rs
arminsabouri: >so we'd have two transactions

mempools dont retain two transaction spending the same inputs. if you submit a tx spending the same input it either needs to adhere to [RBF rules](https://bitcoinops.org/en/topics/replace-by-fee/) or it will be rejected as a double spend. Just to be clear we are not doing RBF

arminsabouri:
also @scottmillner one small conflict in cargo.toml

bin/test-suite/src/suite/consensus/frost/test_track_mempool.rs
```diff
@@ -0,0 +1,248 @@
...
+
+    // assert the pegout is not in the pending pegout list
+    let pending_pegouts = clients[0]
+        .get_pending_pegouts(tonic::Request::new(client::Empty {}))
+        .await
+        .expect("get pending pegouts")
+        .into_inner();
+    let pending_pegout_ids =
+        pending_pegouts.pending_pegouts.iter().map(|p| p.pegout_id.clone()).collect::<Vec<_>>();
+    assert!(!pending_pegout_ids.contains(&pegout_id_bytes.to_vec()));
```
lamafab, line 152: I'd do a positive check here, making sure that there at least is one correctly.

bin/test-suite/src/suite/consensus/frost/test_track_mempool.rs
```diff
@@ -0,0 +1,248 @@
...
+const NUM_PEGINS: usize = 5;
+
+/// Test that the mempool is tracked correctly
+///
+/// Pegin then pegout which will:
+/// create, sign, broadcast, and track a transaction but don't include it in a block.
+/// Then restart bitcoind to clear the mempool of the tracked tx.
+/// Then generate some blocks so the tracked tx is older than the checkpoint which should add
+/// the tracked tx back to the pending pegout list since it's not in the mempool or in a block.
+pub async fn test_track_mempool(
```
lamafab, line 31: I do not grasp the full scope yet to assess this test confidently, FYI.

bin/btc-server/src/pegout_scheduler/mod.rs
scottmillner: gotchya, makes sense: wasn't thinking through that mempool would reject the new tx with same input, thanks!
and yep, not doing rbf

so ill just log that the tx is in the mempool and `continue` so it's not added back to `pending_pegouts` 
- it'll eventually be included in a block or dropped which we retry

bin/btc-server/src/pegout_scheduler/mod.rs
scottmillner: ah yep, totally wasn't paying attention there
thanks!

done!

bin/btc-server/src/pegout_scheduler/mod.rs
arminsabouri: yes agreed

bin/test-suite/src/suite/consensus/common/bitcoind_node.rs
scottmillner: totally agree: there's a few things we need to "cleanup" in the test-suite

bin/test-suite/src/suite/consensus/frost/test_track_mempool.rs
scottmillner: more than happy to walk through the scope of this pr and this test

bin/test-suite/src/suite/consensus/frost/test_track_mempool.rs
scottmillner: as written, there should be no `pending_pegouts` at this point:
- so could have just checked `pending_pegout_ids.is_empty()`
- but wanted to specifically check against `pegout_id_bytes` in case this test was refactored to include mulitple pegouts in the future

but maybe just `assert!(pending_pegout_ids.is_empty());` makes more sense

bin/btc-server/src/pegout_scheduler/mod.rs
scottmillner: done!

bin/btc-server/src/test_utils/mod.rs
scottmillner: done!

bin/btc-server/src/pegout_scheduler/mod.rs
scottmillner: done!

bin/btc-server/Cargo.toml
scottmillner: done!

scottmillner:
> also @scottmillner one small conflict in cargo.toml

done!

bin/test-suite/src/suite/consensus/common/bitcoind_node.rs
scottmillner: done!

rwlockbg (dismissed):
Very good PR! Very well written in Rust changes!

lamafab (dismissed):
:+1:

SatwikPrabhu (dismissed):
Sorry for the late review. Awesome work though! Looks good to me

bin/btc-server/src/pegout_scheduler/mod.rs
```diff
@@ -611,6 +501,57 @@ impl PegoutScheduler {
...
+            // a tx that is in a deeply confirmed block should have been handled already
+            // so check if still in mempool
+            let tx = self.txs.get(&txid).expect("tx should exist").clone();
+            match bitcoind.get_mempool_entry(&txid) {
+                Ok(_) => {
+                    warn!("Tx {} still in the mempool", &txid);
+                    // nothing else to do: eventually the tx will be confirmed or dropped
+                    continue;
+                }
+                Err(e) => {
```
arminsabouri, line 541: Should we do specific error checking here? I imagine there are multiple reasons why this call could fail and in most senarios we don't want to un track these txs

bin/btc-server/src/pegout_scheduler/mod.rs
```diff
@@ -611,6 +501,57 @@ impl PegoutScheduler {
...
+        // Determine the timestamp of the checkpoint block
+        let cp_time = checkpoint.block_time();
+        // Get txs older than timestamp
+        let maybe_dropped_txs = self
+            .txs
+            .values()
+            .filter(|tx| tx.created < cp_time)
+            .map(|tx| tx.txid)
+            .collect::<Vec<_>>();
+        info!(
```
arminsabouri: Nit: this is a heavy log. maybe debug!

bin/btc-server/src/pegout_scheduler/mod.rs
arminsabouri: I probably forget the internals of sync_until but my impression of it is that it will only untrack a tx after its deeply confirmed. So the tx is still in the chain awaiting confirmation. but here we are checking if the tx is in the mempool and untracking it when it errors out but its actually still on-chain. Lmk if not thinking about this right

scottmillner: yes, sync_until `will only untrack a tx after its deeply confirmed`

i think the question is around: `but its actually still on-chain`
- if our code functions as intended, the tx is not on chain if we get to this code
- if it was on chain, it's either in the mempool, or included in a deeply confirmed block
- but it's not in the mempool or any block bc sync_until would have verified that and removed it

but a sanity check here would defend against a bug in our code and if im not thinking about this correctly lol

ive added the onchain check

done!

bin/btc-server/src/pegout_scheduler/mod.rs
scottmillner: done!

bin/btc-server/src/pegout_scheduler/mod.rs
scottmillner: good idea: added a check to confirm the error is bc tx doesn't exist and only untrack in that case

done!

lamafab (approved):

SatwikPrabhu (approved):

arminsabouri (approved):

scottmillner:
ran the tests directly on the runner and all passed
the ci job kept disconnecting:

int_tests

PR #595, opened 2025-02-18
Botanix up

arminsabouri:
Introducing a new binary, Botanix up. The goal of this binary is to easily auto generate the cometbft binaries for new testnet deployments, local dev setup and setting up new devs. The configs are all setup using existing test suite utils. Feature requested by @birozuru  for the new federation setup

Example run in the bin/botanix-up dir:

`cargo run --num-nodes 2 --multisig-min-signers 2 --multisig-max-signers 2`

```
.
├── Cargo.toml
├── cometbft
│   ├── node-0
│   │   ├── config
│   │   │   ├── config.toml
│   │   │   ├── genesis.json
│   │   │   ├── node_key.json
│   │   │   └── priv_validator_key.json
│   │   └── data
│   │       └── priv_validator_state.json
│   └── node-1
│       ├── config
│       │   ├── config.toml
│       │   ├── genesis.json
│       │   ├── node_key.json
│       │   └── priv_validator_key.json
│       └── data
│           └── priv_validator_state.json
├── poa
│   ├── federation.toml
│   ├── node-0
│   │   ├── bjwt.hex
│   │   └── discovery-secret
│   └── node-1
│       ├── bjwt.hex
│       └── discovery-secret
└── src
    ├── cli.rs
    └── main.rs

```

### DIFF
--- a/bin/botanix-up/src/main.rs
+++ b/bin/botanix-up/src/main.rs
@@ -6,13 +6,20 @@ use crate::comet_node::{get_enode, TestSignal};
 use anyhow::{Context, Result as AnyResult};
 use clap::Parser;
 use cli::Cli;
+use reth_node_core::{
+    args::{FedMemberPubKey, FederationTomlConfig},
+    primitives::Address,
+};
+use secp256k1::SECP256K1;
 use std::{
     fs,
+    io::Write,
     path::{Path, PathBuf},
 };
 use test_suite::suite::consensus::common::{
     comet_node::{self, updated_genesis_file, GenesisValidator, PrivValidator},
-    poa_node::ABCI_PORT_BASE,
+    poa_node::{ABCI_PORT_BASE, DISCOVERY_PORT_BASE},
+    MINTING_CONTRACT_BYTECODE,
 };
 use tokio::{self, sync::broadcast::channel};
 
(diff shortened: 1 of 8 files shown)


scottmillner (commented):
🔥 this is sure handy 😃

scottmillner (dismissed):

SatwikPrabhu (dismissed):

lamafab (dismissed):

lamafab:
You know, I do think it be worth it implementing a common utility function that checks whether the necessary bins (`bitcoind`, `cometbft`, etc) are in your `$PATH`, and print an explicit error if not  - I fall for this often 😵‍💫

Tool works great, otherwise.

bin/botanix-up/src/main.rs
```diff
@@ -0,0 +1,187 @@
...
+
+    Ok(())
+}
+
+async fn inner_main() -> AnyResult<()> {
+    let cli = Cli::parse();
+    // Basic sanity checks
+    cli.validate()?;
+    let output_path = PathBuf::from(
+        cli.output_path.unwrap_or(std::env::current_dir()?.to_str().unwrap().to_string()),
```
lamafab, line 169: Maybe it's worth writing everything into a single `output/` directory?  Basically _pwd_ + `/output/`

rwlockbg (dismissed):
Looks fine to me, but wonder if this should be part of the repo @0xBEEFCAF3 @scottmillner  Thinking of it, it feels like a tool rather than a part o the protocol ? Maybe worth putting in a separate repo and having it as a submodule here ?

bin/botanix-up/src/main.rs
arminsabouri: yes great idea!

arminsabouri:
> Looks fine to me, but wonder if this should be part of the repo @0xBEEFCAF3 @scottmillner Thinking of it, it feels like a tool rather than a part o the protocol ? Maybe worth putting in a separate repo and having it as a submodule here ?

It definetly is a tool rather than part of the protocol. We could seperate out non protocol binaries (btc-server, fed-utils, botanix-up). Then its a question to monorepo or not. And that feels outside of scope of this PR

bin/botanix-up/src/main.rs
arminsabouri: Done

scottmillner (approved):

lamafab (approved):

rwlockbg (approved):

PR #571, opened 2025-02-05
Chore: Pre-audit plumbing tasks

SatwikPrabhu:
Fix for https://github.com/botanix-labs/botanix/issues/895

PR #599, opened 2025-02-21
fix: during block sync frost task should be disabled

rwlockbg:
Ticket:   https://github.com/botanix-labs/botanix/issues/901


bin/btc-server/src/coordinator/mod.rs
```diff
@@ -149,15 +149,14 @@ pub async fn make_tx(
...
-                    .map(|u: &Utxo| {
+                utxos.get(op).ok_or_else(|| CoordinatorError::MissingUtxoForConflictingInput).map(
+                    |u: &Utxo| {
                         // Conflicting utxos will be added to available utxos before finishing
                         // coin selection
                         conflicting_utxos.insert(*op, u.clone());
                         u.clone()
-                    })
+                    },
+                )
```
arminsabouri: Is this just a formatting change?

bin/reth/src/commands/poa/mod.rs
```diff
@@ -788,12 +791,13 @@ impl<Ext: clap::Args + fmt::Debug> PoaNodeCommand<Ext> {
...
         // different structs
-
         if is_fed_node {
             executor.spawn_critical(
                 "Frost Task",
                 Box::pin(async move {
-                    frost_task.expect("frost task exists").start_task().await;
+                    if block_sync_complete_rx.recv().await.is_ok() {
+                        frost_task.expect("frost task exists").start_task().await;
+                    }
```
arminsabouri: The problem is cbft can sync at any point. There isn't a garuntee that it will only sync at the beginning. My recommendation is to add a new is_syncing bool to the rwlock protected potion to `storage`.

arminsabouri: If this bool is on, the tasks that are dependent on this busy loop until its flagged off

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -513,6 +536,37 @@ where
...
+    let status = light_client.status().await?;
+
+    let sync_info = status.sync_info;
+
+    // check with the catching up flag
+    if sync_info.catching_up {
+        return Ok(true);
+    }
+
+    // compare block time with current time
```
arminsabouri: Can we not just rely on `sync_info`?

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -1290,7 +1344,30 @@ where
...
         );
+
         debug!("finalize_block request: {:?}", request);
+
+        // check block sync status
+        let block_sync_complete_tx = self.block_sync_complete_tx.clone();
+        let cbft_rpc_provider =
+            self.cbft_rpc_provider.build_and_connect().expect("light client to connect");
+        self.task_executor.spawn_blocking(Box::pin(async move {
+            match is_syncing(cbft_rpc_provider).await {
```
arminsabouri: Instead of a one-time task I would just provide the frost_task with the light client or cometfbt rpc provider (if there is a way to get is_syncing).

arminsabouri (changes requested):
I would prefere to remove the additional message queue in favor of just passing the data provider directly. Two reasons why
1. Keeps the code simple. Don't need another task
2. There is garuntee that comet will not sync. A gap in blocks can form at anypoint.
 
As an aside, most of the changes here are formatting changes. I'm sure this was just an output of `make fmt` or lint. But how did this happen? Did we push wrongly formatted code to main? Perhaps in #571

bin/btc-server/src/coordinator/mod.rs
rwlockbg: yes, the current main has not been formatted! Hence the changes.

crates/consensus/authority/src/comet_bft/abci.rs
rwlockbg: I hope we can, I jsut added the rest in case it fails. But it does make sense to rely on it, ofc. I have removed the other checks.

bin/reth/src/commands/poa/mod.rs
rwlockbg: I can do that, ofc. I thought the issue reported was that syncing would usually be at the beginning when several blocks are to be pulled until it catches up. At this point the frost task should be made active. But if permanent syncing is to be considered, then we have to continuously keep pausing the frost task and release when syncing is ready, right ?

crates/consensus/authority/src/comet_bft/abci.rs

bin/reth/src/commands/poa/mod.rs
rwlockbg: done

crates/consensus/authority/src/comet_bft/abci.rs

scottmillner (dismissed):

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -60,6 +60,17 @@ use tendermint_proto::{
     },
 };
 
+#[derive(Debug, Error)]
```
arminsabouri: Is this error used? Looks like its re-defined in frost_task.rs

crates/consensus/authority/src/frost_task.rs
```diff
@@ -295,6 +319,24 @@ where
         let mut canon_state_notifs = self.storage.client.subscribe_to_canonical_state();
 
         loop {
+            // get sync status
+            match self.is_syncing().await {
+                Ok(is_syncing) => {
+                    self.storage.inner.write().await.is_block_syncing = is_syncing;
+                    if !is_syncing {
```
arminsabouri: Don't we want to the reverse of this logic? Stall the task if there is a block sync

arminsabouri (changes requested):

crates/consensus/authority/src/frost_task.rs
rwlockbg: yep, had the right info! but the check was wrong, sorry. Fixed it now.

crates/consensus/authority/src/comet_bft/abci.rs
rwlockbg: Yep, had it twice, forgot to remove it after code-rearrangement. My linter did not show me an unused struct, weird ...

scottmillner (approved):

arminsabouri (approved):

scottmillner:
int test pass locally, merging

PR #603, opened 2025-02-26
fix(rpc): set agg_pk from block 1 before validating

scottmillner:
small bug: need to set the aggregate pubkey for rpc in `finalize_block` now instead of `commit()`
this is bc we recently added the check `self.validate_block` which requires the agg pk to be stored

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -759,7 +759,7 @@ start-non-fed-server-1:
 	--bitcoind.username "${BITCOIND_USER}" \
 	--bitcoind.password "${BITCOIND_PWD}" \
 	--p2p-secret-key "${NON_FED_1_DIR}/discovery-secret" \
-	--port 30306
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -759,7 +759,7 @@ start-non-fed-server-1:
 	--bitcoind.username "${BITCOIND_USER}" \
 	--bitcoind.password "${BITCOIND_PWD}" \
 	--p2p-secret-key "${NON_FED_1_DIR}/discovery-secret" \
-	--port 30306
+	--port 30306 \
```
scottmillner, line 762: backslash was missing so rpc wouldn't run on correct abci port

rwlockbg (approved):

lamafab (approved):

arminsabouri (approved):

PR #604, opened 2025-02-26
feat(btc-server): Pending Pegout Filtering & Pruning

lamafab:
This is my (second) attempt to fix issue https://github.com/botanix-labs/botanix/issues/903; I tried to keep it as simple as possible without introducing steep changes and not requiring us to change the UTXO selection mechanism, which would be high-risk and not something we'd want to do shortly before the audit. All of this will change anyway once we move toward a dynamic and federated validator model. Essentially, this introduces two new methods to the `btc-server` database API: one to only select economically viable pegouts, and another method to prune pegouts after exceeding a certain amount.

This is first opened as a **Draft to discuss** whether this proposal is a good way to implement the desired outcome.

## Summary
This PR adds two key features to improve pegout handling:

* **Economical pegout filtering**: Only process pegouts where the value exceeds the estimated transaction fee, preventing net-negative transfers that would cost more in fees than they're worth.
* **Pegout pruning**: Automatically prune pending pegouts when their number exceeds a threshold (10,000), prioritizing higher-value transfers.

## Implementation details

* Added `cord_pending_pegouts()` method to filter economically viable pegouts based on current fee rates
* Implemented `prune_pending_pegouts()` to remove low-value pegouts when count exceeds threshold
* Modified `get_psbt` workflow (for coordinator) to check economic viability before processing
* Added pruning call when receiving new pending pegouts from consensus
* Added basic fee calculation estimates:
	1. The output's own size in vbytes
	2. A fair portion of transaction overhead
	3. A generous estimate for input costs, assuming one input per pegout

## Further Thoughts

After studying this section of the codebasefor a while:

1. In the long run we need to think about proper UTXO management and consolidation, and calculate precise fee requirements for pegouts based on that. As noted earlier, this depends on how the dynamic federation model will look like in the near future.
2. The database would benefit from a proper querying language; we often load **all** pegouts into memory for processing when single-lookup or queries would be more appropriate. Most damningly, we load all pegouts every single time a new pegout is stored in order to calculate the updated Merkle root!

### DIFF
--- a/bin/btc-server/src/database/mod.rs
+++ b/bin/btc-server/src/database/mod.rs
@@ -1009,6 +1009,96 @@ mod tests {
     }
 
     #[test]
+    fn can_prune_pegouts_by_count() {
+        let (db, _temp_dir) = setup_db();
+
+        // create a 10 random pegouts
+        for i in 0..10 {
+            let pegout_id = PegoutId::new([i as u8; 32], 0);
+            let req = pegout_scheduler::PegoutRequest {
+                id: pegout_id,
+                spk: random_p2wpkh_script(),
+                value: Amount::from_sat(10_000 * (i as u64 + 1)),
+                botanix_height: 1,
+            };
+            db.store_pending_pegout(&req).unwrap();
+        }
+
+        // Retain ten pegouts (all).
+        db.do_prune_pending_pegouts(10).unwrap();
+        let pegouts = db.get_pending_pegouts().unwrap();
+        assert_eq!(pegouts.len(), 10);
+        assert_eq!(pegouts[0].value, Amount::from_sat(10_000));
+        assert_eq!(pegouts[9].value, Amount::from_sat(100_000));
+
+        // Retain four pegouts.
+        db.do_prune_pending_pegouts(4).unwrap();
+        let pegouts = db.get_pending_pegouts().unwrap();
+        assert_eq!(pegouts.len(), 4);
+        assert_eq!(pegouts[0].value, Amount::from_sat(70_000));
+        assert_eq!(pegouts[3].value, Amount::from_sat(100_000));
+
+        // Retain zero pegouts (none).
+        db.do_prune_pending_pegouts(0).unwrap();
+        let pegouts = db.get_pending_pegouts().unwrap();
+        assert!(pegouts.is_empty());
+    }
+
+    #[test]
+    fn can_coordinate_pegout_efficiency() {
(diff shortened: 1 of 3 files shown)


arminsabouri:
>The database would benefit from a proper querying language;

This was the motivation behind https://github.com/botanix-labs/botanix/issues/876 ! Unless we want to chagne our sled db strucutre we are kinda stuck with O(n) loads. Its not acceptable for true production. Sled also has really poor atomic tx interface. I would say instead of fixing sled lets pick a better tool.

>Only process pegouts where the value exceeds the estimated transaction fee, preventing net-negative transfers that would cost more in fees than they're worth.

The metric to filter by imo should be pegouts which entered the queue first. As those will be the customers who have waited the longest.

>Pegout pruning

Just to be clear we dont want to remove pegouts from the queue. We simply want to ensure that each pegout tx has a sane number of outputs. If there are more outputs than this limit then we should pick the ones that have been waiting the longest. 

>A generous estimate for input costs, assuming one input per pegout

This doesnt have to be the case, there can be >1 input being spent per pegout output

bin/btc-server/src/database/mod.rs
```diff
@@ -700,6 +702,93 @@ impl Db {
...
+    }
+
+    /// Prune pending pegouts once [`MAX_PENDING_PEGOUTS`] is reached. Pegouts
+    /// with higher values are retained.
+    pub fn prune_pending_pegouts(&self) -> Result<(), Error> {
+        self.do_prune_pending_pegouts(MAX_PENDING_PEGOUTS)
+    }
+
+    /// Prune pending pegouts once `retain_amt` is reached. Pegouts
+    /// with higher values are retained.
```
scottmillner, line 767: unfortunately we can't prune pegouts this way:
- at this point, users have already pegged out on the L2 and their funds have been burned
- so we still have to keep track of their pending pegout to make sure it's eventually honored on the L1

lamafab: So I must have misunderstood you; you mentioned that on the testnet you sometimes you just delete the database to deal with this? Or were you referring to the Utxo's, not the individual pegouts?

scottmillner: Sorry for any confusion.

Yes, we have deleted the db on testnet and users lose their pegouts.  But since this is only testnet funds we deemed this acceptable but not something we can do on mainnet.

lamafab:
_Replacement will follow soon_

PR #607, opened 2025-02-26
Fix/rpc mempool gossip

scottmillner:
NOTE: **This is an awful temporary fix but does work.**

Ultimately the TransactionManager future is getting starved.  I wasn't able to figure out the exact reason why.  This wakes the waker periodically so the future gets polled and pending txs are gossiped to peers.

### DIFF
--- a/crates/net/network/src/transactions/mod.rs
+++ b/crates/net/network/src/transactions/mod.rs
@@ -1236,6 +1236,16 @@ where
 
         let this = self.get_mut();
 
+        // Ensure periodic wake-up in case of starvation
+        // TODO(should not be used in mainnet): the root cause for starvation needs to be determined
+        // Temporary fix: this future is sometimes starved and
+        // never polled again This means pending txs are never gossiped to peers.
+        let waker = cx.waker().clone();
+        tokio::task::spawn(async move {
+            tokio::time::sleep(Duration::from_secs(3)).await;
+            waker.wake_by_ref();
+        });
+
         // All streams are polled until their corresponding budget is exhausted, then we manually
         // yield back control to tokio. See `NetworkManager` for more context on the design
         // pattern.


lamafab:
Do periodic wakeup signals cause that Future to get pushed up the priority list in the tokio event queue? Just wondering why this would work as a fix, unless that Future fails to do it's job when the actual wakeup signal should be triggered.

crates/net/network/src/transactions/mod.rs
```diff
@@ -1236,6 +1236,15 @@ where
 
         let this = self.get_mut();
 
+        // Ensure periodic wake-up in case of starvation
+        // Temporary fix: this future is sometimes starved and never polled again
```
lamafab: I'd put in an explicit _"TODO"_ so it can be grep-ed.

scottmillner: added a big TODO (wanted to add a throw-up emoji) 😆

scottmillner:
> Do periodic wakeup signals cause that Future to get pushed up the priority list in the tokio event queue? Just wondering why this would work as a fix, unless that Future fails to do it's job when the actual wakeup signal should be triggered.

the waker is never woken up sometime bc it never gets past the conditional

```
if maybe_more_network_events ||
            maybe_more_commands ||
            maybe_more_tx_events ||
            maybe_more_tx_fetch_events ||
            maybe_more_pool_imports ||
            maybe_more_pending_txns
        {
            // make sure we're woken up again
            cx.waker().wake_by_ref();
            return Poll::Pending
        }
```

arminsabouri:
There is an interesting comment at the top of this file
> An endless future. Preemption ensure that future is non-blocking, nonetheless. See
/// [`crate::NetworkManager`] for more context on the design pattern.

Is it possible that one (or all) of our tasks that is not implemented using this never ending future pattern is not yielding? Perhaps easy to one thing to try would be to disable all non-consensus crutial tasks and re-implement the abci driver as an endless future and see if we can reproduce

scottmillner:
closing for now but will reopen if needed

PR #609, opened 2025-02-27
feat(btc-server): Upper pegout bound & priority by age

lamafab:
Fixes https://github.com/botanix-labs/botanix/issues/903, replaces https://github.com/botanix-labs/Macbeth/pull/604

**_LATEST -> https://github.com/botanix-labs/Macbeth/pull/609#issuecomment-2711318648_**

## Summary

- ~Added a `created` field to the `pegout_scheduler::PegoutRequest` struct that stores a `SystemTime` timestamp. This design decision makes it consistent with `pegout_scheduler::Tx`~
- ~Updated the protobuf definition to include a timestamp field for pegouts~
- ~Modified the pegout deserialization to support backward compatibility:~
  - ~When reading existing pegouts from the database that lack timestamps, `SystemTime::now()` is used as the default value~
  - ~This ensures seamless compatibility with existing database records~
- Added a new `coord_pending_pegouts` method that:
  - Limits the number of pegouts to process in a single transaction
  - Sorts pegouts by creation time (oldest first) to implement FIFO ordering
- Introduced a 500-pegout upper bound for transactions with proper documentation
- ~Set the pegout creation timestamp to the block's timestamp when generated~

## Testing:

- Added test cases to verify FIFO ordering behavior
- ~Included a backward compatibility test to ensure proper handling of legacy pegout requests~

In the long run, all pending pegouts will contain timestamps, eliminating the need for the backward compatibility mechanism.

### DIFF
--- a/bin/btc-server/src/bin/main.rs
+++ b/bin/btc-server/src/bin/main.rs
@@ -57,6 +57,19 @@ use btcserverlib::{
 
 const JWT_HEADER_KEY: &str = "trace-proto-bin";
 
+/// The upper bound on pegouts in a single transaction. We use a _reasonable_
+/// number based on the following properties:
+///
+/// * Bitcoin's upper bound on a transaction is 100kb
+/// * 32 bytes required for an output (ie. pegout)
+/// * 110 bytes required for an input
+/// * We assume each output has an input
+///   * In practice, it's more likely that an individual input will map to multiple smaller outputs.
+///     But a larger output could also consume multiple inputs.
+///
+/// With a pegout bound of 500, we can conclude: (32 + 110) * 500 = 71_000
+const UPPER_PEGOUT_BOUND: usize = 500;
+
 macro_rules! already_exists {
     ($($arg:tt)*) => {{
         tonic::Status::already_exists(format!($($arg)*))
@@ -968,7 +981,11 @@ where
         }
 
         debug!("Cord Fee rate: {:?}", fee_rate);
-        let pending_pegouts = self.db.get_pending_pegouts().to_status()?;
+
+        // Select up to `UPPER_PEGOUT_BOUND` pegouts, sorted by age in ascending
+        // order. Respectively, the oldest pegouts come first.
+        let pending_pegouts = self.db.coord_pending_pegouts(UPPER_PEGOUT_BOUND).to_status()?;
+
         if let Some(telemetry) = self.telemetry.as_ref() {
             telemetry.update_pending_pegouts(pending_pegouts.len() as i64);
         }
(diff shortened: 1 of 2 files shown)


arminsabouri:
Approach ACK! I can review the code when its out of draft
>Updated the protobuf definition to include a timestamp field for pegouts


Instead of accepting a ts from the caller why not just use the current sys time. For the pruposes of FIFO sorting I think this would work fine

scottmillner:
approach ack!

lamafab:
@0xBEEFCAF3 

> Instead of accepting a ts from the caller why not just use the current sys time. For the pruposes of FIFO sorting I think this would work fine

The idea is that the pegout timestamp reflects when that pegout was created (L2 EVM block), not when the pegout was first identified by the `btc-server`. We can assume that in practice the difference between the two is quite narrow for an up-to-date node, but syncing historic blocks would change that. While this would not affect the FIFO ordering itself, I do think that the timestamp for pegouts should be accurate and deterministic, and simply using `SystemTime::now()` can be misleading. We might also use that field for any future work. I do think using the block header timestamp is justified, but maybe I'm overthinking this.

Additionally, this design decision is consistent with the `TrackedTx` structure:

```proto
message TrackedTx {
    bytes txid = 1;
    Transaction tx = 2;
    repeated uint32 pegout_idxs = 3;
    repeated PendingPegout pegout_requests = 4;
    repeated uint32 change_idxs = 5;
    google.protobuf.Timestamp created = 6;
}
```

SatwikPrabhu:
> The idea is that the pegout timestamp reflects when that pegout was created (L2 EVM block), not when the pegout was first identified by the btc-server.

Agree with this. Just wondering though that we would need to get the info of creation time from sidecar right? Or do you mean the time that we first get the pegout from sidecar?

lamafab:
> Agree with this. Just wondering though that we would need to get the info of creation time from sidecar right? Or do you mean the time that we first get the pegout from sidecar?

Discussed internally; accepting a timestamp from a light client is not advised, potential attack vector.

**Other topic up for discussion**: should we just use the `botanix_height` instead of a timestamp? This could also serve as a better identifier to lookup more information about the block where the pegout was created, including the block header timestamp.

lamafab:
Updated/rebased this PR to simply sort based on and rely on the Botanix height instead of a custom timestamp; it's much simpler now :)

EDIT: I'm not going to bother with the lints now, that's a separate issue to be solved (version mismatches)

scottmillner (approved):

SatwikPrabhu (approved):

rwlockbg (approved):

PR #611, opened 2025-03-07
Introduce `ActivationManager` for network upgrade proposals (via CometBFT)

lamafab:
This PR introduces a standalone `ActivationManager` which implements our proposed [Network Upgrade specification](https://github.com/botanix-labs/Macbeth/pull/611#issuecomment-3644142962).

Early feedback on the specification and enforcement logic is appreciated.

***

**EDIT (2025-03-20)**: Comments and recommendations have been addressed, **ready to merge**!

Changes:
* Renamed `threshold` to `approval_rate`
* Renamed `accepting` to `is_compliant`
* Wrapped major/minor version indicators in new-types
  * respectively: `RuntimeVersion { MajorVersion(u16), MinorVersion(u16) }`
* Adjusted approval rate formula calculations; instead of checking `min_validator_count` independently, it's now part of calculation itself. Respectively:

```rust
        let total = votes.len().max(min_validator_count);
        let votes_received = votes.iter().filter(|(_, e)| e.vote == Vote::Aye).count();

        // Calculate percentage (0-100) of votes received
        let quorum = (votes_received * 100).div_ceil(total);
```

This means that if we for example require at least three votes and we have one Aye vote, we calculate `approval_vote = 34` and can just check `approval_vote >= 67` - instead of having to do `approval_vote >= 67 && votes_len >= 3`. 

**EDIT (2025-03-17)**: PR is out of Draft and now ready for review!

Unit test coverage -> https://github.com/botanix-labs/Macbeth/pull/611#issuecomment-2729803006

**EDIT (2025-03-14)**: Completed comprehensive unit test suite:

*  Case: Vote Aye and wait for acceptance threshold
*  Case: Vote Aye but votes expire after sampling period; upgrade does not pass
*  Case: Vote Nay but still accept upgrade when conditions are met
*  Case: Vote Nay and refuse to upgrade (explicit fork)
*  Case: Ignore upgrade completely (implicit fork)
*  Case: Majority votes Nay but is willing to accept; upgrade does not pass
*  Case: Properly handle version transitions at exact target height
*  Case: Reject outdated version
*  Case: Reject unsupported (future) version
*  Case: Upgrade never passes due to insufficient conditions
*  Case: Lagging validator can catch-up (sync) with finalized upgrade
*  Case: Validators ignore votes on mismatched upgrade versions 

Migrated and expanded all test cases using the improved testing structures! I'll add more documentation and do some final cleanups, and then we're done. I will convert this from a Draft to a PR when appropriate.

### DIFF
--- a/crates/consensus/authority/src/activation_manager/mod.rs
+++ b/crates/consensus/authority/src/activation_manager/mod.rs
@@ -251,10 +251,7 @@ impl ConditionList {
     /// Returns true only when all conditions are satisfied, indicating
     /// that the network is ready to activate the upgrade.
     pub fn all_passing(&self) -> bool {
-        self.comp_req &&
-            self.aye_approval_req &&
-            self.comp_approval_req &&
-            self.block_height_req
+        self.comp_req && self.aye_approval_req && self.comp_approval_req && self.block_height_req
     }
 }
 
--- a/crates/consensus/authority/src/activation_manager/mod.rs
+++ b/crates/consensus/authority/src/activation_manager/mod.rs
@@ -233,7 +233,7 @@ struct NetworkUpgrade {
 #[derive(Debug, Clone, Copy, PartialEq)]
 pub struct ConditionList {
     /// Whether the validator is compliant with the upgrade
-    compliant_req: bool,
+    comp_req: bool,
 
     /// Whether the quorum approval rate for Aye votes has been reached
     aye_approval_req: bool,
@@ -251,7 +251,7 @@ impl ConditionList {
     /// Returns true only when all conditions are satisfied, indicating
     /// that the network is ready to activate the upgrade.
     pub fn all_passing(&self) -> bool {
-        self.compliant_req &&
+        self.comp_req &&
             self.aye_approval_req &&
             self.comp_approval_req &&
             self.block_height_req
@@ -729,7 +729,7 @@ where
         debug_assert!(total_votes >= upgrade.min_validator_count);
 
         let list = ConditionList {
-            compliant_req: upgrade.is_compliant,
+            comp_req: upgrade.is_compliant,
             aye_approval_req: aye_approval >= upgrade.quorum,
             comp_approval_req: comp_approval >= upgrade.quorum,
             block_height_req: block_height >= upgrade.target_height,
--- a/crates/consensus/authority/src/activation_manager/mod.rs
+++ b/crates/consensus/authority/src/activation_manager/mod.rs
@@ -1,0 +1,791 @@
+//! # Activation Manager
+//!
+//! Coordinates and manages network protocol upgrades in the Botanix blockchain.
+//!
+//! The activation manager is responsible for safely transitioning the blockchain
+//! network to new protocol versions without causing chain splits or consensus
+//! failures. It implements a distributed voting and activation mechanism that
+//! allows validators to signal support, prepare for, and eventually activate
+//! upgrades in a coordinated manner.
+//!
+//! ## Overview
+//!
+//! Network upgrades in Botanix follow a two-phase approach:
+//! 1. **Signaling Phase**: Validators vote on upgrade proposals to gauge support
+//! 2. **Compliance Phase**: Validators indicate readiness to process upgraded blocks
+//!
+//! An upgrade activates only when all of the following conditions are met:
+//! - A sufficient percentage of validators have voted "Aye" (meeting the quorum approval rate)
+//! - A sufficient percentage of validators are compliant with the upgrade (meeting the quorum
+//!   approval rate)
+//! - A minimum number of distinct validators have participated in voting
+//! - The current block height has reached or passed the scheduled target height
+//! - The local node is compliant with the upgrade
+//!
+//! ## Integration with CometBFT
+//!
+//! The activation manager integrates with CometBFT's ABCI at three critical points:
+//!
+//! 1. **Block Proposal** (`on_prepare_proposal`): Determines which runtime version to use and
+//!    includes the proposer's vote on pending upgrades.
+//!
+//! 2. **Block Validation** (`on_process_proposal`): Validates incoming block proposals based on
+//!    their version and the current upgrade state.
+//!
+//! 3. **Block Finalization** (`on_finalize_block`): Tracks votes, finalizes blocks, and handles
+//!    version transitions when upgrades are activated.
+//!
+//! ## Configuration Modes
+//!
+//! Nodes can be configured in three modes:
+//!
+//! - **Ignore Upgrades**: Default mode where the node doesn't participate in upgrades and rejects
+//!   blocks with versions other than the current active version.
+//!
+//! - **Signal Only**: The node votes on upgrades but doesn't accept upgraded blocks, allowing
+//!   validators to signal intentions before updating their software.
+//!
+//! - **Accept Upgrades**: The node votes on and accepts upgrades when conditions are met, enabling
+//!   full participation in the upgrade process.
+//!
+//! ## Vote Tracking and Expiration
+//!
+//! The manager tracks votes from validators and calculates support approval rates for:
+//! - Aye votes (percentage of validators voting in favor)
+//! - Compliant validators (percentage of validators ready to upgrade)
+//!
+//! Votes automatically expire after a configured retention period (default: 30 days)
+//! to ensure that upgrade decisions reflect recent consensus rather than outdated votes.
+//!
+//! ## Usage
+//!
+//! Nodes typically start in the "Ignore Upgrades" mode by default. When a potential
+//! upgrade is discussed, validators can switch to "Signal Only" mode to participate in
+//! voting. Once a new node version with upgrade support is released, validators can
+//! switch to "Accept Upgrades" mode to fully participate in the upgrade process.
+
+use crate::comet_bft::non_deterministic_data::NetworkUpgradePayload;
+use reth_provider::{ActivationManagerReaderWriter, ProviderResult};
+use std::sync::{Arc, RwLock};
+
+// Reexports
+pub use reth_db::models::activation_manager::{RuntimeVersion, Vote};
+
+#[cfg(test)]
+mod tests;
+
+/// The minimum required quorum percentage for network upgrades.
+///
+/// This represents the minimum percentage (67%) of validators that must:
+/// 1. Vote "Aye" for an upgrade proposal
+/// 2. Be configured to accept the upgrade
+///
+/// This approval rate ensures a supermajority consensus for any network upgrade,
+/// reducing the risk of chain splits while allowing for a reasonable approval rate
+/// that can be achieved in practice.
+///
+/// The value is set at 67% to align with CometBFT's 2/3 consensus approval rate,
+/// ensuring that network upgrades only activate when they have similar levels
+/// of support as other consensus decisions.
+pub const MIN_QUORUM: usize = 67;
+
+/// The minimum validator count used for calculating approval rates.
+///
+/// This constant sets a minimum denominator when calculating approval rates,
+/// ensuring that upgrades require broad support even when few validators have
+/// explicitly voted. By using this minimum in the denominator, we prevent
+/// scenarios where a small number of active validators could reach the required
+/// approval percentage without true network-wide support.
+pub const MIN_VALIDATOR_COUNT: usize = 3;
+// (TODO lamafab): This should be increased with time as we move towards larger
+// and dynamic federations, eventually.
+
+/// The period (in blocks) for which votes are retained and considered valid.
+///
+/// Votes older than this many blocks are automatically pruned and no longer
+/// count toward quorum calculations. This ensures that upgrade decisions
+/// reflect the current state of the network and prevents old votes from
+/// influencing current decisions.
+///
+/// At an average rate of 12 blocks per minute (5 second blocks), this equals:
+/// - 518,400 blocks
+/// - 43,200 minutes
+/// - 720 hours
+/// - 30 days
+///
+/// This extended retention period allows validators sufficient time to:
+/// 1. Signal support for an upgrade
+/// 2. Update their software to prepare for the upgrade
+/// 3. Switch their configuration to accept the upgrade
+///
+/// While ensuring votes eventually expire if an upgrade does not proceed.
+pub const VOTE_RETENTION_PERIOD: u64 = 518_400;
+
+/// The decision returned by the activation manager during the prepare proposal
+/// phase.
+///
+/// This struct represents the decision made by a validator when preparing a new
+/// block proposal. It contains the runtime version to use for the block, any
+/// vote the validator wishes to include in the proposal, and the current state
+/// of upgrade conditions.
+#[derive(Debug, Clone, PartialEq)]
+pub struct OnPrepareProposalDecision {
+    /// The runtime version that should be used for the block proposal. This
+    /// will be either the active version or an upgrade version if conditions
+    /// are met.
+    pub version: RuntimeVersion,
+
+    /// The validator's vote on a pending upgrade to include in the proposal.
+    /// This is included in the Non-Deterministic Data (NDD) of the block
+    /// proposal. If None, no vote is included in the proposal.
+    pub vote: Option<NetworkUpgradePayload>,
+
+    /// The current state of all upgrade conditions. This is used for diagnostic
+    /// purposes to understand why an upgrade is being proposed or not proposed.
+    pub conditions: Option<ConditionList>,
+}
+
+/// The decision returned by the activation manager when processing a block
+/// proposal.
+///
+/// This enum represents whether a validator should accept or reject a proposed
+/// block based on the runtime version and network upgrade conditions.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum OnProcessProposalDecision {
+    /// Accept and process the block proposal.
+    ///
+    /// This is returned when the block version is valid (either the active
+    /// version or an upgrade version that meets all conditions).
+    Process {
+        /// The runtime version of the block being processed
+        version: RuntimeVersion,
+        /// The current state of all upgrade conditions (for diagnostic
+        /// purposes)
+        conditions: Option<ConditionList>,
+    },
+
+    /// Reject the block proposal.
+    ///
+    /// This is returned when the block version is invalid (not the active
+    /// version and not an upgrade version that meets all conditions).
+    RejectBlock {
+        /// The runtime version of the block being rejected
+        version: RuntimeVersion,
+        /// The current state of all upgrade conditions (for diagnostic purposes)
+        conditions: Option<ConditionList>,
+    },
+}
+
+/// The decision returned by the activation manager when finalizing a block.
+///
+/// This enum represents whether a validator should finalize a block or reject
+/// it as a dead end (meaning the validator cannot continue on this chain).
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum OnFinalizeBlockDecision {
+    /// Finalize the block and continue processing the chain.
+    ///
+    /// This is returned for blocks with versions the validator can process.
+    Finalize {
+        /// The runtime version of the block being finalized
+        version: RuntimeVersion,
+    },
+
+    /// Reject the block as a dead end that the validator cannot follow.
+    ///
+    /// This is returned for blocks with versions the validator explicitly
+    /// refuses to support or cannot process. This results in a consensus split.
+    RejectBlockDeadEnd {
+        /// The runtime version of the block being rejected
+        version: RuntimeVersion,
+    },
+}
+
+/// Internal representation of a network upgrade configuration.
+///
+/// This struct holds all the parameters and state related to a pending network
+/// upgrade that a validator is tracking.
+struct NetworkUpgrade {
+    /// The runtime version of the proposed upgrade
+    version: RuntimeVersion,
+
+    /// The percentage approval rate (0-100) that must be reached for both
+    /// Aye votes and compliant validators to activate the upgrade
+    quorum: usize,
+
+    /// The minimum number of distinct validators that must participate
+    /// in voting for the upgrade to be considered valid
+    min_validator_count: usize,
+
+    /// The minimum block height at which the upgrade can activate
+    target_height: u64,
+
+    /// This validator's vote on the upgrade proposal (Aye/Nay)
+    our_vote: Vote,
+
+    /// Whether this validator is compliant with the upgrade
+    is_compliant: bool,
+}
+
+/// A list of boolean flags indicating the status of each upgrade condition.
+///
+/// This struct tracks whether each condition required for a network upgrade is
+/// currently satisfied. It's used for diagnostics and decision making.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct ConditionList {
+    /// Whether the validator is compliant with the upgrade
+    compliant_req: bool,
+
+    /// Whether the quorum approval rate for Aye votes has been reached
+    aye_approvals_req: bool,
+
+    /// Whether the quorum approval rate for compliant validators has been reached
+    comp_approval_req: bool,
+
+    /// Whether the current block height is at or above the target height
+    block_height_req: bool,
+}
+
+impl ConditionList {
+    /// Checks if all upgrade conditions are passing.
+    ///
+    /// Returns true only when all conditions are satisfied, indicating
+    /// that the network is ready to activate the upgrade.
+    pub fn all_passing(&self) -> bool {
+        self.compliant_req &&
+            self.aye_approvals_req &&
+            self.comp_approval_req &&
+            self.block_height_req
+    }
+}
+
+/// Builder for constructing an `ActivationManager` with specific configuration.
+///
+/// This builder provides methods to configure how a validator will handle
+/// network upgrades - whether it will ignore them, signal votes without
+/// being compliant, or fully accept upgrades when conditions are met.
+pub struct ActivationManagerBuilder<DB> {
+    /// The database client implementing ActivationManagerReaderWriter
+    client: DB,
+
+    /// The currently active runtime version
+    active_version: RuntimeVersion,
+
+    /// How long votes are retained in blocks before expiring
+    vote_retention_period: u64,
+
+    /// Configuration for a pending network upgrade, if any
+    upgrade: Option<NetworkUpgrade>,
+}
+
+impl<DB> ActivationManagerBuilder<DB>
+where
+    DB: ActivationManagerReaderWriter,
+{
+    /// Creates a new ActivationManagerBuilder with specified database client and active version.
+    ///
+    /// This is the primary constructor for initializing the builder with a storage backend
+    /// and the current active runtime version.
+    ///
+    /// # Parameters
+    /// * `client` - The database client implementing ActivationManagerReaderWriter
+    /// * `active_version` - The currently active runtime version
+    ///
+    /// # Returns
+    /// * A new ActivationManagerBuilder configured with default settings
+    pub fn new(client: DB, active_version: RuntimeVersion) -> Self {
+        ActivationManagerBuilder {
+            client,
+            active_version,
+            vote_retention_period: VOTE_RETENTION_PERIOD,
+            upgrade: None,
+        }
+    }
+
+    /// Sets a custom vote retention period.
+    ///
+    /// This method allows customizing how long votes are retained before
+    /// being pruned. The default is `VOTE_RETENTION_PERIOD` (518,400 blocks).
+    ///
+    /// # Parameters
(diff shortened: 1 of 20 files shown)


lamafab:
Unit test coverage:

### `accept_lagging_validator.rs`
```rust
/// Tests that a lagging validator can join the network after an upgrade has occurred.
///
/// This test verifies that:
/// 1. Two validators (ALICE, BOB) can successfully activate an upgrade when they meet all conditions
/// 2. A third validator (EVE) who joins later and is configured to accept the upgrade can 
///    successfully process blocks with the new version
/// 3. The lagging validator correctly rejects upgraded blocks during the process_proposal phase
///    but accepts them during finalize_block phase
/// 4. After finalizing the first upgraded block, the lagging validator properly updates its
///    active version and can fully participate in the network
```

### `reject_ignored_upgrade.rs`
```rust
/// Tests that a validator can reject an upgrade that it ignores.
///
/// This test verifies that:
/// 1. Two validators (ALICE, BOB) vote for and are willing to accept an upgrade
/// 2. A third validator (EVE) ignores the upgrade entirely
/// 3. When conditions are met (2 validators required, height reached), ALICE proposes upgraded blocks
/// 4. EVE consistently rejects these upgraded blocks in both the process_proposal and finalize_block phases
/// 5. This leads to a consensus split where ALICE and BOB operate on the upgraded chain
///    while EVE rejects those blocks as dead ends
```

### `reject_mismatched_vote.rs`
```rust
/// Tests that validators tracking different upgrade versions don't count each other's votes.
///
/// This test verifies that:
/// 1. Two validators (ALICE, BOB) vote for and track upgrade version 2.0
/// 2. A third validator (EVE) votes for a different upgrade version (3.0)
/// 3. The validators only count votes for their specific tracked version
/// 4. Even though all validators are voting "Aye", the upgrade doesn't activate because
///    from each perspective, the minimum validator requirement is not met
/// 5. This protects the network from confusion when multiple potential upgrades are being discussed
```

### `reject_outdated_or_unsupported_version.rs`
```rust
/// Tests that validators reject blocks with outdated or unsupported versions.
///
/// This test verifies that:
/// 1. All validators successfully activate an upgrade to version 2.0
/// 2. EVE is then reset to use the previous version (1.0)
/// 3. When EVE proposes blocks with the outdated version:
///    - ALICE and BOB reject these blocks during process_proposal but accept them during finalize_block
///    - This behavior allows historical sync while preventing outdated block production
/// 4. When ALICE proposes blocks with the upgraded version:
///    - EVE rejects these blocks during both process_proposal and finalize_block
///    - This creates a consensus split where EVE cannot follow the upgraded chain
```

### `vote_expiration.rs`
```rust
/// Tests that votes expire after the retention period and are no longer counted.
///
/// This test verifies that:
/// 1. All validators vote for and are willing to accept an upgrade
/// 2. With a reduced vote retention period of 10 blocks, votes begin to expire
/// 3. By block 12, BOB's vote (cast in block 1) expires and is no longer counted
/// 4. By block 13, EVE's vote (cast in block 2) also expires
/// 5. Even when the target height is reached, the upgrade doesn't activate because
///    expired votes reduce the validator count below the minimum requirement
/// 6. This ensures that upgrade decisions reflect recent consensus rather than outdated votes
```

### `vote_majority_nay_accepting_and_reject.rs`
```rust
/// Tests that an upgrade is rejected when most validators vote Nay despite accepting.
///
/// This test verifies that:
/// 1. All validators are configured to accept an upgrade (accepting = true)
/// 2. ALICE votes Aye, but BOB and EVE vote Nay
/// 3. Despite having 100% of validators accepting the upgrade, it does not activate
///    because the Aye vote threshold (only 34%) is below the required quorum
/// 4. This demonstrates that both acceptance and explicit Aye votes are required
///    for an upgrade to activate
```

### `vote_nay_and_accept.rs`
```rust
/// Tests that an upgrade activates even with a validator voting Nay but accepting.
///
/// This test verifies that:
/// 1. Two validators (ALICE, BOB) vote Aye for an upgrade
/// 2. One validator (EVE) votes Nay but is configured to accept the upgrade
/// 3. The upgrade still activates because:
///    - All validators are accepting the upgrade (100% acceptance)
///    - Aye votes reach 67%, meeting the quorum requirement
/// 4. After activation, all validators, including EVE, process blocks with the new version
/// 5. This shows that validators can signal disagreement while still accepting majority decisions
```

### `vote_nay_and_reject.rs`
```rust
/// Tests that a validator can vote Nay and reject an upgrade.
///
/// This test verifies that:
/// 1. Two validators (ALICE, BOB) vote Aye and are willing to accept an upgrade
/// 2. One validator (EVE) votes Nay and is configured to reject the upgrade (accepting = false)
/// 3. When conditions are met, ALICE proposes upgraded blocks
/// 4. EVE consistently rejects these upgraded blocks in both process_proposal and finalize_block phases
/// 5. This leads to a consensus split where ALICE and BOB operate on the upgraded chain
///    while EVE rejects those blocks as dead ends
```

### `wait_for_accepting.rs`
```rust
/// Tests the upgrade flow where validators first signal support, then later accept.
///
/// This test verifies that:
/// 1. All validators initially signal Aye votes but are not ready to accept the upgrade
/// 2. Despite unanimous Aye votes, the upgrade doesn't activate because no validators are accepting it
/// 3. After all validators update to accept the upgrade and additional voting occurs:
///    - By block 5, all conditions are met (100% Aye votes, 100% acceptance, target height reached)
///    - EVE proposes the first upgraded block which is accepted by all
/// 4. This demonstrates the two-phase upgrade process where validators can signal support
///    before they're technically ready to handle the upgrade
```

crates/storage/db-api/src/models/activation_manager.rs
```diff
@@ -0,0 +1,89 @@
+//! Activation manager models for network upgrades and voting.
+
+use reth_codecs::{add_arbitrary_tests, Compact};
+use serde::{Deserialize, Serialize};
+use std::cmp::Ordering;
+
+/// Represents a validator's vote on a network upgrade proposal.
```
rwlockbg, line 7: @lamafab Just a general question - does not commet support voting in a similar way as the state sync via abci.rs. If I remember correctly they had some voting mechanism too.

crates/storage/db-api/src/models/activation_manager.rs
```diff
@@ -0,0 +1,89 @@
...
+/// - If major versions are equal, compare minor versions
+///
+/// This ordered relationship is crucial for the upgrade process, as it ensures that:
+/// 1. Nodes can determine whether a version is an upgrade or downgrade
+/// 2. The activation manager can enforce one-way upgrade progression
+/// 3. Historical blocks during sync can be validated against appropriate version thresholds
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct RuntimeVersion(
+    /// Major version component, incremented for breaking changes (hard fork)
+    pub u16,
```
rwlockbg: Why cannot we do this a bit more rusty - :
`pub enum RuntimeVersion { 
  Major(u16),
  Minor(u16)
}`

crates/storage/db-api/src/models/activation_manager.rs
```diff
@@ -0,0 +1,89 @@
...
+        match self.0.cmp(&other.0) {
+            Ordering::Equal => self.1.cmp(&other.1),
+            ordering => ordering,
+        }
+    }
+}
+
+#[test]
+fn runtime_version() {
+    let v0_0 = RuntimeVersion(0, 0);
```
rwlockbg: I am pretty sure there was a nice crate doing all of that, ofc your impl is fine here, but maybe consider it for more flexibility?

crates/consensus/authority/src/activation_manager/mod.rs
```diff
@@ -0,0 +1,800 @@
...
+        // Then track the vote.
+        self.client.update_upgrading_vote(addr, vote.vote, vote.accepting, block_height)?;
+
+        // Prune old votes.
+        let prune_below = block_height.saturating_sub(self.vote_retention_period);
+        self.client.remove_upgrading_votes(prune_below)?;
+
+        Ok(())
+    }
+}
```
rwlockbg, line 788: I am missing a bit of context here - is this activation manager supposed to be activated somewhere and / or integrated into the comet-bft abci interface via the rpc methods ? If so, is this the subject of the next PR ?

crates/consensus/authority/src/activation_manager/mod.rs
```diff
@@ -0,0 +1,800 @@
...
+
+/// Builder for constructing an `ActivationManager` with specific configuration.
+///
+/// This builder provides methods to configure how a validator will handle
+/// network upgrades - whether it will ignore them, signal votes without
+/// accepting, or fully accept upgrades when conditions are met.
+pub struct ActivationManagerBuilder<DB> {
+    /// The database client implementing ActivationManagerReaderWriter
```
rwlockbg, line 264: Is this activation manager rw trait defacto going to be implemented onto db entities in a subsequent PR ?

crates/consensus/authority/src/activation_manager/mod.rs
```diff
@@ -0,0 +1,798 @@
...
+        &self,
+        block_height: u64,
+    ) -> ProviderResult<OnPrepareProposalDecision> {
+        let mut our_vote = None;
+        let mut conditions = None;
+
+        // Check if we're tracking a network upgrade.
+        let maybe_upgrade = self.upgrade.read().expect("poisoned lock");
+        if let Some(upgrade) = maybe_upgrade.as_ref() {
+            assert!(upgrade.quorum >= MIN_QUORUM);
```
rwlockbg, line 525: Do we want to panic at these points ? Should not the node simply live further if potential upgrades are violated at any point ? To me it makes sense to just do tracing::error and maybe throw (?) the error to the higher instance driving the manager and let it be handled there. Maybe add some error handling ?

crates/consensus/authority/src/activation_manager/mod.rs
```diff
@@ -0,0 +1,798 @@
...
+    /// # Panics
+    /// * If the upgrade version is not greater than the active version
+    pub fn build_signal_network_upgrade(
+        mut self,
+        upgrade_version: RuntimeVersion,
+        our_vote: Vote,
+    ) -> ActivationManager<DB> {
+        assert!(self.active_version < upgrade_version);
+
+        #[rustfmt::skip]
```
rwlockbg, line 362: Since we have plenty of these - cannot we enable them all on file level at the very top:  `#![ .... ] ` ?

rwlockbg (commented):
Great PR - so far this is the state machine logic basically and the underlying mechanism for accepting an upgrade. I presume next PRs will be actual integration. Left some comments, but overall looking nice and nice tests too!

crates/storage/db-api/src/models/activation_manager.rs
rwlockbg: or is this going to be the subject of the next integration round ?

crates/consensus/authority/src/activation_manager/mod.rs
lamafab: Indeed it is, in part 4 (upcoming PR) -> https://github.com/botanix-labs/botanix/issues/928

crates/consensus/authority/src/activation_manager/mod.rs
lamafab: Yes, in part 3 (upcoming PR) -> https://github.com/botanix-labs/botanix/issues/928

crates/consensus/authority/src/activation_manager/mod.rs
lamafab: Those types of panics can only occur if the `ActivationManager` is incorrectly implemented by us. This assert is technically not required (resp. outdated) since it's checked earlier. I will revisit this.

crates/consensus/authority/src/activation_manager/mod.rs
lamafab: I would not skip `fmt` on the entire file, only for individual components where the skip makes the it more readable.

crates/storage/db-api/src/models/activation_manager.rs
lamafab: Afaik, CometBFT does not provide a voting mechanism for such application specific logic. In upcoming PRs, the vote is extracted in Comets' `on_finalize` method via the NDD and then passed to the `ActivationManager`. In this PR, there is now that stand-alone `NetworkUpgradePayload` structure, which will later be included in the NDD. That's how votes are cast.

More details in the spec: https://github.com/botanix-labs/botanix/issues/928#issuecomment-2707699416

crates/consensus/authority/src/activation_manager/mod.rs
```diff
@@ -0,0 +1,798 @@
...
+            self.active_version < upgrade_version,
+            "upgrade version is older than the active version"
+        );
+
+        assert!(quorum >= MIN_QUORUM, "minimum validator count is less than the minimum required");
+
+        #[cfg(not(test))]
+        assert!(
+            min_validator_count >= MIN_VALIDATOR_COUNT,
+            "minimum validator count is less than the minimum required"
```
SatwikPrabhu: Just wondering what is the need to have these panics here instead of gracefully handling error cases?

crates/consensus/authority/src/activation_manager/mod.rs
```diff
@@ -0,0 +1,798 @@
...
+                })
+            }
+
+            conditions = Some(c);
+        }
+
+        // Default case: propose a block with the current active version. We
+        // include our vote (if any) to signal our upgrade intentions, even when
+        // not yet proposing upgraded blocks.
+        let active_version = self.active_version.read().expect("poisoned lock");
```
SatwikPrabhu, line 551: Maybe good not to panic here?

crates/consensus/authority/src/activation_manager/mod.rs
```diff
@@ -0,0 +1,798 @@
...
+    /// * `OnProcessProposalDecision::RejectBlock` if the block should be rejected
+    pub fn on_process_proposal(
+        &self,
+        block_height: u64,
+        block_version: RuntimeVersion,
+    ) -> ProviderResult<OnProcessProposalDecision> {
+        let mut conditions = None;
+
+        // Check if block matches a pending upgrade we're tracking.
+        let maybe_upgrade = self.upgrade.read().expect("poisoned lock");
```
SatwikPrabhu, line 584: Same as https://github.com/botanix-labs/Macbeth/pull/611/files#r2001478439. I see that there are multiple such instances throughout. Maybe revisit all of them.

crates/consensus/authority/src/activation_manager/mod.rs
lamafab: Basically; https://github.com/botanix-labs/Macbeth/pull/611#discussion_r2001315861

Those panics only happen when the ActivationManager is incorrectly implemented by us, and it happens immediately on startup (setup happens in a future PR).

They're a security precaution to make sure the upgrade conditions are not too loose. It's not expected that those could ever fail, unlike cases where you process user input, for example.

crates/consensus/authority/src/activation_manager/mod.rs
rwlockbg: no objections on my end, just a suggestion

crates/consensus/authority/src/activation_manager/mod.rs
rwlockbg: got it

crates/consensus/authority/src/activation_manager/mod.rs
rwlockbg: got it

crates/storage/db-api/src/models/activation_manager.rs
rwlockbg: fine with me

crates/consensus/authority/src/activation_manager/mod.rs
lamafab: How do you see those panics going wrong? The scopes are very short, and I don't see a way where a poisoned lock could be encountered.

crates/storage/db-api/src/models/activation_manager.rs
lamafab: We could you something like [`semver`](https://crates.io/crates/semver), but considering that we're only ever dealing with soft-forks (`MINOR`) and hard-forks (`MAJOR`), I don't think we need that third `PATCH` versioning option.

Can you clarify by what you mean by flexibility, respectively in what scenarios that would be useful?

lamafab: [BIP0008](https://github.com/bitcoin/bips/blob/master/bip-0008.mediawiki#specification) contains an extra `"name"` parameter, for example, but IMO that's kind of unnecessary(?)

crates/consensus/authority/src/activation_manager/mod.rs
SatwikPrabhu: That is true I guess. Should have been more mindful of what we are panicing for here

crates/storage/db-api/src/models/activation_manager.rs
rwlockbg: It was just a suggestion, dont want you to change any code for that. Closing ehre.

scottmillner:
love the test fixture and test helper methods!! 🔥 🔥 🔥 
awesome work with the very well defined spec and it's implementation

putting my approval but little nit: do prefer @rwlockbg suggestion below
```
pub enum RuntimeVersion {  Major(u16), Minor(u16) }
```

scottmillner (dismissed):

arminsabouri:
@lamafab Do you mind squashing relevant commits (typos, lint, trigger CI). Ideally these changes are made into the commits that introduce them

lamafab:
@0xBEEFCAF3 

> @lamafab Do you mind squashing relevant commits (typos, lint, trigger CI). Ideally these changes are made into the commits that introduce them

I'll probably need to rewrite the entire history anyway after following up on Satwiks' recommendations  - I"ll cherry-pick all changes and label them accordingly, then force-push. ETA EOD.

lamafab:
Regarding: https://github.com/botanix-labs/Macbeth/actions/runs/13976740327/job/39132293647?pr=611

```
TRY 1 TRMNTG [>120.000s] reth-authority-consensus comet_bft::abci::tests::test_finalize_block_with_signed_tx
   TRY 1 TMT [ 120.020s] reth-authority-consensus comet_bft::abci::tests::test_finalize_block_with_signed_tx
──── TRY 1 STDOUT:       reth-authority-consensus comet_bft::abci::tests::test_finalize_block_with_signed_tx

running 1 test
test comet_bft::abci::tests::test_finalize_block_with_signed_tx has been running for over 60 seconds
──── TRY 1 STDERR:       reth-authority-consensus comet_bft::abci::tests::test_finalize_block_with_signed_tx

thread 'comet_bft::abci::tests::test_finalize_block_with_signed_tx' panicked at crates/consensus/authority/src/comet_bft/abci.rs:1351:31:
Error building block in finalize block: Validation(EVM { hash: 0xb222209cbdf13b5574f4fb7ae7af4f0bf0abd614bf4c7de67b9ecfa117aca874, error: Transaction(InvalidChainId) })
stack backtrace:
```

I didn't do any changes in this area, and when running the command locally all tests pass:

```
cargo nextest run \
    --locked --features "ethereum conflicting_input" \
    --workspace --exclude example-custom-evm --exclude example-stateful-precompile --exclude ef-tests \
    -E "kind(lib) | kind(bin) | kind(proc-macro)"
```

I'm just going to re-run.

rwlockbg (dismissed):
I am approving it but you need to get hte pipelines green.

SatwikPrabhu:
Btw. I'm also getting the same failure @lamafab https://github.com/botanix-labs/Macbeth/actions/runs/13982320523/job/39149966590?pr=620

SatwikPrabhu (dismissed):

lamafab:
I just renamed a some fields in `ConditionList`, minor cosmetic changes.

scottmillner (approved):
i was able to get all the int tests to pass locally
so when there's another approval happy for it to go in

SatwikPrabhu (approved):

lamafab:
# Botanix Network Upgrade Specification

(_This spec has been revised with Claude.AI_)

## 1. Introduction

This document outlines the specification for the Network Upgrade mechanism in the Botanix ecosystem. This mechanism enables coordinated updates to the consensus rules and network protocol while preserving blockchain safety and minimizing the risk of chain splits.

## 2. Terminology

- **Active Version**: The currently enforced runtime version.
- **Upgrade Version**: A proposed new runtime version that may become active after meeting required conditions.
- **Runtime Version**: A pair of values `(HARD, SOFT)` where `HARD` represents breaking changes and `SOFT` represents backward-compatible changes.
- **Vote**: A validator's explicit signal regarding a proposed upgrade (Aye, Nay, or Absent).
- **Compliance**: A validator's readiness to process blocks with the upgrade version when conditions are met.
- **Quorum**: The percentage of validators that must support an upgrade for it to be activated.
- **Target Height**: The minimum block height at which an upgrade can activate if quorum is reached.

## 3. Components

### 3.1 ActivationManager

The `ActivationManager` MUST track network upgrade proposals, calculate approval rates, and determine when proposed upgrades should become active. It MUST interface with the ABCI component during block production and validation.

### 3.2 Network Upgrade Payload

Each block proposal MAY include a `NetworkUpgradePayload` within its Non-Deterministic Data (NDD) transaction, containing:

- **version**: The proposed Runtime Version
- **vote**: The block proposer's vote (Aye/Nay/Absent)
- **is_compliant**: Whether the proposer is ready to accept the upgrade

## 4. Configuration

### 4.1 Constants

The system defines the following global constants:

- **MIN_QUORUM**: 67% - The minimum acceptance approval rate required for an upgrade
- **MIN_VALIDATOR_COUNT**: 3 - The minimum number of distinct validators that must participate in voting
- **VOTE_RETENTION_PERIOD**: 518,400 blocks - The period for which votes are retained (approximately 30 days at 12 blocks per minute)

These global constants ensure network security by requiring a supermajority consensus across a minimum number of validators before any upgrade can proceed, while maintaining a reasonable voting window.

### 4.2 Default Configuration

By default and for the majority of a node's lifetime, the `ActivationManager` MUST be built with:

- `build_ignore_network_upgrade()`: Continue with current version, ignore all upgrade proposals

This configuration:
- Does not participate in any upgrade voting process
- Leaves the `NetworkUpgradePayload` in the NDD empty
- MUST reject any blocks with upgraded versions
- Results in a consensus split if an upgrade activates on the network
- Node operators SHOULD be aware that continuing operation after a network upgrade without upgrading results in undefined network behavior

### 4.3 Upgrade Signaling

When a community-driven upgrade proposal is being discussed, validators MAY signal their intentions by configuring the `ActivationManager` using:

- `build_signal_network_upgrade(upgrade_version, vote)`: Signal vote without accepting the upgrade

Key characteristics:
- This configuration SHOULD be possible without requiring a software update
- The runtime version and vote parameters SHOULD be configurable via CLI flags, environment variables, or configuration files
- The `NetworkUpgradePayload` in the NDD does include:
  - The specified `upgrade_version`
  - The validator's `vote`
  - The `is_compliant` parameter set to `false`, indicating readiness to signal but not to process upgraded blocks
- Blocks proposing the upgrade version MUST be rejected even if quorum is reached

### 4.4 Upgrade Acceptance

If an upgrade proposal reaches community consensus and is scheduled for activation, **the maintainers MUST release a new Botanix node version** with the `ActivationManager` using:

- `build_ACCEPT_network_upgrade(upgrade_version, quorum, target_height, vote)`: Signal vote and accept the upgrade if conditions are met

Key characteristics:
- The release MUST include all necessary logic, code changes, and migrations required to fully support the new runtime version.
- The upgrade parameters MUST be hardcoded in this release to ensure consistency.
- Sets `NetworkUpgradePayload` with both the validator's vote AND `is_compliant=true`
- Enables the validator to process blocks with the upgraded version once all upgrade conditions are met
- Automatically transitions the node to the new version once the first upgraded block is finalized
- Prunes voting data after the transition to start fresh with the new version

Validators MUST update to this new version before all upgrade conditions are met to participate in the coordinated upgrade process.

## 5. Upgrade Conditions

Blocks with the upgraded version MUST only be proposed and backed if ALL of the following are true:
   - The validator is configured to accept the upgrade
      - Config parameter: `is_compliant`
   - The network has reached sufficient quorum of _Aye_ votes for the upgrade
      - Config parameter: `quorum`
      - Must be greater than or equal to MIN_QUORUM
   - The network has reached sufficient quorum of **compliant** validators for the upgrade
      - Config parameter: `quorum`
      - Must be greater than or equal to MIN_QUORUM
   - The current block height is at or above the scheduled target height
      - Config parameter: `target_height`

### 5.1 Approval Rate Calculation

The `ActivationManager` maintains two separate approval rate calculations:
1. **Aye Approval Rate**: Percentage of validators voting _Aye_, regardless of compliance status.
2. **Compliance Approval Rate**: Percentage of validators with `is_compliant=true`, regardless of vote.

By requiring both approval rates to meet quorum, we ensure upgrades only proceed when there is both sufficient community support AND operational readiness. This allows validators to signal support for an upgrade they aren't yet ready to implement, or to prepare for an upgrade they oppose (but will follow if passed by the community).

For the calculation, the following values must be retrieved:

$$
\begin{align*}
V &= \set{v_1, v_2, \ldots, v_n} \text{ the set of all votes} \\
Y &= \set{v \in V \mid v \text{ is \textit{Aye}}} \text{ the subset of Aye votes} \\
C &= \set{v \in V \mid v \text{ is \textit{compliant}}} \text{ the subset of \textit{compliant} votes} \\
y &= |Y| \text{ the number of \textit{Aye} votes} \\
c &= |C| \text{ the number of \textit{compliant} validators} \\
n &= |V| \text{ the total number of votes} \\
m &= \texttt{MIN-VALIDATOR-COUNT} \text{ the minimum required validator count} \\
T &= \max(n, m) \text{ the effective total for percentage calculation}
\end{align*}
$$

The exact formula for calculating the **aye approval rate** is:

$$
\begin{equation}
t^{y} = \left\lceil \frac{y \times 100}{T} \right\rceil
\end{equation}
$$

The exact formula for calculating the **compliance approval rate** is:

$$
\begin{equation}
t^{c} = \left\lceil \frac{c \times 100}{T} \right\rceil
\end{equation}
$$

The quorum is considered met if and only if:
1. $t^{y} \geq {}$ `quorum` - The percentage of Aye votes meets or exceeds the minimum required threshold
2. $t^{c} \geq {}$ `quorum` - The percentage of compliant validators meets or exceeds the minimum required threshold

The ceiling function $\lceil \cdot \rceil$ in the formula ensures we round up to the nearest percentage point, implementing ceiling integer division. Using $\max(n, m)$ as the denominator ensures that when there are fewer validators voting than the minimum required count, we calculate the percentage against the minimum count, making it more difficult to reach quorum with only a few validators.

## 6. Process Flow for Compliant Validators

This process flow describe how `ActivationManager` must be implemented into CometBFT's block production and finalization methods.

### 6.2 Block Preparation

1. During CometBFTs' `prepare_proposal`, an `ActivationManager` MUST:
   - Return the appropriate block version to propose (active or upgrade)
   - Include the proposer's vote on pending upgrades, if any.
   - Propose an upgraded block only when all upgrade conditions are met.

### 6.3 Block Validation

1. During CometBFTs' `process_proposal`, all validators MUST:
   - Accept blocks with the active version
   - Accept blocks with an upgrade version only when all upgrade conditions are met.
   - Reject blocks with unknown versions or when all upgrade conditions are not met.

2. Validators configured to reject an upgrade MUST NOT process blocks with the upgraded version, regardless of network support levels.

### 6.4 Block Finalization

1. During CometBFTs' `finalize_block`, validators MUST:
   - Track the proposer's vote for network upgrades
   - Prune outdated votes by removing all votes from blocks older than (current_height - VOTE_RETENTION_PERIOD)
   - Accept any block version that is less than or equal to the active version
      - This mechanism is required for historical sync
   - Update their active version if an upgraded block is finalized
   - Reject blocks if explicitly configured to reject that upgrade
   - Clean up all voting data once a version transition occurs or when the upgrade block is rejected

2. When a validator finalizes a block with an upgraded version, **it MUST set that version as the new active version for all future block production and validation**.

3. Validators not accepting an upgrade MUST deliberately reject upgraded blocks, resulting in a consensus split and undefined network behavior.

#### 6.4.1 Syncing Considerations

The `on_finalize_block` method MUST NOT repeat upgrade condition checks for the following reasons:
   * Syncing nodes lack historical context about previous network state and voting patterns
   * Blocks finalized by CometBFT consensus have already been endorsed by a supermajority of validators
   * By the time a block reaches finalization, its validity including version transitions has been confirmed
   * Re-validating these conditions during sync would require maintaining and processing historical vote data that may no longer be available, and would introduce additional complexity

This separation of responsibilities ensures that nodes can reliably sync the canonical chain while maintaining the ability to reject future unwanted upgrades.

PR #610, opened 2025-03-05
Feature: Pegin proofs v1

SatwikPrabhu:
Pegin proof v1 based on the discussion in https://github.com/botanix-labs/botanix/issues/909

PR #612, opened 2025-03-09
Feature(consensus): Pegin Proofs v1

SatwikPrabhu:
Opened a new PR for this since the other PR became a mess because of erroneous squashes. Please refer to the discussions in https://github.com/botanix-labs/Macbeth/pull/610 for context and comments already provided. 

Here are the things still left to do in this:

1. Need to provide the correct hex for deserialize_pegin_metadata_v1 to match the new pegin meta v1 -- Done
3. Fix any clippy errors and rebase conflicts - Done
4. Add documentation for some new helper functions that I added - Done
5. Add integration test. As mentioned does not block merging this PR
6. Modify batch pegins int tests - Done

Feel free to add any other things in the above todo list.


crates/consensus/authority/src/excecution_utils.rs
```diff
@@ -347,12 +350,32 @@ pub(crate) mod authority_execution_utils {
             .with_bundle_update()
             .build();
 
+            let data_dir = if let Some(dir) = reth_node_core::dirs::database_path() {
```
arminsabouri: Why not use the DB provider provided above?

arminsabouri: This also assumes the default path -- I think. Why not just the provider factory paradigm. Its already there

crates/primitives/src/botanix/peg_contract.rs
```diff
@@ -152,20 +162,202 @@ pub struct PeginMeta {
     pub tx: bitcoin::Transaction,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct PeginMetaV1 {
+    pub v0: PeginMetaV0,
+    pub ref_block_hash: Option<B256>,
```
arminsabouri: Should this be optional?

arminsabouri: Should this be pub(struct)?

crates/primitives/src/botanix/peg_contract.rs
```diff
@@ -152,20 +162,202 @@ pub struct PeginMeta {
...
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum PeginMeta {
+    V0(PeginMetaV0),
+    V1(PeginMetaV1),
+}
```
arminsabouri: Not seeing the benefit of this enum. Pegins can either be v0 or v1. We impl the same methods (ser/deser) on both. V1 can call `self.serialize()` and then serialize the additional field after the buffer has been written to

crates/primitives/src/botanix/peg_contract.rs
```diff
@@ -152,20 +162,202 @@ pub struct PeginMeta {
...
+
 impl PeginMeta {
+    pub fn version(&self) -> u32 {
+        match self {
+            PeginMeta::V0(meta) => meta.version,
+            PeginMeta::V1(meta) => meta.v0.version,
+        }
+    }
```
arminsabouri: These setters and getters also seem unneccecary.  These fields can all be `pub(struct)`

crates/primitives/src/botanix/peg_contract.rs
```diff
@@ -152,20 +162,202 @@ pub struct PeginMeta {
...
+        for header in self.block_headers() {
             header.consensus_encode(&mut bytes).unwrap();
         }
-        self.merkle_proof.consensus_encode(&mut bytes).unwrap();
-        self.tx.consensus_encode(&mut bytes).unwrap();
+        self.merkle_proof().consensus_encode(&mut bytes).unwrap();
+        self.tx().consensus_encode(&mut bytes).unwrap();
+        if let Some(hash) = self.ref_block_hash() {
+            bytes.extend_from_slice(&hash.as_slice());
+        }
```
arminsabouri: This should be included in the v1 serialize method after calling inner serialize.
```rust
let mut  inner_serialized: vec[u8] = self.inner.serialize();
inner_serialized.extend_from_slice(self.ref_block_hash());
...
```

arminsabouri: Additionally, I have a strong prefrence to use rust-bitcoin consensus encode / decode through out the pegin proof serialization format. [u8; 32] should be supported

crates/primitives/src/botanix/peg_contract.rs
```diff
@@ -341,44 +637,105 @@ mod tests {
...
         };
+        
+        PeginMeta::V1(PeginMetaV1 {
+            v0,
+            ref_block_hash: Some(B256::from_slice(&[0; 32])),
+        })
+    }
+
+    #[test]
+    fn serialize_pegin_metadat_v0() {
```
arminsabouri:
```suggestion
    fn serialize_pegin_metadata_v0() {
```

arminsabouri (changes requested):
I'm three files deep. But I already had some changes I noticed needed to be made that may change the rest of the code.

crates/consensus/authority/src/excecution_utils.rs
SatwikPrabhu: I've given an explanation for this in https://github.com/botanix-labs/Macbeth/pull/610#discussion_r1984784777
The key issue is the use of different `Database` types in the authority consensus crate and the ethereum execution code. I did try to reconcile this difference, but wasn't able to do it as such.

crates/primitives/src/botanix/peg_contract.rs
SatwikPrabhu: Actually yeah. We can remove the optional here because its V1

crates/primitives/src/botanix/peg_contract.rs
SatwikPrabhu: > Pegins can either be v0 or v1. We impl the same methods (ser/deser) on both. V1 can call self.serialize() and then serialize the additional field after the buffer has been written to

Agree with this. 

> Not seeing the benefit of this enum.

I think we still need this enum because we need one `PeginMeta` type to assign to the `meta` field of `PeginData`

crates/consensus/authority/src/excecution_utils.rs
arminsabouri: Sorry I should have said factory. The factory can create a general DB provider not constraint by any traits. And it's already available in this scope

crates/ethereum/evm/src/execute.rs
```diff
@@ -426,7 +457,38 @@ where
...
+                            revert_address: pegin_data.account,
+                            revert_amount: pegin_data.amount,
+                        })? {
+                            let header = block.header;
+                            let package = header.botanix_consensus_package(self.bitcoin_network, self.bitcoind_factory.clone()).map_err(|_| MintContractError::InvalidPeginData {
+                                error: format!("Failed to get botanix consensus package"),
+                                revert_address: pegin_data.account,
+                                revert_amount: pegin_data.amount,
+                            })?;
+                            bitcoin_checkpoint = package.bitcoin_checkpoint;
```
lamafab, line 470: Since in v1 the checkpoint is user-provided; is there any possible security concern here? Afaik the `btc-server` will (should) prevent a UTXO from being minted twice, but since currently Pegins expire after 200 blocks, we have to make sure that the `btc-server` handles such cases accordingly. **The EVM Pegin validation logic itself does not check for UTXO reuse.**

EDIT: Meaning; the `btc-server` should NEVER allow a UTXO from being used more than once for a mint.

lamafab (commented):
Reviewed yesterday; LGTM, but see attached comment below.

lamafab:
Btw can we close this: https://github.com/botanix-labs/Macbeth/pull/610 ?

crates/consensus/authority/src/excecution_utils.rs
SatwikPrabhu: I think I have managed to do it this way by using the provider factory already available. Will wait and see if this leads to any test failures.

crates/primitives/src/botanix/peg_contract.rs
SatwikPrabhu: Tried what we discussed offline yesterday that we can decouple `PeginMeta` from `PeginData` as pass meta as a parameter in validate. Think the enum is still better because we need to have a common type/trait for the meta variable to be passed.

crates/primitives/src/botanix/mint_validation.rs
```diff
@@ -197,7 +197,7 @@ pub fn try_parse_mint_event(
         let mut proofs = Vec::new();
         let mut offset = 0;
         while offset < meta_bytes.len() {
-            let (proof, proof_size) =
+            let (proof, proof_size) =
```
arminsabouri: White space added?

crates/storage/codecs/derive/src/compact/generator.rs
```diff
@@ -141,32 +141,30 @@ fn generate_from_compact(fields: &FieldList, ident: &Ident, is_zstd: bool) -> To
     // If the type has compression support, then check the `__zstd` flag. Otherwise, use the default
     // code branch. However, even if it's a type with compression support, not all values are
     // to be compressed (thus the zstd flag). Ideally only the bigger ones.
-    is_zstd
```
arminsabouri: Where did this change come from?

crates/ethereum/evm/src/execute.rs
arminsabouri: The bitcoin block height provided during pegin prevents pegin replays. Please take a look at `minting.sol`

crates/ethereum/evm/src/execute.rs
```diff
@@ -59,54 +62,66 @@ use std::sync::Arc;
...
     bitcoin_network: bitcoin::Network,
+    provider: ProviderFactory<RethDB>,
 }
 
 /// Create a noop executor provider with chain spec
 pub fn create_noop_executor_provider(
     chain_spec: Arc<ChainSpec>,
-) -> EthExecutorProvider<MockBitcoindFactory> {
+) -> EthExecutorProvider<MockBitcoindFactory, Arc<TempDatabase<DatabaseEnv>>, EthEvmConfig> {
+    let db_path = tempfile::tempdir()
```
arminsabouri: unused?

crates/ethereum/evm/Cargo.toml
```diff
@@ -34,13 +34,17 @@ reth-execution-types = { workspace = true }
...
+reth-provider = { workspace = true }
 
 # Ethereum
 revm-primitives = { workspace = true }
 
 # tracing
 tracing = { workspace = true }
 
+# testing
+tempfile = { workspace = true }
```
arminsabouri: If its only used during testing please move to dev-deps. But it also doesnt seem like this dependency is needed at all.

clippy.toml
```diff
@@ -1,2 +1,2 @@
-msrv = "1.80"
+msrv = "1.80.0"
```
arminsabouri: Is this change related to pegin v1? If there are general clippy fixes they should live in a seperate commit or just a seperate PR. Shoving all this in one commit (esp a commit that touches consensus) makes to review process harder

crates/primitives/src/botanix/peg_contract.rs
```diff
@@ -152,7 +162,7 @@ pub struct PeginMeta {
     pub tx: bitcoin::Transaction,
 }
 
-impl PeginMeta {
+impl PeginMetaV0 {
```
arminsabouri, line 172: Lets leave `PeginMeta` untouches where its implicit thats its the first version.

crates/primitives/src/botanix/peg_contract.rs
```diff
@@ -216,6 +225,158 @@ impl PeginMeta {
...
+    pub inner: PeginMetaV0,
+    /// Reference block hash
+    pub ref_block_hash: B256,
+}
+
+impl PeginMetaV1 {
+    /// Serialize a pegin meta
+    pub fn serialize(&self) -> Vec<u8> {
+        let mut bytes = self.inner.serialize();
+        self.ref_block_hash.consensus_encode(&mut bytes).unwrap();
```
arminsabouri: Please handle errors

arminsabouri (changes requested):

crates/storage/codecs/derive/src/compact/generator.rs
SatwikPrabhu: This was a clippy error in the ci

crates/primitives/src/botanix/peg_contract.rs
SatwikPrabhu: Actually this is the way that we are currently serializing without handling errors. Should I add error handling for other fields in `PeginMeta` as well?

arminsabouri: Oof yes. Those should definetly be handled. I would **fix those in a seperate PR**. that should be an easy fix and can be included before the changes in this PR
https://github.com/botanix-labs/botanix/issues/930

crates/consensus/authority/src/excecution_utils.rs
SatwikPrabhu: Seems to be working

crates/primitives/src/botanix/peg_contract.rs
SatwikPrabhu: Done!

crates/primitives/src/botanix/peg_contract.rs
SatwikPrabhu: Done!

crates/primitives/src/botanix/peg_contract.rs
SatwikPrabhu: Done!

crates/primitives/src/botanix/mint_validation.rs
SatwikPrabhu: Fixed!

crates/ethereum/evm/src/execute.rs
SatwikPrabhu: Fixed!

crates/ethereum/evm/Cargo.toml
SatwikPrabhu: Fixed!

clippy.toml
SatwikPrabhu: Reverted!

crates/primitives/src/botanix/peg_contract.rs
SatwikPrabhu: Done!

SatwikPrabhu:
Rebased and CI is also green. 

Currently we have an enum `PeginMeta` that has the two versions for the meta. There is a possibility of improving the ergonomics of this design as pointed out by @0xBEEFCAF3 (discussed in the comments of this PR as well as some offline discussions) by decoupling `PeginMeta` from `PeginData` (since `PeginData` will not be versioned) and using a trait to represent both versions of the proof. I feel that it is better to explore and make this change in a separate PR rather rather than this PR. We can then merge this so that it can be part of the audit. I've created a separate ticket to track the above improvements --> https://github.com/botanix-labs/botanix/issues/931. Let me know your thoughts on this. Let me know if this sounds good.

scottmillner (approved):
approving so this can go into audit if team agrees
but i definitely want a separate pr that addresses @0xBEEFCAF3 comments on not needing an enum to designate which pegin version we're handling

ci is green and the desired functionality exists

bin/test-suite/src/suite/consensus/frost/test_frost_e2e.rs
```diff
@@ -140,18 +140,20 @@ pub async fn frost_e2e_stable(
...
-            gateway_address_response.aggregate_public_key.as_str(),
-        )
-        .expect("valid public key"),
-        tx: pegin_tx.clone(),
-        merkle_proof: pmt,
-        block_headers: headers,
-    };
+    let meta = PeginMeta::V0(
+        PeginMetaV0 {
+            version: 0,
```
lamafab, line 145: Use `PEGIN_META_VERSION_V0`, for consistency :)

lamafab (approved):
Noice 🚀

As Scott mentioned, open a separate PR to address cleanups and nitpicks.

scottmillner:
@SatwikPrabhu in another PR, let's use V1 in at least one int test (for example we can use V1 in `test_frost_e2e` since there is not test coverage for the new pegin version)

i created a ticket

https://github.com/botanix-labs/botanix/issues/932

bin/btc-server/src/coordinator/mod.rs
```diff
@@ -23,6 +23,7 @@ pub mod error;
 #[allow(dead_code)]
 const MIN_RELAY_FEE_RATE_SAT_VB: u64 = 1;
```
arminsabouri, line 25: This seems like a clippy error that is easy to fix (just box the large error variant) and probably shouldn't be ignored

bin/test-suite/src/suite/consensus/frost/test_batch_pegins.rs
```diff
@@ -173,16 +217,24 @@ pub async fn batch_pegins(
...
         let bitcoin_block_height = conf_block_info.height;
-        let meta = PeginMeta {
-            version: 0,
-            outpoint: bitcoin::OutPoint::new(pegin_tx.compute_txid(), vout as u32),
-            address: eth_account,
-            aggregate_publickey: agg_pk,
-            tx: pegin_tx.clone(),
-            merkle_proof: pmt,
-            block_headers: headers.clone(),
+        let version = if pegins.len() < NUM_PEGINS_PER_VERSION as usize { 0 } else { 1 };
```
arminsabouri, line 220: This seems odd. Why not just take the version as a param from the caller?

crates/consensus/authority/src/excecution_utils.rs
```diff
@@ -353,6 +353,7 @@ pub(crate) mod authority_execution_utils {
             db,
             bitcoind_factory.clone(),
             bitcoin_network,
+            database_provider.clone(),
```
arminsabouri, line 356: Do we need to pass the entire factory? Usually it's fine to instantiate the provider and use that.

crates/primitives/src/botanix/peg_contract.rs
```diff
@@ -132,7 +149,7 @@ impl PeginData {
 
 /// Pegin metadata structure
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct PeginMeta {
+pub struct PeginMetaV0 {
```
arminsabouri, line 152: We can implicitly assume PeginMeta is version 0. No need for this name change

bin/btc-server/src/coordinator/mod.rs
```diff
@@ -23,6 +23,7 @@ pub mod error;
...
 
+#[allow(clippy::result_large_err)]
```
arminsabouri, line 26: Is this related to pegin v1 proof? If not let's fix this in a separate commit or pr.

arminsabouri:
I'd like to see two test cases
One for a e2e peginv1 another using a peginvX. Where X is just some future version we might use one day

bin/btc-server/src/coordinator/mod.rs
SatwikPrabhu: These are clippy errors that keep showing up. Needed to do these changes to make CI green. Either we merge with a failed clippy CI job or we fix them on the fly. I would have preferred to have PR that specifically fix these, but I do not see a pattern of when some of these clippy errors show up.

crates/consensus/authority/src/excecution_utils.rs
SatwikPrabhu: Opened a ticket to track this --> https://github.com/botanix-labs/botanix/issues/936

bin/test-suite/src/suite/consensus/frost/test_batch_pegins.rs
SatwikPrabhu: Will take care of this as part of https://github.com/botanix-labs/botanix/issues/932

bin/btc-server/src/coordinator/mod.rs
arminsabouri: Have you taken a look at https://rust-lang.github.io/rust-clippy/master/index.html#result_large_err. There is a even a suggestion for a fix.

arminsabouri: I would have also prefered another PR to fix clippy issues before this PR got merged. For the future.

scottmillner: apologies that's on me since i merged in
i'll create a pr and implement the fix you noted
https://rust-lang.github.io/rust-clippy/master/index.html#result_large_err

thanks for sharing that!

SatwikPrabhu: > I would have also prefered another PR to fix clippy issues before this PR got merged. For the future.

Apologies for that. Will indeed note this for the future

SatwikPrabhu: @scottmillner Let me know if you dont find time for this. I can do this as well :)

scottmillner: addressed here: https://github.com/botanix-labs/Macbeth/pull/621

PR #613, opened 2025-03-09
fix(test-suite): skip comet sync check during int tests

scottmillner:


scottmillner:
will reopen if needed
this should be fixed in: https://github.com/botanix-labs/botanix/issues/929

PR #614, opened 2025-03-13
chore(abci): error out on txs being recived via `check_tx`

arminsabouri:
Txs should not be entering and persisting in the cometBFT mempool. On our testnet we "firewall" off the cometRPC but others may not. We expect txs to be entering the reth mempool via RPC, a user can submit a txs via comet's RPC however it will never get included in a block. Validating a tx that will never be in a block set a bad precedence and practice. 

Edit: I left the check tx code commented out in case its useful later. Happy to just remove it as well.

### DIFF
--- a/bin/reth/src/commands/poa/mod.rs
+++ b/bin/reth/src/commands/poa/mod.rs
@@ -799,13 +799,11 @@ impl<Ext: clap::Args + fmt::Debug> PoaNodeCommand<Ext> {
         }
 
         // NOTE: the node will block here until DKG has completed
-        let eth_tx_validator = validator.validator.clone();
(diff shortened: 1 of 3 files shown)


scottmillner (dismissed):

lamafab (dismissed):
_Briefly discussed internally_

SatwikPrabhu (dismissed):

scottmillner (dismissed):

scottmillner (dismissed):
thanks for doing this!

lamafab:
Some weird linting complaints, again...

scottmillner (approved):

lamafab (approved):

rwlockbg (approved):

PR #615, opened 2025-03-14
fix:  enable cometbft - rpc connection before syncing check

rwlockbg:
This this PR does:

1. This PR resolves the issues raised in [929](https://github.com/botanix-labs/botanix/issues/929). The current implementation allows having the frost task getting notified about abci being started (also reth rpcs) which is the prerequisite for the comet-reth connection. If the latter is not satisfied, the `is_syncing` condition in the frost task would always be failing as there is simply no other way of having reth-comet connected. On the other hand, the frost task loop needs to proceed at least one iteration before we start checking for the syncing condition in order to have the dkg started See here -> https://github.com/botanix-labs/Macbeth/blob/main/bin/reth/src/commands/poa/mod.rs#L800. For this reason I have added a condition that checks in each iteration if the abci has already been started and only when, then we would start checking for syncing on cometbft side. If syncing, the frost task would do empty cycles until sync is complete.

2. I have also fixed some linting stuff acc. to the latest Rust version we are using. We SHOULD USE 1.85 as 1.87 has issues with backwards compat atm and we get compilation errors. I have set this to the pipelines too.

3. I have fixed compilation warnings and errors acc. to v1.85 standards.

### DIFF
--- a/bin/federation-utils/src/handler.rs
+++ b/bin/federation-utils/src/handler.rs
@@ -3,7 +3,7 @@ use crate::{
     config::{get_default_config_path, Config},
     errors::WalletError,
 };
-use base64::decode;
+use base64::Engine;
 use ethers::{prelude::*, types::Transaction};
 use hex::encode as hex_encode;
 use k256::{ecdsa::SigningKey as Secp256k1SigningKey, elliptic_curve::generic_array::GenericArray};
@@ -159,7 +159,9 @@ fn deserialize_secret_key(seceret_path: &str) -> Result<String, WalletError> {
     }
     match parsed_key.priv_key.key_type.as_str() {
         "tendermint/PrivKeySecp256k1" => {
-            let priv_key_bytes = decode(&parsed_key.priv_key.value)
+            let engine = base64::engine::general_purpose::STANDARD;
+            let priv_key_bytes = engine
+                .decode(&parsed_key.priv_key.value)
                 .map_err(|e| WalletError::CustomError(format!("Base64 decode error: {:?}", e)))?;
 
             // Ensure the key is exactly 32 bytes
--- a/bin/reth/src/commands/poa/mod.rs
+++ b/bin/reth/src/commands/poa/mod.rs
@@ -724,6 +724,7 @@ impl<Ext: clap::Args + fmt::Debug> PoaNodeCommand<Ext> {
             .with_host(cometbft_rpc_host);
 
         // Build authority Consensus
+        let (abci_started_tx, abci_started_rx) = tokio::sync::oneshot::channel::<()>();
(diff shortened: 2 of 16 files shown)


bin/federation-utils/src/handler.rs
```diff
@@ -159,7 +159,9 @@ fn deserialize_secret_key(seceret_path: &str) -> Result<String, WalletError> {
     }
     match parsed_key.priv_key.key_type.as_str() {
         "tendermint/PrivKeySecp256k1" => {
-            let priv_key_bytes = decode(&parsed_key.priv_key.value)
+            let engine = base64::engine::general_purpose::STANDARD;
```
lamafab, line 162: This `base64` API deprecation is super annoying, it shows up all over the Rust ecosystem 🙃

bin/reth/src/commands/poa/mod.rs
```diff
@@ -888,6 +889,7 @@ impl<Ext: clap::Args + fmt::Debug> PoaNodeCommand<Ext> {
 
             launch_rpc.await?
         };
+        abci_started_tx.send(()).expect("abci started tx");
```
lamafab, line 892: Wouldn't it be better if this was triggered in `ABCIClientBuilder::start_server`? But that's probably just a nitpick. I would hence also ignore the error and have it dropped quietly, instead of panicking.

lamafab (approved):
LGTM 👍

bin/federation-utils/src/handler.rs
rwlockbg: yes, we need to fix them throughout. We can do another PR just for that cleanup.

bin/reth/src/commands/poa/mod.rs
rwlockbg: Is it not essential for us to have that oneshot transferred to the frost task to give it a go ? Hence why I decided for panicking. Lets keep those important things in the poa/mod.rs expect-ing, elsewhere proper error-handling.

lamafab: >  it not essential for us to have that oneshot transferred to the frost task to give it a go ?

Technically, only Fed members start the Frost Task:

```rust
        if is_fed_node {
            executor.spawn_critical(
                "Frost Task",
                Box::pin(async move {
                    frost_task.expect("frost task exists").start_task(abci_started_rx).await;
                }),
            );
        }
```

For normal nodes, the `tx` value gets sent to a channel that is never read. This is fine since the `rx` is still alive when that happens, and gets dropped later. Point is, it's only essential for Fed members. But we'd probably notice if this would become an issue some day, anyway.

scottmillner (approved):
🔥 thanks for fixing the clippy issue as well!

PR #619, opened 2025-03-18
feat: Added rpc endpoint for getting agg pub key

rwlockbg:


lamafab (dismissed):

arminsabouri (commented):
Not sure what diff I should be looking at. But the gateway address endpoint provides the agg pk in its response. Is there something missing from that response?

rwlockbg:
> Not sure what diff I should be looking at. But the gateway address endpoint provides the agg pk in its response. Is there something missing from that response?

The idea of this endpoint is to get direct access to the agg pub key for monitoring. One way to ascertain all nodes have done the dkg successfully is to call that endpoint and fetch the pub key and compare it. The gateway address one needs an eth address and tweaking and is unpractical for just monitoring. I discussed it with @scottmillner yesterday.

scottmillner:
@rwlockbg can you move the `make fmt` changes to a separate commit so we have another commit just with endpoint changes, thanks!

rwlockbg:
> @rwlockbg can you move the `make fmt` changes to a separate commit so we have another commit just with endpoint changes, thanks!

Not sure how to revert them  back to state 0 :(

scottmillner:
> > @rwlockbg can you move the `make fmt` changes to a separate commit so we have another commit just with endpoint changes, thanks!
> 
> Not sure how to revert them back to state 0 :(

i'd just do `git reset --soft HEAD~1`
then remove the fmt changes and commit the rpc changes then git push --force
then run make fmt again and commit those

rwlockbg:
> > > @rwlockbg can you move the `make fmt` changes to a separate commit so we have another commit just with endpoint changes, thanks!
> > 
> > 
> > Not sure how to revert them back to state 0 :(
> 
> i'd just do `git reset --soft HEAD~1` then remove the fmt changes and commit the rpc changes then git push --force then run make fmt again and commit those

Dont we want to merge it all so all pipelines flash green ? If only the rpc changes, pipelines will be red!

scottmillner:
> > > > @rwlockbg can you move the `make fmt` changes to a separate commit so we have another commit just with endpoint changes, thanks!
> > > 
> > > 
> > > Not sure how to revert them back to state 0 :(
> > 
> > 
> > i'd just do `git reset --soft HEAD~1` then remove the fmt changes and commit the rpc changes then git push --force then run make fmt again and commit those
> 
> Dont we want to merge it all so all pipelines flash green ? If only the rpc changes, pipelines will be red!

we still will have the format changes, just two commits instead of one
- first commit: rpc changes
- second commit: make fmt

i reset and did two commits

arminsabouri:
> One way to ascertain all nodes have done the dkg successfully is to call that endpoint and fetch the pub key and compare it. The gateway address one needs an eth address and tweaking and is unpractical for just monitoring.

Seems unneccecary to create an entire new endpoint for this. 
Either
1. Pass a dummy eth address
2. Pass a flag to the existing endpoint indicating that we just want the agg pk

scottmillner:
> Seems unneccecary to create an entire new endpoint for this.


if we decide not to add a new endpoint,
i like option one (passing dummy eth address) bc semantically having a flag to only get app_pub_key from the gatewayaddress endpoint feels a bit confusing to me

scottmillner (approved):

lamafab (approved):

PR #620, opened 2025-03-19
Refactor(Pegins): Improve pegin versioning ergonomics

SatwikPrabhu:
This PR improves the ergonomics of the Pegin version design as well as decouple `PeginMeta` from `PeginData` as the latter is not likely to have future versions. Currently, we have an enum managing the two versions. This PR changes to use a trait to this management so that we need to do minimal code changes in the future when we have new versions. 

This is the fix for https://github.com/botanix-labs/botanix/issues/931

### DIFF
--- a/bin/test-suite/src/suite/consensus/frost/test_batch_pegins.rs
+++ b/bin/test-suite/src/suite/consensus/frost/test_batch_pegins.rs
@@ -10,7 +10,7 @@ use ethers::{
 use reth_primitives::{
     botanix::{
         mint_validation::MINT_TOPIC,
-        peg_contract::{PeginData, PeginMeta, PeginMetaV0, PeginMetaV1},
+        peg_contract::{PeginData, PeginMeta, PeginMetaV1, PeginMetaTrait},
         utils::AmountExt,
     },
     B256,
@@ -43,29 +43,35 @@ fn create_pegin_meta(
     pmt: PartialMerkleTree,
     headers: Vec<bitcoin::block::Header>,
     reference_hash: Option<B256>,
-) -> PeginMeta {
+) -> Box<dyn PeginMetaTrait> {
     match version {
-        PEGIN_META_VERSION_V0 => PeginMeta::V0(PeginMetaV0 {
-            version: PEGIN_META_VERSION_V0,
-            outpoint: bitcoin::OutPoint::new(pegin_tx.compute_txid(), vout as u32),
-            address: eth_account,
-            aggregate_publickey: agg_pk,
-            tx: pegin_tx.clone(),
-            merkle_proof: pmt,
-            block_headers: headers,
-        }),
-        PEGIN_META_VERSION_V1 => PeginMeta::V1(PeginMetaV1 {
-            inner: PeginMetaV0 {
-                version: PEGIN_META_VERSION_V1,
+        PEGIN_META_VERSION_V0 => {
+            let meta = PeginMeta {
+                version: PEGIN_META_VERSION_V0,
                 outpoint: bitcoin::OutPoint::new(pegin_tx.compute_txid(), vout as u32),
                 address: eth_account,
-                aggregate_publickey: agg_pk,
+                aggregate_public_key: agg_pk,
                 tx: pegin_tx.clone(),
                 merkle_proof: pmt,
                 block_headers: headers,
-            },
-            ref_block_hash: reference_hash.unwrap(),
-        }),
+            };
+            Box::new(meta) as Box<dyn PeginMetaTrait>
+        },
+        PEGIN_META_VERSION_V1 => {
+            let meta = PeginMetaV1 {
+                inner: PeginMeta {
+                    version: PEGIN_META_VERSION_V1,
+                    outpoint: bitcoin::OutPoint::new(pegin_tx.compute_txid(), vout as u32),
+                    address: eth_account,
+                    aggregate_public_key: agg_pk,
+                    tx: pegin_tx.clone(),
+                    merkle_proof: pmt,
+                    block_headers: headers,
+                },
+                ref_block_hash: reference_hash.unwrap(),
+            };
+            Box::new(meta) as Box<dyn PeginMetaTrait>
+        },
         _ => panic!("unsupported version"),
     }
 }
@@ -223,7 +229,7 @@ pub async fn batch_pegins(
             None
         };
 
-        let meta = create_pegin_meta(
+        let meta = vec![create_pegin_meta(
             version,
             &pegin_tx,
             vout,
@@ -232,18 +238,17 @@ pub async fn batch_pegins(
             pmt,
             headers.clone(),
             reference_hash,
-        );
+        )];
 
         // validate the pegin data first offchain before submitting
         let pegin_data = PeginData {
             account: Address::from_slice(eth_address.as_bytes()),
             amount,
             bitcoin_block_height: bitcoin_block_height as u32,
-            meta: vec![meta.clone()],
         };
 
-        pegin_data.validate(&checkpoint, &agg_pk).expect("pegin data should be valid!");
-        pegins.push(pegin_data);
+        meta[0].validate(&checkpoint, &agg_pk, pegin_data.clone()).expect("pegin data should be valid!");
+        pegins.push((pegin_data, meta));
     }
 
     // mint all the pegins
@@ -258,13 +263,13 @@ pub async fn batch_pegins(
     let mut nonce = provider.nonce().await;
     for (_, pegin) in pegins.iter().enumerate() {
         // There is only one pegin that needs to be serialized
-        let serialized_pegin_meta = pegin.meta[0].serialize().unwrap();
+        let serialized_pegin_meta = pegin.1[0].serialize().unwrap();
         let metadata = ethers::core::types::Bytes::from(serialized_pegin_meta.clone());
         let tx_hash = provider
             .non_confirmed_mint(
-                ethers::core::types::Address::from_slice(pegin.account.as_slice()),
-                pegin.amount,
-                pegin.bitcoin_block_height,
+                ethers::core::types::Address::from_slice(pegin.0.account.as_ref()),
+                pegin.0.amount,
+                pegin.0.bitcoin_block_height,
                 metadata,
                 refund_address,
                 nonce,
@@ -295,7 +300,7 @@ pub async fn batch_pegins(
     tokio::time::sleep(Duration::from_secs(5)).await;
     // Ensure each eth address has a non zero balance
     for (_, pegin) in pegins.iter().enumerate() {
-        let eth_address_balance = provider.get_botanix_balance(pegin.account).await;
+        let eth_address_balance = provider.get_botanix_balance(pegin.0.account).await;
         assert!(!eth_address_balance.expect("get balance").is_zero());
     }
 
@@ -315,7 +320,7 @@ pub async fn batch_pegins(
         let utxo = utxos.iter().find(|utxo| {
             bitcoin::Txid::from_slice(utxo.outpoint.as_ref().unwrap().txid.as_slice())
                 .expect("valid txid") ==
-                pegin.meta[0].tx().compute_txid()
+                pegin.1[0].as_any().downcast_ref::<PeginMeta>().unwrap().tx.compute_txid()
         });
         assert!(utxo.is_some());
     }
--- a/crates/ethereum/evm/src/execute.rs
+++ b/crates/ethereum/evm/src/execute.rs
@@ -34,7 +34,7 @@ use reth_primitives::{
         mint_validation::{
             try_parse_burn_event, try_parse_mint_event, MintContractError, MINT_CONTRACT_ADDRESS,
         },
-        peg_contract::{PeginData, PegoutWithId},
+        peg_contract::{EssentialPeginData, PeginData, PeginMeta, PeginMetaV1, PegoutWithId},
     },
     header_ext::HeaderExt,
     Address, BlockNumber, BlockWithSenders, EthereumHardfork, Header, Receipt, Request, TxHash,
@@ -58,6 +58,9 @@ use revm_primitives::{
 use alloc::{boxed::Box, sync::Arc, vec, vec::Vec};
 #[cfg(feature = "std")]
 use std::sync::Arc;
+
+/// Result type for mint contract checks, containing validated pegin and pegout data
+pub type MintContractCheckResult = (Vec<(PeginData, Vec<EssentialPeginData>)>, Vec<PegoutWithId>);
 
 /// Provides executors to execute regular ethereum blocks
 #[derive(Debug, Clone)]
@@ -163,7 +166,7 @@ struct EthExecuteOutput {
     requests: Vec<Request>,
     gas_used: u64,
     total_block_fees: u128,
-    pegins: Vec<PeginData>,
+    pegins: Vec<(PeginData, Vec<EssentialPeginData>)>,
     pegouts: Vec<PegoutWithId>,
 }
 
@@ -237,8 +240,8 @@ where
         let mut receipts = Vec::with_capacity(block.body.len());
 
         for (sender, transaction) in block.transactions_with_sender() {
-            // The sum of the transaction’s gas limit, Tg, and the gas utilized in this block prior,
-            // must be no greater than the block’s gasLimit.
+            // The sum of the transaction's gas limit, Tg, and the gas utilized in this block prior,
+            // must be no greater than the block's gasLimit.
             let block_available_gas = block.header.gas_limit - cumulative_gas_used;
             if transaction.gas_limit() > block_available_gas {
                 return Err(BlockValidationError::TransactionGasLimitMoreThanAvailableBlockGas {
@@ -429,7 +432,7 @@ where
         botanix_consensus_pkg: &BotanixConsensusPackage,
         tx_hash: TxHash,
         provider: ProviderFactory<RethDB>,
-    ) -> Result<(Vec<PeginData>, Vec<PegoutWithId>), MintContractError> {
+    ) -> Result<MintContractCheckResult, MintContractError> {
         let consensus_pkg = botanix_consensus_pkg;
         let btc_network = consensus_pkg.btc_network;
 
@@ -437,9 +440,9 @@ where
         let mut pegins = vec![];
         let mut pegouts = vec![];
         for log in result.logs() {
-            let pegin_data = match try_parse_mint_event(log)? {
-                None => continue,
-                Some(p) => p,
+            let (pegin_data, pegin_meta) = match try_parse_mint_event(log)? {
+                (Some(p),Some(m)) => (p, m),
+                _ => continue,
             };
 
             // Get the reference block hash from the pegin metadata.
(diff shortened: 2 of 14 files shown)


bin/test-suite/src/suite/consensus/frost/test_batch_pegins.rs
```diff
@@ -258,13 +263,13 @@ pub async fn batch_pegins(
...
         // There is only one pegin that needs to be serialized
-        let serialized_pegin_meta = pegin.meta[0].serialize().unwrap();
+        let serialized_pegin_meta = pegin.1[0].serialize().unwrap();
         let metadata = ethers::core::types::Bytes::from(serialized_pegin_meta.clone());
         let tx_hash = provider
             .non_confirmed_mint(
-                ethers::core::types::Address::from_slice(pegin.account.as_slice()),
-                pegin.amount,
-                pegin.bitcoin_block_height,
+                ethers::core::types::Address::from_slice(pegin.0.account.as_slice()),
```
rwlockbg: Cannot you implement as_ref() here ?

crates/ethereum/evm/src/execute.rs
```diff
@@ -495,30 +495,35 @@ where
                 });
             }
             let aggregate_public_key = consensus_pkg.aggregate_public_key;
-            match pegin_data.validate(&bitcoin_checkpoint, &aggregate_public_key) {
-                Ok(aggregate_value) => {
-                    if pegin_data.amount >= aggregate_value {
+            let mut aggregate_value = ethers::types::U256::from_str_radix("0", 10).expect("valid amount");
```
arminsabouri, line 504: Is this a safe expect?

crates/primitives/src/botanix/peg_contract.rs
```diff
@@ -37,119 +38,166 @@ pub struct PeginData {
...
+
+// Implement PartialEq for Box<dyn PeginMetaTrait>
+impl PartialEq for Box<dyn PeginMetaTrait> {
+    fn eq(&self, other: &Self) -> bool {
+        // Objects are equal if they have the same type tag and the equals method returns true
+        self.type_tag() == other.type_tag() && self.equals(other.as_ref())
     }
 }
 
+// Implement Eq for Box<dyn PeginMetaTrait>
```
arminsabouri: Comment redundant

crates/primitives/src/botanix/peg_contract.rs
```diff
@@ -37,119 +38,166 @@ pub struct PeginData {
...
+    fn type_tag(&self) -> u32;
+}
+
+impl Clone for Box<dyn PeginMetaTrait> {
+    fn clone(&self) -> Self {
+        self.clone_box()
+    }
+}
+
+// Implement PartialEq for Box<dyn PeginMetaTrait>
```
arminsabouri: Comment redundant

crates/primitives/src/botanix/peg_contract.rs
```diff
@@ -37,119 +38,164 @@ pub struct PeginData {
...
+        Ok(output_value)
+    }
+
+    /// Get the txout for the pegin
+    fn txout(&self) -> &TxOut;
+
+    /// Get the version of the pegin metadata
+    fn version(&self) -> u32;
+
+    /// Get the merkle proof of the pegin tx
```
scottmillner: just a general comment: prefer not to have getters for fields that are pub since you can just access them directly and makes the trait overly burdensome to implement

SatwikPrabhu:
> Why not just take the version as a param from the caller? in test_batch_pegins

I wanted to do this change as part of https://github.com/botanix-labs/botanix/issues/932 as it is not really in line with the changes in this PR.

> Do we need to pass the entire factory? Usually it's fine to instantiate the provider and use that in execution_utils

Opened a new ticket (https://github.com/botanix-labs/botanix/issues/936) for this again as I want to keep this ticket just for changing from enum to trait rather than a bunch of different changes. Does that make sense?

crates/primitives/src/botanix/peg_contract.rs
SatwikPrabhu: Valid point. But if we do get away with these getters then, we would have to check every time to see whether we have a `PeginMeta` or `PeginMetaV1` to access the fields. 

Maybe an intermediary solution is to have a single getter function to get all the fields instead of individual getters for each field. This may reduce the burden of implementing the trait.

crates/ethereum/evm/src/execute.rs
arminsabouri: I realized this code may have been migrated from somewhere else. But I think we should still answer that question

crates/ethereum/evm/src/execute.rs
```diff
@@ -59,6 +59,9 @@ use alloc::{boxed::Box, sync::Arc, vec, vec::Vec};
 #[cfg(feature = "std")]
 use std::sync::Arc;
 
+/// Result type for mint contract checks, containing validated pegin and pegout data
+pub type MintContractCheckResult = (Vec<(PeginData, Vec<Box<dyn PeginMetaTrait>>)>, Vec<PegoutWithId>);
```
arminsabouri: Is this a different PeginData for every peginMeta? I dont think it is. I would strucutre this more like
(PeginData, Vec<...>)

crates/primitives/src/botanix/mint_validation.rs
```diff
@@ -197,23 +200,56 @@ pub fn try_parse_mint_event(
...
+            })?;
+            let (proof, proof_size) = match version {
+                PEGIN_META_VERSION_V0 => {
+                    let (meta, size) = PeginMeta::deserialize(&meta_bytes[offset..]).map_err(|e| {
+                        let err = ParseMintEventError::InvalidPeginData {
+                            error: e,
+                            revert_address: destination,
+                            revert_amount: amount,
+                        };
+                        error!("Failed to parse pegin meta: {:?}", err);
```
arminsabouri:
```suggestion
                        error!("Failed to parse pegin v0 meta: {:?}", err);
```
And same change for the one below

crates/primitives/src/botanix/mint_validation.rs
```diff
@@ -197,23 +200,56 @@ pub fn try_parse_mint_event(
...
+                        err
+                    })?;
+                    let meta = Box::new(meta) as Box<dyn PeginMetaTrait>;
+                    (meta, size)
+                }
+                _ => return Err(ParseMintEventError::InvalidPeginData {
+                    error: PeginDataError::Invalid("Invalid meta version. Only accepting 0 or 1"),
+                    revert_address: destination,
+                    revert_amount: amount,
```
arminsabouri: Is this correct? if we we're to do an update in the future wouldn't this change make it a hardfork?

crates/primitives/src/botanix/peg_contract.rs
```diff
@@ -37,119 +38,164 @@ pub struct PeginData {
...
-    /// Pegin metadata
-    pub meta: Vec<PeginMeta>,
 }
 
-impl PeginData {
-    /// Validates multiple pegin proofs against the current bitcoin block hash
-    /// Returns the aggregate value of all the pegin amounts
-    pub fn validate(
+/// Deserialize a meta version from the beginning of a byte slice.
+pub fn deserialize_meta_version(mut bytes: &[u8]) -> Result<u32, PeginDataError> {
```
arminsabouri: Nit: this shouldnt be mut bytes as it moves the buffer cursor. Now the deserialize method needs to be aware that another method was called to read the first few bytes. I would just pull u32 and try to validate against it. Then feed a mut bytes into deserialize

crates/primitives/src/botanix/peg_contract.rs
```diff
@@ -37,119 +38,164 @@ pub struct PeginData {
...
+pub trait PeginMetaTrait: Debug + 'static + Send + Sync {
+    /// Serialize a pegin meta
+    fn serialize(&self) -> Result<Vec<u8>, PeginDataError>;
+
+    /// Deserialize a pegin meta
+    fn deserialize(bytes: &[u8]) -> Result<(Self, usize), PeginDataError> where Self: Sized;
+
+    /// Validates a pegin proof against the current bitcoin block hash
+    /// Returns the value of the pegin
+    fn validate(
```
arminsabouri, line 100: Can you please provide some context on the changes here in validate? I expected a few changes given that this code is now using `pegin_data` instead of `self.` but it looks like the whole method was re-formatted making it hard to review the actual changes

crates/primitives/src/botanix/peg_contract.rs
```diff
@@ -37,119 +38,164 @@ pub struct PeginData {
...
         &self,
         bitcoin_commitment: &(bitcoin::block::Header, u32),
         aggregate_pk: &secp256k1::PublicKey,
+        pegin_data: PeginData,
     ) -> Result<U256, PeginDataError> {
-        // the aggregate value from all the pegin proofs
-        let mut aggregate_value = U256::from_str_radix("0", 10).expect("valid amount");
```
arminsabouri, line 53: Where did this go? Did it get moved outside this method?

crates/primitives/src/botanix/peg_contract.rs
```diff
@@ -224,36 +268,92 @@ impl PeginMetaV0 {
...
+    fn outpoint(&self) -> bitcoin::OutPoint {
+        self.outpoint
+    }
+
+    fn tx(&self) -> bitcoin::Transaction {
+        self.tx.clone()
+    }
+
+    fn ref_block_hash(&self) -> Option<B256> {
+        None
```
arminsabouri: Hmmm this is my biggest concern so far. I dont think peginv0 needs to know about the ref_block_hash. I understand why this needs to be included here but if this PR is concerned with ergonomics I do think we want to handle this.

crates/primitives/src/botanix/peg_contract.rs
```diff
@@ -224,36 +268,92 @@ impl PeginMetaV0 {
...
+    fn address(&self) -> Address {
+        self.address
+    }
+
+    fn aggregate_public_key(&self) -> secp256k1::PublicKey {
+        self.aggregate_publickey
+    }
+
+    fn block_headers(&self) -> Vec<Header> {
+        self.block_headers.clone()
```
arminsabouri: This could be a Heavy clone. Can we just return a ref?

crates/primitives/src/botanix/peg_contract.rs
scottmillner: im not following...using a getter or accessing a pub field is the same thing

scottmillner:
forgot to mention: i would like https://github.com/botanix-labs/botanix/issues/932 to go in before this pr
and this pr can refactor the tests if needed

Why?
- i don't want us to get in the habit of merging in critical code without having an int test for it
- i merged in the previous pr without the test bc i wanted it included in the audit but would not have done so otherwise

i can work on the test pr while you work on this one

arminsabouri (changes requested):
Thanks for sticking on this. Already a lot clearner. A few things.
Its interesting that you went with a trait approach. One thing that still appears to me as tech debt that may hurt us later is the fact that ref_block_hash is optional and needs to be included in the peginv0 metadata. I do want to question that a bit and ask if that is the most rusty idomatic option. Alternatively we could have distinct types in their own sub/mods. 
```
-v0
----- mod.rs
----- error.rs
-v1
...
```

Since pegin metas will always be update as a softfork you can always encapsulate them (as you partially already have). However here we would have a independent validate, ser, deser, etc... methods for each distinct type. And pegin metas are not trait objects but primitive types.

v1's validate method may look like 

```rust 
struct v1PeginMeta {
v1: PeginMeta,
ref_block_hash: B256, // NOT OPTIONAL 
}
....
fn validate(self, pegin_ctx: PeginData): Result<...> {
       self.validate_ref_block_hash(self.ref_block_hash)?
       self.v1.validate(pegin_ctx)?;
}
```

Of course there is an issue of how do we know which type to instiate. We can do as you already have and read the first few bytes and identify the pegin proof version. But such a method should not live on either v0 or v1 types. Should be independent.

There may be some issues with distinct types through out the code. For example in [notifications](https://github.com/botanix-labs/Macbeth/pull/620/files#diff-66ce7935586595273665b30b2f7be150c5e63a8c042ad05f5bd6256933f7e038R80) where on each block commit all the different tasks are notified of incoming pegins. I would question do the tasks actually need the entire PeginMeta Proof. This isn't even a cheap struct to clone and throw around. I would posit that most of the tasks mainly need the outpoint and pevout (txout). So perhaps after validation we need a DTO and that can be provided to the tasks and eventually btc_server. For the exact fields we need I would look at what btc-server needs when being notified about pegins. 

Note that with this approach cannot easily mix pegin meta version. For instance you cannot have a peginv0 followed by a pegin v1. This is complicated to handle anyways and encourges the user to do weird and costly things. Its also error-prone behavior. Should Consensus validate aginst mixed-version. I'm leaning towards yes. Eitherway we should see a test for this.

crates/ethereum/evm/src/execute.rs
SatwikPrabhu: I think that's how I have structured it currently, if I'm not mistaken. A PeginData with a Vec of PeginMeta

crates/primitives/src/botanix/peg_contract.rs
SatwikPrabhu: I am passing a clone of the meta bytes to this function. So, it should not affect things if we are moving the cursor right?

crates/primitives/src/botanix/peg_contract.rs
SatwikPrabhu: Apologies. Should have added more comments. The function `validate` used to validate the vector of `PeginMeta` and return the aggregate amount of each proof. Also, it used to error out if even one of the proofs wasnt valid. Now, this function is part of `PeginMetaTrait` and deals with only one `PeginMeta`, the aggregation of the amount happens at the caller and the caller now passes `PeginData` since it is not directly accessible in `self.` anymore.

crates/primitives/src/botanix/peg_contract.rs
SatwikPrabhu: Yep. This got moved to the caller as explained in https://github.com/botanix-labs/Macbeth/pull/620#discussion_r2005921599

crates/ethereum/evm/src/execute.rs
arminsabouri: yes mis-read sorry.

crates/primitives/src/botanix/peg_contract.rs
SatwikPrabhu: I get that. I'll try to think of a way to do this

crates/primitives/src/botanix/mint_validation.rs
SatwikPrabhu: I don't think I fully understand what you mean here.

crates/primitives/src/botanix/peg_contract.rs
SatwikPrabhu: If we do not have getters, then we'll have to determine whether a trait object is v0 or v1, every time. I found doing that to be a bit tedious. Am I missing something here?

In any case, it looks like I'll need to move away from the common trait to address the new comments by @0xBEEFCAF3

lamafab:
So what's wrong with using enums for this, exactly? I would not use a Trait for that and have `dyn Trait` be passed around.

Something like:

```rust
enum PeginData {
    V0(PeginDataV1),
    V1(PeginDataV2),
}

impl PeginData {
    /// Get the txout for the pegin
    fn txout(&self) -> &TxOut;

    /// Get the version of the pegin metadata
    fn version(&self) -> u32;

    /// Get the merkle proof of the pegin tx
    fn merkle_proof(&self) -> PartialMerkleTree;

    /// Get the outpoint of the pegin tx
    fn outpoint(&self) -> bitcoin::OutPoint;

    // ...
}

struct PeginDataV0;

struct PeginDataV1;

// Version specific methods.
impl PeginDataV1 {
    /// Get the ref block hash of the pegin tx
    fn ref_block_hash(&self) -> BlockHash;
}
```

As @SatwikPrabhu mentioned, having getters is justified for this:

> But if we do get away with these getters then, we would have to check every time to see whether we have a `PeginMeta` or `PeginMetaV1` to access the fields.

For version specific fields, such as `ref_block_hash`, you SHOULD be required to match:

```rust
let version = pegin.version();
do_something_with_version(version);

let txout = pegin.txout();
do_something_with_txout(txout);

match pegin {
    PeginData::V0(pegin) => {
        // ...
    }
    PeginData::V1(pegin) => {
        // ...
        let checkpoint = pegin.ref_block_hash();
        do_something_with_checkpoint(checkpoint);
    }
}

```

Separating `PeginData` from `PeginMeta` like that is a mistake, imo:

```rust
pub type MintContractCheckResult = (Vec<(PeginData, Vec<Box<dyn PeginMetaTrait>>)>, Vec<PegoutWithId>);
```

You quickly start having issues like those (pseudo-code):

```rust
match (data, meta) {
    (PeginData::V0(data), PeginMeta::V0(meta)) => {
        // ...
    }
    (PeginData::V0(data), PeginMeta::VX(meta)) => {
        panic!("pegin data v0 does not support meta vX");
    }
    (PeginData::V1(data), PeginMeta::VX(meta)) => {
        // ...
    }
}
```

If you catch my point...

crates/primitives/src/botanix/peg_contract.rs
scottmillner: copy that

scottmillner:
this ticket seems to have grown in scope: @0xBEEFCAF3 are we approaching this as a tech debt item?
or perhaps i misunderstood the original scope of the first pr (to support a new pegin version)

it seems we're in agreement that the trait approach is not ideal

i'm not a fan of the tuple below for the reason mentioned by @lamafab: 
```
pub type MintContractCheckResult = (Vec<(PeginData, Vec<Box<dyn PeginMetaTrait>>)>, Vec<PegoutWithId>);
```

SatwikPrabhu:
> i'm not a fan of this: pub type MintContractCheckResult = (Vec<(PeginData, Vec<Box<dyn PeginMetaTrait>>)>, Vec<PegoutWithId>); but perhaps this can be done differently

I generally get clippy errors if I pass this big type directly

scottmillner:
> > i'm not a fan of this: pub type MintContractCheckResult = (Vec<(PeginData, Vec<Box>)>, Vec); but perhaps this can be done differently
> 
> I generally get clippy errors if I pass this big type directly

I'm more concerned with the flow @lamafab showed when we are returning a tuple. IMO this makes the code harder to reason about and seems unnecessary. I appreciate you working on this major refactor but wondering if this was the original intent of the first pr?  This feels a bit like scope creep and maybe we should revisit what we want to achieve.

SatwikPrabhu:
Tried to address the comments in https://github.com/botanix-labs/Macbeth/pull/620#pullrequestreview-2699472436. @0xBEEFCAF3 Let me know if this version is something along the lines of what you had in mind. There still maybe some minor changes that still needs to be done.

SatwikPrabhu:
The unit and int tests are passing for me locally but failing here

scottmillner:
im putting this in draft for now as we decide which changes we want to do

scottmillner:
closing for now
will reopen when we circle back

PR #625, opened 2025-03-21
Fix(Pegins): Ensure all proofs in a pegin have the same version and reference block hash

SatwikPrabhu:
Fix for https://github.com/botanix-labs/botanix/issues/935

### DIFF
--- a/bin/test-suite/src/suite/consensus/invalid_transactions/test_invalid_pegin.rs
+++ b/bin/test-suite/src/suite/consensus/invalid_transactions/test_invalid_pegin.rs
@@ -1,6 +1,12 @@
 use std::{str::FromStr, time::Duration};
 
-use bitcoin::{hashes::Hash, merkle_tree::PartialMerkleTree, Amount, Txid};
+use bitcoin::{
+    blockdata::block::Header,
+    hashes::Hash,
+    merkle_tree::PartialMerkleTree,
+    Amount,
+    Txid
+};
 use bitcoincore_rpc::RpcApi;
 use ethers::{prelude::Provider, providers::Http};
 use reth_primitives::{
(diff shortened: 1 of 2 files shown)


lamafab (dismissed):
I'd do that stuff directly in `try_parse_mint_event` tbh, but otherwise looks good.

SatwikPrabhu:
> I'd do that stuff directly in try_parse_mint_event tbh, but otherwise looks good.

Valid point. But maybe its simpler to keep `try_parse_mint_event` to just read/parse/extract stuff from logs and exclude other kinds of logic here?

lamafab:
This kind of leads back to this: https://github.com/botanix-labs/Macbeth/pull/620#issuecomment-2741883235

Imo the validity should be ensured the moment the `PeginData`/`PeginMeta` types are created.

crates/ethereum/evm/src/execute.rs
```diff
@@ -479,7 +479,39 @@ where
...
+                (meta.version(), meta.ref_block_hash())
+            } else {
+                return Err(MintContractError::InvalidPeginData {
+                    error: "No proofs found in pegin data".to_string(),
+                    revert_address: pegin_data.account,
+                    revert_amount: pegin_data.amount,
+                })
+            };
+
+            for meta in pegin_data.clone().meta {
```
arminsabouri: Can you use into iter()?

rwlockbg (dismissed):

scottmillner (dismissed):

scottmillner:
please test these cases (mixed versions, mixed ref block hashes)

create invalid pegins for both cases in
`test_invalid_pegin`

SatwikPrabhu:
Ive added these cases to the test. Its passing locally for me. Waiting for them to pass here in CI

SatwikPrabhu:
All tests have passed. Looks like the int test fix works 🔥 
Thanks @rwlockbg and @scottmillner

arminsabouri (approved):
Wow! The tests have come along way. Nice :+1: 
A question that came up: Are we assuming that the block height in PeginData is corresponding to the blockhash. I feel like there is an implicit assumption but not an explicit check. Please clarify

bin/test-suite/src/suite/consensus/invalid_transactions/test_invalid_pegin.rs
```diff
@@ -135,198 +322,79 @@ pub async fn invalid_pegin(
...
+    for (invalid_pegin_meta, description) in invalid_pegin_metas {
+        it_info_print!("Invalid pegin meta: {}", description);
+        let mut serialized_pegin_meta = Vec::new();
+        for meta in invalid_pegin_meta {
+            let serialized = meta.serialize().unwrap();
+            serialized_pegin_meta.extend_from_slice(&serialized);
+        }
+        it_info_print!("Serialized pegin meta: ", hex::encode(serialized_pegin_meta.clone()));
+
+        // send the pegin transactions to all fed members
```
lamafab, line 348: Nitpick: this comment should probably be changed - I assume it's outdated. It's repeated later on again when the actual `mint()` call happens.

lamafab (approved):

scottmillner (approved):

SatwikPrabhu:
> A question that came up: Are we assuming that the block height in PeginData is corresponding to the blockhash. I feel like there is an implicit assumption but not an explicit check. Please clarify

I just assumed for testing sake since we dont have the mechanism of getting info from sidecar as such. I think for testing purposes, this is sufficient.

bin/test-suite/src/suite/consensus/invalid_transactions/test_invalid_pegin.rs
SatwikPrabhu: Nice catch. Yeah, I probably missed/duplicated some comments while refactoring. Probably we need to take care of this at some point.

arminsabouri:
> > A question that came up: Are we assuming that the block height in PeginData is corresponding to the blockhash. I feel like there is an implicit assumption but not an explicit check. Please clarify
> 
> I just assumed for testing sake since we dont have the mechanism of getting info from sidecar as such. I think for testing purposes, this is sufficient.

The information only flows from sidecar to reth. Not the either way around. PeginData is already parsed and available in the context of minting-contract checks. My question is outside the scope of this PR - not really related to testing. Either way I think this is fine

SatwikPrabhu:
> > > A question that came up: Are we assuming that the block height in PeginData is corresponding to the blockhash. I feel like there is an implicit assumption but not an explicit check. Please clarify
> > 
> > 
> > I just assumed for testing sake since we dont have the mechanism of getting info from sidecar as such. I think for testing purposes, this is sufficient.
> 
> The information only flows from sidecar to reth. Not the either way around. PeginData is already parsed and available in the context of minting-contract checks. My question is outside the scope of this PR - not really related to testing. Either way I think this is fine

Apologies. Thought that the initial comment was regarding the testing assumptions. 
Regarding the question --> I dont see an implicit assumption here. Isnt it the case that the checkpoint corresponding to the ref block hash would be ideally very close (not necessarily corresponding) to the block height in pegin data. I'm not sure what the check would be here.

PR #626, opened 2025-03-21
Fix(Pegins): Fix version related validate checks and unit tests

SatwikPrabhu:
Fix for https://github.com/botanix-labs/botanix/issues/940

### DIFF
--- a/crates/primitives/src/botanix/peg_contract.rs
+++ b/crates/primitives/src/botanix/peg_contract.rs
@@ -53,23 +53,15 @@ impl PeginData {
         let mut aggregate_value = U256::from_str_radix("0", 10).expect("valid amount");
         let commit_hash = bitcoin_commitment.0.block_hash();
         for pegin in &self.meta {
+            if ![PEGIN_META_VERSION_V0, PEGIN_META_VERSION_V1].contains(&pegin.version()) {
+                return Err(PeginDataError::Invalid(
+                    "invalid meta version: only accepting version 0 or 1",
+                ));
+            };
+            
             let pegin = match pegin {
-                PeginMeta::V0(meta) => {
-                    if meta.version != PEGIN_META_VERSION_V0 {
-                        return Err(PeginDataError::Invalid(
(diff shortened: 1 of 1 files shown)


crates/primitives/src/botanix/peg_contract.rs
```diff
@@ -53,23 +53,15 @@ impl PeginData {
         let mut aggregate_value = U256::from_str_radix("0", 10).expect("valid amount");
         let commit_hash = bitcoin_commitment.0.block_hash();
         for pegin in &self.meta {
+            if ![PEGIN_META_VERSION_V0, PEGIN_META_VERSION_V1].contains(&pegin.version()) {
```
lamafab, line 56: This kind of changes the core validation logic; for example, this does allow the usage of `PeginMeta::V0` with `pegin.version() == 1`. Is that on purpose?

lamafab (changes requested):

crates/primitives/src/botanix/peg_contract.rs
SatwikPrabhu: The thing is we will already check for this in `botanix_mint_contract_checks` as part of https://github.com/botanix-labs/Macbeth/pull/625. So I felt that it would be a redundant check here. Can add this here if you feel strongly about it.

lamafab: Yeah, specifically in `PeginMeta::deserialize`. Btw I think it's worth enforcing that version check directly in `PeginMetaV0::deserialize` and `PeginMetaV1::deserialize`, respectively. The the former should check that `version == 0` and the later `version == 1` - or error if otherwise. But that's an issue for another day.

lamafab (approved):

rwlockbg (approved):

scottmillner (approved):

scottmillner:
Please don't merge until after https://github.com/botanix-labs/Macbeth/pull/625 is merged
since 625 enforces all proofs have the same version and ref block hash if v1

PR #627, opened 2025-03-22
ci(int-tests): each int test is separate step in workflow

scottmillner:
For the int test worflow, I'm adding each test as a separate step to make it easier to identify which test failed instead of searching through logs.

After this I'd like to explore running each test as a separate job in parallel but would need to randomly assign port values to avoid conflicts

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -20,7 +20,7 @@ concurrency:
 jobs:
   e2e:
     name: test / e2e
-    runs-on: 
+    runs-on:
       group: macbeth
     env:
       RUST_BACKTRACE: 1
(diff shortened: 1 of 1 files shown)


SatwikPrabhu:
I guess we dont need the file `test_runner.sh` anymore and can be removed?

scottmillner:
> I guess we dont need the file `test_runner.sh` anymore and can be removed?

let's keep it for now
it's a nice to have when testing locally: you can run all the tests at once or individually

lamafab (approved):
Nice!

Btw I noticed that `bin/btc-server/src/test_utils/integration.rs` is an undeclared/unused module entirely.

SatwikPrabhu (approved):

rwlockbg (approved):

PR #629, opened 2025-03-24
fix(audit): set bounds and add checks on pegin deserialize to prevent…

rwlockbg:
… DoS attacks

### DIFF
--- a/bin/test-suite/src/utils.rs
+++ b/bin/test-suite/src/utils.rs
@@ -156,8 +156,8 @@ pub struct BlockChainInfoRes {
 }
 
 pub fn get_checkpoint_block_hash(bitcoind: &impl RpcApi) -> Result<Vec<u8>, Error> {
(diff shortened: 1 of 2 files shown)


arminsabouri (approved):
cACK. Thanks for hopping on this. This was ticket somewhere at somepoint. Any thoughts on recourse? i.e what do we do with customers funds if they go past this consensus limit. I suppose with v1 this is much less of a concern but this edge case could still be there

lamafab (approved):

lamafab:
@0xBEEFCAF3 
> i.e what do we do with customers funds if they go past this consensus limit. I suppose with v1 this is much less of a concern but this edge case could still be there

We can probably assume that edge cases like those are not a realistic scenario that will affect normal users. Maybe in the future we will have a refund mechanism for (invalid) pegins or so, for whatever reason, but as of now this is not something worth considering.

PR #630, opened 2025-03-24
fix: gossiping bug fixed by always polling new pool imports

rwlockbg:
When we push to `pool_imports` the import future is never polled thus we only actually import transaction once the `TransactionsManager` is polled next time which might take a while

Solution is to poll `pool_imports` after all actions which might result in new futures being pushed to it


lamafab (dismissed):

SatwikPrabhu (dismissed):

scottmillner:
it's working like a charm!! 🔥

scottmillner (dismissed):

scottmillner (approved):

scottmillner:
unit test pass locally, merging

PR #632, opened 2025-03-25
Fix(Pegins): Avoid passing entire factory to eth executor

SatwikPrabhu:
Fix for https://github.com/botanix-labs/botanix/issues/936


arminsabouri (approved):
utACK

rwlockbg (approved):

PR #634, opened 2025-03-27
feat: Added pending pegouts wallet state sync

rwlockbg:


rwlockbg:
PS> I still need to add an integration test!

arminsabouri (changes requested):
What problem are we solving here? Its hard to evalute this solution when the problem space is multi dimentional and hard to reason about. It would make sense to solve -- what ever we are solving --  inversly where we first create the problematic condition then solve it. Not the other way.

There are couple problems here:
* No validation on the pending pegouts list. A peer can give you an arbituary pegout list
* How do you pick a peer list? What if one peer has some pending pegouts list and another subset's list is different? 
* Replays still seem like an BIG issue here. You can settle a list pegouts. Some time passes and you are out of sync. You attempt to sync with a peer who is also out of sync or malicously inserts previously settled pegouts. 

I'm not convinced a sync approach could work well here. But again, I would say the first step is to clearly outline what the problem we are solving is? This can take the form of a test or even a document with steps.  

Nack

scottmillner:
> What problem are we solving here? Its hard to evalute this solution when the problem space is multi dimentional and hard to reason about. It would make sense to solve -- what ever we are solving -- inversly where we first create the problematic condition then solve it. Not the other way.
> 
> There are couple problems here:
> 
> * No validation on the pending pegouts list. A peer can give you an arbituary pegout list
> * How do you pick a peer list? What if one peer has some pending pegouts list and another subset's list is different?
> * Replays still seem like an BIG issue here. You can settle a list pegouts. Some time passes and you are out of sync. You attempt to sync with a peer who is also out of sync or malicously inserts previously settled pegouts.
> 
> I'm not convinced a sync approach could work well here. But again, I would say the first step is to clearly outline what the problem we are solving is? This can take the form of a test or even a document with steps.
> 
> Nack

Thanks for asking and let me know what you think.

**The Problem**
Psbts are being blind signed: signers are accepting anything given to them by the coordinator. Signers are not validating psbts (calling `validate_psbt`).

**The Solution**
The ultimate solution is what @0xBEEFCAF3 has detailed out for async pegouts.

**Intermediate Solution Until Async Pegouts**

- Signers validate psbts which includes enforcing the inputs in the psbt match their list of pending pegouts
- Signers are no longer blind signing and this should help prevent signing for the same output multiple times (not as robust as async pegouts for sure)
- Signers will need to sync on pending pegouts which presents more questions that @0xBEEFCAF3 mentioned above

**Naive Approach**
- Sync with all peers (15 for mainnet) asking for their pending pegouts list
- Create a superset of all the lists: this absolutely opens us up to the attack of replaying outputs
- To prevent replaying outputs: enable the `conflicting_input` feature (separate pr)
- This feature requires signers to validate psbts and track them once broadcasted (they will store `tracked_txs`)
- When signers are validating a psbt, they will enforce a conflicting input exists if they notice one of the outputs has already been honored in a `tracked_tx`. This helps mitigate replaying an output from the pending pegouts superset list
- Assumption: as long as one of the signers is aware of the tracked tx that contains an output being replayed in the pending pegouts list, the replay attack is prevented by a conflicting input

arminsabouri:
> * Sync with all peers (15 for mainnet) asking for their pending pegouts list

This seems to have **strong** liveness requirments, no? 

>To prevent replaying outputs: enable the conflicting_input feature (separate pr)

You are still vulnerable to replays once the pegouts is sufficiently deep and no longer tracked. I don't know how to get around this with out 
* Finalized outputs
* White list of spks (Liquid style PAK proof)
*Or somekinda proof system that the federation never spent to this output

>Assumption: as long as one of the signers is aware of the tracked tx that contains an output being replayed in the pending pegouts list, the replay attack is prevented by a conflicting input

Why one signer? I feel like im missing something. Replays are a problem even after txs are tracked.

scottmillner:
> > * Sync with all peers (15 for mainnet) asking for their pending pegouts list
> 
> This seems to have **strong** liveness requirments, no?
> 
> > To prevent replaying outputs: enable the conflicting_input feature (separate pr)
> 
> You are still vulnerable to replays once the pegouts is sufficiently deep and no longer tracked. I don't know how to get around this with out
> 
> * Finalized outputs
> * White list of spks (Liquid style PAK proof)
>   *Or somekinda proof system that the federation never spent to this output
> 
> > Assumption: as long as one of the signers is aware of the tracked tx that contains an output being replayed in the pending pegouts list, the replay attack is prevented by a conflicting input
> 
> Why one signer? I feel like im missing something. Replays are a problem even after txs are tracked.

Thanks for the feedback:
- yeah wasn't thinking about once a tracked_tx is deeply confirmed and then removed from the db 

`Why one signer? I feel like im missing something. Replays are a problem even after txs are tracked.`
- i was just thinking while it's still tracked but you're right, this does nothing to prevent replay attack once it's deeply confirmed

As an intermediate fix to prevent replays:
- once a tracked_tx is deeply confirmed, we store it in the db as our `Finalized outputs` list that we cross-check before signing
- this list will diverge as well so you could still replay an output if the minimum required number of signers does not have the deeply confirmed tx in its finalized outputs list.

bin/reth/src/commands/poa/mod.rs
```diff
@@ -766,6 +767,17 @@ impl<Ext: clap::Args + fmt::Debug> PoaNodeCommand<Ext> {
             }),
         );
 
+        if let Some(wallet_sync) = wallet_sync {
```
scottmillner, line 805: we're only syncing on startup but will need to add more logic:
- when signers are validating a psbt and it doesn't have the exact same pegouts stored in pending pegouts, they will reject signing
- so they need to then sync on pending pegouts

crates/consensus/authority/src/wallet_state_sync.rs
```diff
@@ -0,0 +1,198 @@
...
+                                        pending_pegouts_decompressed,
+                                    ) else {
+                                        error!(target: "consensus::authority::sync_wallet_state", "Failed to deserialize pending pegouts");
+                                        continue;
+                                    };
+                                    pending_pegouts_decompressed.pending_pegouts
+                                }
+                            };
+
+                            // Report to btc server to reset the wallet state
```
scottmillner: As is, this syncs on the first peer that responds
We need to sample more than one peer before syncing.

We could start with sampling all peers (but as armin mentioned, this is a strong liveness requirement).  Create a superset from all the pending pegouts list and sync once we have that.

But we have to have the conflicting input feature enable first to prevent replays (only prevents replays before their deeply confirmed).

But as i write this, this clearly (to me) doesn't seem like the correct path we should be going down. And sorry for any wasted work. We need some type of finalized outputs list to crosscheck before signing.

See my other comment about this

scottmillner:
as @0xBEEFCAF3 said: 

we need some kind of list of finalized pegouts list:

- which would be the l2 txid of the pegout
- and signers would check that list before signing a psbt
- without this, there is no protection against replays once the tx is deeply confirmed
- and if we have this list, we don't need wallet state syncing functionality

@0xBEEFCAF3 copied this from another comment above but your thoughts:

As an intermediate fix to prevent replays:

- Once a tracked_tx is deeply confirmed, we store it in the db as our `finalized pegouts` list that we cross-check before signing

- This list will diverge as well so you could still replay an output if the minimum required number of signers does not have the deeply confirmed tx in its finalized outputs list. The greater the number of minimum signers, the less likely this becomes.

bin/btc-server/proto/btc_server.proto
```diff
@@ -96,9 +102,7 @@ message ResetAllUtxosRequest {
 }
 
 message ResetWalletStateRequest {
-    repeated Utxo utxos = 1;
-    repeated TrackedTx tracked_txs = 2;
-    repeated PendingPegout pending_pegouts = 3;
+    repeated bytes finalzied_pegout_ids = 1;
```
arminsabouri: Typo

bin/btc-server/src/bin/main.rs
```diff
@@ -683,39 +699,22 @@ where
...
-        //     .collect::<Result<Vec<PegoutRequest>, _>>()
-        //     .map_err(|e| internal!("Failed to convert pending pegout: {}", e))?;
-        // let pending_pegouts_refs: Vec<&PegoutRequest> = pending_pegouts.iter().collect();
-        // self.db.reset_pending_pegouts(&pending_pegouts_refs).to_status()?;
+        self.validate_jwt(&req)?;
+        let req = req.into_inner();
+        info!("Received reset wallet state request");
+
+        // handle finalized pegout ids
+        let finalzied_pegout_ids = req
```
arminsabouri: Typo

bin/btc-server/src/database/mod.rs
```diff
@@ -739,6 +747,83 @@ impl Db {
         Ok(self.pending_pegouts.clear()?)
     }
 
+    // --------------NEW---------------------
```
arminsabouri: Unnecessary comment

bin/btc-server/src/database/mod.rs
```diff
@@ -688,7 +696,7 @@ impl Db {
             .get(id.as_bytes())?
             .map(|b| ciborium::de::from_reader(b.as_ref()).expect("corrupt db: pending pegout")))
     }
-
+    // -----------------------------------
```
arminsabouri: Do we need this comment?

bin/btc-server/src/database/mod.rs
```diff
@@ -739,6 +747,83 @@ impl Db {
...
+                        database_tx.insert(req.as_bytes().to_vec(), &bytes[..])?;
+                    }
+                }
+                Ok::<(), ConflictableTransactionError>(())
+            })
+            .map_err(|e: TransactionError<_>| Error::Transaction(e.to_string()))?;
+        Ok(())
+    }
+
+    // ----------------------------
```
arminsabouri: Same thing about this comment

crates/net/network/src/frost/messages.rs
```diff
@@ -72,27 +72,33 @@ impl SignRequest {
 /// A structured wallet state message
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct WalletStateRequest {
+    /// uuid of the state sync request
```
arminsabouri: Wallet state sync request?

bin/btc-server/proto/btc_server.proto
arminsabouri: This list will grow fast. We should have some plan in place for sending chunks of pegout ids.

bin/btc-server/src/database/mod.rs
```diff
@@ -739,6 +747,83 @@ impl Db {
...
+    /// Removes finalized pegout ids from the database.
+    pub fn remove_finalized_pegout_ids(
+        &self,
+        finalized_pegout_ids: &[PegoutId],
+    ) -> Result<(), Error> {
+        for pegout_id in finalized_pegout_ids.iter() {
+            self.finalized_pegout_ids.remove(&pegout_id.as_bytes()[..])?;
+        }
+        Ok(())
+    }
```
arminsabouri: Is there ever a case where we need to remove a specific pegoutid not the whole thing?

bin/btc-server/src/database/mod.rs
```diff
@@ -739,6 +747,83 @@ impl Db {
...
+    /// Store a list of finalized pegout ids
+    pub fn store_finalized_pegout_ids(
+        &self,
+        finalized_pegout_ids: &[&PegoutId],
+    ) -> Result<(), Error> {
+        match finalized_pegout_ids.len() {
+            0 => Ok(()),
+            1 => self
+                .store_finalized_pegout_ids(&[finalized_pegout_ids.first().expect("to have tx")]),
+            _ => self.store_finalized_pegout_ids_atomically(finalized_pegout_ids),
```
arminsabouri: When we consider what we are doing with this dataset we really need to do queries (preventing pegout replays and fast insertions (adding new finalized pegouts). 
 In an ideal setting we have some hash based index that can give us fast constant time look ups and insertions. Like a Redis or Memcache, but storing this large list is also not ideal. With a SQL based database and custom indexes  the best one can achive is O(log n) which is also pretty good considering the whole thing is not loaded in mem (i think).

With Sled the problem is we have to do a full tree scan  (O(n)) and we have to load every pegout id in memory. And we expect this list to grow fast. So maybe we fix this here or another PR. But definetly before this gets released (@scottmillner )

So one way to do this would be to keep a merkelized representation of the pegout ids in sled at all times. Both inssertions and lookups should be O(log n)

arminsabouri: The nice thing about merkelizing the pegoutids is when inserting the finalized output list in the OP_RETURN we don't need to any additional computation. Just read the root :)

bin/reth/src/commands/poa/mod.rs
```diff
@@ -768,6 +769,17 @@ impl<Ext: clap::Args + fmt::Debug> PoaNodeCommand<Ext> {
...
+            executor.spawn_critical(
+                "Wallet Sync",
+                Box::pin(async move {
+                    if let Err(e) = wallet_sync.sync_wallet_state().await {
+                        error!(target: "reth::cli", "Wallet Sync Error: {:?}", e);
+                    }
+                }),
+            );
```
arminsabouri, line 813: Just to be clear finalized output list syncing doesnt need to a long-running task. The BTC server will signal that it will not sign (r1 frost signing, nonce gen) a PSBT bc it doesnt have the latest list of finalized pegouts. "Wallet sync" would kick off from there

arminsabouri (changes requested):
This has the right structure. My biggest comments were around the the wallet state task it self and wether we need it. And how the pegoutids should be represented going into sled. Some typos here and there as well.

We def should get some more eyes on this @lamafab @SatwikPrabhu

bin/btc-server/src/database/mod.rs
scottmillner: Agreed and thanks for the detailed comments.

Yep, we can't ship this as is for mainnet.
Let's merkelize pegout ids and load that in memory.  @rwlockbg im fine including this in this pr or breaking it out into a new one: whatever you prefer.

We've talked about a SQL db before and definitely worth exploring (and something we should do imo).
I've created a ticket for this: 
https://github.com/botanix-labs/botanix/issues/976

bin/reth/src/commands/poa/mod.rs
scottmillner: enforcing the pending pegouts list will be removed in https://github.com/botanix-labs/botanix/issues/956

crates/consensus/authority/src/wallet_state_sync.rs
scottmillner: we've decided on having a finalized outputs(pegoutids) list to cross-check during signing

bin/reth/src/commands/poa/mod.rs
scottmillner: Here's what I was thinking and let me know your thoughts:
- finalized pegouts list will diverge when signers don't participate in a signing round (go offline for example)
- edge case is where the min num of required signers have missed a tracked tx so those outputs can be replayed

This seems very unlikey (is it possible?) but was thinking syncing finalized pegouts periodically would prevent this?

crates/consensus/authority/src/wallet_state_sync.rs
```diff
@@ -0,0 +1,200 @@
...
+                    Some(peer_message_context) => {
+                        info!(target: "consensus::authority::sync_wallet_state", "Received wallet state from peer {:?}", peer_message_context.peer_id);
+
+                        // Note: we ignore empty messages bc they are peer requests for wallet state
+                        // which are handled by the frost task or are malicious/faulty requests
+                        // that would cause the btc-server to wipe its state
+                        if let PeerMessageResponse::WalletState(wallet_state) =
+                            peer_message_context.message
+                        {
+                            // process the utxos
```
scottmillner: utxos > `finalized pegouts` (or just `wallet state`)

crates/consensus/authority/src/wallet_state_sync.rs
```diff
@@ -0,0 +1,200 @@
...
+            .send_command(FrostCommand::GetPeerMessagesStream(peer_messages_tx))?;
+        let mut peer_messages_rx = peer_messages_rx.await.expect("peer messages rx to be open");
+
+        let data_parser = self.data_parser.clone();
+        let mut canon_events = self.storage.client.subscribe_to_canonical_state();
+
+        self.task_executor.clone().spawn(async move {
+            // try getting the wallet state from the peers we requested it from
+            loop {
+                match peer_messages_rx.recv().await {
```
scottmillner, line 271: This updates the wallet state for a single peer at a time:
- This favors the most lively peer and increases the likelihood of having a list with missing finalized pegouts
- Would be nice to have some form of peer sampling to create an aggregate list of finalized pegouts and send that to btc-server
- But also don't want to have too strong of a constraint here: not sure what a reasonable subset of peers makes sense? The min required signers? But if we do this every epoch would could just end up constantly syncing as we wait for the subset to respond and never actually sync?

@0xBEEFCAF3 @rwlockbg

bin/btc-server/proto/btc_server.proto
rwlockbg: fixed. Ty

bin/btc-server/src/bin/main.rs
rwlockbg: fixed, ty

bin/btc-server/src/database/mod.rs
rwlockbg: sorry, was a temp thing, removed

bin/btc-server/src/database/mod.rs
rwlockbg: removed

bin/btc-server/proto/btc_server.proto
rwlockbg: Might make sense, how long could it grow - chunks or maybe capping in terms of returning the latest N items  ? @0xBEEFCAF3 @scottmillner @lamafab

bin/btc-server/src/database/mod.rs
rwlockbg: Nope, but I have still added it for completeness. If it is an issue, happy to remove it

bin/btc-server/src/database/mod.rs
rwlockbg: @0xBEEFCAF3 @scottmillner @lamafab Or maybe this is the time we switch to rocksdb finally ?

crates/consensus/authority/src/wallet_state_sync.rs
rwlockbg: fixed, ty

bin/btc-server/src/database/mod.rs
scottmillner: we should consider this but not right now imo

we can start with doing:

```
So one way to do this would be to keep a merkelized representation of the pegout ids in sled at all times. Both inssertions and lookups should be O(log n)
```

IMO, this is acceptable in the short term. Once we have tested this feature and it works, we can start exploring rocksdb or something else.

bin/btc-server/proto/btc_server.proto
scottmillner: im fine with chunks or latest N items

crates/net/network/src/frost/messages.rs
rwlockbg: fixed, ty.

crates/consensus/authority/src/wallet_state_sync.rs
rwlockbg: Please check the latest implementation @scottmillner @0xBEEFCAF3

bin/reth/src/commands/poa/mod.rs
rwlockbg: I also agree with @scottmillner - most of our designs in the consensus auth crate are designed around long running tasks. Please have a look at how I have re-designed it now @0xBEEFCAF3  and @scottmillner

bin/btc-server/proto/btc_server.proto
arminsabouri: It grows indefinetly.

bin/btc-server/src/database/mod.rs
arminsabouri: It may make more sense to remove the whole tree as not to load the entire set of pegoutids into memory.

bin/btc-server/proto/btc_server.proto
rwlockbg: @0xBEEFCAF3 @scottmillner  How do you imagine the chunks being sent over the wire via grpc . There is 2 ways. 1. some sort of indexation/pagination is needed to tell us how many chunks there are and how many remain to be sent so that the caller can do multiple requests with pagination till all of them are sent. 
2. The other option is to introduce unidirectional grpc streaming which is the most elegant way to get the job done. A stream will be open until all of the chunks are sent in batches over the wire to the receiver. 

Opinions ?

scottmillner: I think retry logic is easier with pagination since the peer would know the last chunk it received and could continue sequentially but quite a bit of overhead with this approach.

**So I like option 2.** As the data set grows, streaming seems to be more efficient and will reduce the overall time to sync. But need to think of recovery strategy if the stream closes before finished.

bin/reth/src/commands/poa/mod.rs
scottmillner: I like the re-design but would like @0xBEEFCAF3 thoughts on what I posted above about the edge case where min number of signers have missed a tracked tx.

Also how does the btc server know `it doesnt have the latest list of finalized pegouts.`?

crates/consensus/authority/src/wallet_state_sync.rs
scottmillner: Thanks for the refactor! Logic LGTM.  Enforcing min num of signers per session makes sense but as that number grows it might not be practical.  Good number to start with and then can lower if needed. This is a bit trial and error.

scottmillner:
@rwlockbg do you think it makes sense to do the `unidirectional grpc streaming` in another pr?

So after this pr, do the `merkelized outputs in memory for querying` pr and then the streaming pr.

So for this pr, we can write some tests and make sure it works and merge then do the others.

bin/btc-server/src/database/mod.rs
```diff
@@ -739,6 +747,79 @@ impl Db {
...
+        pegout_ids_requests: &[&PegoutId],
+    ) -> Result<(), Error> {
+        self.finalized_pegout_ids
+            .transaction(|database_tx| {
+                for req in pegout_ids_requests.iter() {
+                    if database_tx.get(req.as_bytes())?.is_none() {
+                        let mut bytes = Vec::new();
+                        ciborium::into_writer(req, &mut bytes)
+                            .map_err(Error::CiboriumWrite)
+                            .expect("Ciborium error");
```
SatwikPrabhu, line 894: Do we need to panic here. We could just do `.map_err(Error::CiboriumWrite)?;` right

bin/btc-server/src/bin/main.rs
```diff
@@ -683,39 +699,22 @@ where
...
+        let req = req.into_inner();
+        info!("Received reset wallet state request");
+
+        // handle finalized pegout ids
+        let finalized_pegout_ids = req
+            .finalized_pegout_ids
+            .into_iter()
+            .map(|v| PegoutId::from_bytes(&v))
+            .collect::<Result<Vec<PegoutId>, _>>()
+            .map_err(|_| internal!("Failed to convert finalized pegout ids"))?;
```
SatwikPrabhu, line 785: Maybe we can capture the underlying error instead of the generic closure

bin/btc-server/src/database/mod.rs
arminsabouri: Following up on this!

SatwikPrabhu (commented):
The code looks good to me(although my understanding is not complete in this part of the codebase)
Left a couple of nits from my side. Other than that maybe it would be nice to add tests to verify the functionality of for example `store_finalized_pegout_ids_atomically`.

bin/reth/src/commands/poa/mod.rs
arminsabouri: >Also how does the btc server know it doesnt have the latest list of finalized pegouts.?

It compares the hash of the merkel root others have commited to on the tx. If that merkel root does match the one you have in DB then  you don't sign and instead sync

crates/consensus/authority/src/wallet_state_sync.rs
arminsabouri: I dont think we need a sample from many peers. Instead you can glean that information from the signers in a previous signing session.

crates/consensus/authority/src/wallet_state_sync.rs
```diff
@@ -0,0 +1,260 @@
...
+                            match current_response_cycle.as_mut() {
+                                Some((uuid, peers)) => {
+                                    if *uuid != request_uuid {
+                                        warn!(target: "consensus::authority::sync_wallet_state", "Received wallet state with different uuid, ignoring");
+                                        continue;
+                                    }
+                                    peers.insert(peer_message_context.peer_id, finalized_pegout_ids);
+
+                                    if peers.len() as u16 >= frost_config.min_signers {
+                                        // consenses the finalized pegout ids
```
arminsabouri: typo?
```suggestion
                                        // compress the finalized pegout ids
```

crates/consensus/authority/src/wallet_state_sync.rs
```diff
@@ -0,0 +1,260 @@
...
+                    // Request the wallet state from all peers for poa epoch blocks only
+                    let uuid = uuid::Uuid::new_v4();
+                    if let Err(e) = self
+                        .to_frost_manager
+                        .send_command(FrostCommand::GetWalletStateFromPeer(uuid))
+                    {
+                        error!(target: "consensus::authority::sync_wallet_state", ?e, "Failed to send get wallet state command to frost manager");
+                    }
+                    // start the current response cycle
+                    current_response_cycle.write().await.replace((uuid, HashMap::new()));
```
arminsabouri: I'm not so sure about this. This seems fragile. The information for which peers have valid wallet state comes from the previous signing sessions. Imo this should be a signal based system not a polling system.

crates/net/network/src/frost/manager.rs
```diff
@@ -390,6 +390,39 @@ impl FrostManager {
...
+                    }
+                }
+
+                // reply to caller
+                // if let Err(e) = tx.send(peer_connections) {
+                //     error!(
+                //         target: "network::frost::on_command",
+                //         "Error replying to call on GetAllConnectedPeers {:?}", e
+                //     );
+                // }
```
arminsabouri: Is this important? if not lets rm

crates/net/network/src/frost/messages.rs
```diff
@@ -494,29 +491,32 @@ impl FrostProtoMessage {
...
-                    tracked_txs,
-                    pending_pegouts,
-                ))
+                // Decode uuid
+                let uuid_len = u32::from_le_bytes(
+                    buf[..4].try_into().expect("Buffer underflow for uuid length"),
+                ) as usize;
+                buf.advance(4);
+                let uuid =
+                    String::from_utf8(buf[..uuid_len].to_vec()).expect("Invalid UTF-8 in uuid");
```
arminsabouri: Non-utf8 characters seems like something we should return an error for. Especially when we are deserializing.

arminsabouri (changes requested):
Couple new things found:  expects that shouldn't be there and some open items from last review that were not addressed.

My biggest pain points are:
* Wallet state sync task should be migrated to a different task (frost task?). When you receieve a R1 frost signing message you should get a psbt that commits to the merkle root of the finalized pegouts list. Only if your btc server's root is the same do you sign. If you see that n signers have commited then you know you need to sync. Having this be replaced by a long-running task that assumes peers have the correct list seems fragile and prone to bugs. Im not too convinced on this design choice.

* We had talked about optimizing look up and inserts via a merkle tree. I still think there is merit in this approach. Esp. considering that btc server will need a hash ready every epoch (10 ish blocks). Re-computing the root every couple minutes seems like something that can be optimized at this stage

arminsabouri:
Couple notes from the call we had regarding this PR in its current form:
* We have to validate each finalized pegout in the super set. Meaning they need to exist on chain and need to be infact deeply confirmed.
* We really should consider a merkle root comparison before this entire list is serliazed and sent over the wire. I know we compress the data but I'm not sure how much of an affect compression even has on a pegout id which is seemingly random bytes. Typically compression works well on structured binary output.

As a side note we should take a note of the growth rate of this list and take that into serious consideration:
If we do an average of a 100 pegouts. this list will only grow ~ 9.6 MB a year. It grown linearly so that is nice. 

But what I am more worries about is the peer that starts giving you garbage data. Lets say a few peer gives you 100x more pegout ids (~1 GB) than you are expecting and some of the pegout ids are invalid. This behavior could be malicous or just a result of bugs that are hard to clean up in a decentralized setting. Do we load the whole thing in mem and start validating? Epochs are short so this could be a DOS vector. How does Bitcoin solve this with UTXO set checks. Its kinda the same problem.

crates/consensus/authority/src/wallet_state_sync.rs
scottmillner:
```
When you receieve a R1 frost signing message you should get a psbt that commits to the merkle root of the finalized pegouts list. Only if your btc server's root is the same do you sign. If you see that n signers have commited then you know you need to sync. 
```

Won't this mean we're syncing on the coordinator? Coordinator creates a psbt and adds a commitment to pegouts merkle root. Signers will need to sync on that list or they'll never sign.  If coordinator drops and misses a finalized pegout, that will never be included in the merkle root which signers sync against.  

If we sync across peers, this reduces the likelihood of a finalized pegout getting missed.  I do agree the polling system might end up being too fragile but also committing to a merkle root loses it's potency if we don't rotate the coordinator.

crates/net/network/src/frost/messages.rs
rwlockbg: There wont be any non utf-8 chars. uuid produces a fully compatible utf-8 string.

bin/btc-server/src/bin/main.rs
rwlockbg: @SatwikPrabhu The PegoutId::from_bytes returns : Result<PegoutId, ()> There is no Error as you see here.

bin/btc-server/src/database/mod.rs
rwlockbg: This is inside a closure and we cannot ?. It is also the way we have done that in all prev atomical txs.

crates/consensus/authority/src/wallet_state_sync.rs
rwlockbg: fixed, thx

crates/net/network/src/frost/manager.rs
rwlockbg: remvoed

crates/net/network/src/frost/messages.rs
rwlockbg: Also, in all these ser/des methods we expect or unwrap. We might need a ticket to refactor all of this, but this is a separate concern.

crates/consensus/authority/src/wallet_state_sync.rs
arminsabouri: >Won't this mean we're syncing on the coordinator?

No, you are stil syncing across peers. But there is two added advantages
* You know who you need to sync with 
* You have an event which tells you when to sync

>If coordinator drops and misses a finalized pegout, that will never be included in the merkle root which signers sync against.

This is percisely why the cooridnator should be rotated in a round robin fashion. If you never rotate the coordinator and the coordinator is out of sync then you risk stalling signing all together.

crates/net/network/src/frost/messages.rs
arminsabouri: >There wont be any non utf-8 chars. uuid produces a fully compatible utf-8 string.

Does this mean that if a peer provides me an non uuid as an id that it would cause a panic? 

>>Also, in all these ser/des methods we expect or unwrap. We might need a ticket to refactor all of this, but this is a separate concern.

Fair enough. If the answer to the above is yes i do think it should be addressed.

arminsabouri:
I think this branch needs a rebase with main. Some commits are duplicated.

crates/net/network/src/frost/manager.rs
```diff
@@ -390,6 +390,31 @@ impl FrostManager {
...
+                let peer_ids: Vec<_> = self.authorities.keys().cloned().collect();
+                // filter all peers with active connections
+                let connected_peers =
+                    peer_ids.iter().filter_map(|peer_id| self.retrieve_peer_data(peer_id));
+
+                for peer in connected_peers {
+                    match peer.peer_commands_tx.send(FrostPeerCommand::PeerMessage(
+                        PeerMessageResponse::WalletState(WalletStateResponse {
+                            uuid: uuid.to_string(),
+                            finalized_pegout_ids: vec![],
```
lamafab, line 443: Is it expected that this is empty? I assume this comment is relevant from `wallet_state_sync.rs`:

```rust
// Note: we ignore empty messages bc they are peer requests for wallet state
// which are handled by the frost task or are malicious/faulty requests
// that would cause the btc-server to wipe its state
```

As noted in another comment:

```rust
// TODO: create separate messages for asking for wallet state and sending
// wallet state
```

crates/consensus/authority/src/frost_task.rs
```diff
@@ -500,7 +446,7 @@ where
                 let peer_id = message_context.peer_id;
                 let frost_identifier = message_context.frost_identifier;
                 match peer_message {
-                    PeerMessageResponse::WalletState(mut response) => {
+                    PeerMessageResponse::WalletState(response) => {
```
lamafab, line 467: Note that there's a `TODO` right after this:

```rust
// TODO: create separate messages for asking for wallet state and sending
// wallet state
```

lamafab: Additionally, I see a potentially dangerous assumption (`.expect("...")`) when retrieving the handle:

```rust
let all_peers_handle = self
	.dkg_state_machine
	.get_all_peers_handle()
	.await
	.expect("expect all peers handle to exist");
let peer_handle =
	all_peers_handle.get(&peer_id).expect("peer handle to exist");
```

crates/consensus/authority/src/frost_task.rs
```diff
@@ -191,94 +171,60 @@ where
...
-        // now compress the prost message
-        let prost_serialized_compressed = self.compressor.compress(&prost_serialized).await.map_err(|e| {
-            error!(target: "consensus::authority::utxo_syncer::get_utxo_set", "Got prost error {:?}", e);
-            UtxoSetSyncSerializationError::DataParser(e)
-        })?;
-        Ok(prost_serialized_compressed)
-    }
-
-    async fn get_serialized_compressed_tracked_txs(
+    async fn get_serialized_compressed_finalized_pegout_ids(
```
lamafab: This naming convention is somewhat misleading; it implies that it will return the finalized pegout Ids, but it just returns `Ok(())`. Maybe replace `get_*` with `forward_*` or `send_*`.

Additionally, I'd pass on the `mpsc::UnboundedSender<_>` direclty instead of the `PeerData`, so it's clear from just the interface what gets sent where.

crates/consensus/authority/src/wallet_state_sync.rs
```diff
@@ -0,0 +1,428 @@
...
+                    Some(peer_message_context) => {
+                        info!(target: "consensus::authority::sync_wallet_state", "Received wallet state from peer {:?}", peer_message_context.peer_id);
+                        // Note: we ignore empty messages bc they are peer requests for wallet state
+                        // which are handled by the frost task or are malicious/faulty requests
+                        // that would cause the btc-server to wipe its state
+                        if let PeerMessageResponse::WalletState(wallet_state) =
+                            peer_message_context.message
+                        {
+                            // try parsing the uuid
+                            let request_uuid = match Uuid::parse_str(&wallet_state.uuid) {
```
lamafab, line 281: Generally, could you expand on how state sync validation works when receiving those chunks? The `Uuid` is accepted as-is, which can be spoofed by the sender and could be used to feed you malicious or invalid data.

lamafab (dismissed):
So I took some time to study all this and generally it looks good, but I have not been involved with this issue in-depth and I'd appreciate a **PR walk-through** with everyone else so we can discuss the important properties that this must meet, specifically worst-case scenarios such as missing pegout Ids, non-pruned pegout Ids and maliciously inserted pegout Ids by peers.

bin/btc-server/src/database/mod.rs
arminsabouri: Following up again

rwlockbg: I have removed the method for removing a single pegout id @0xBEEFCAF3

crates/consensus/authority/src/wallet_state_sync.rs
rwlockbg: There is no state sync validation for now - we can add it towards the end.

rwlockbg: The validation we will add is that whatever the peers send you in terms of finalized pegout ids, you have got to check if they exist onchain (which is enough) before you build the superset.

crates/consensus/authority/src/frost_task.rs
rwlockbg: The todo is an old one, will remove it. All our messaging happens by sending an empty struct inside the request which the responder should populate and send back to us.

crates/consensus/authority/src/frost_task.rs
rwlockbg: Reanmed to send_*. I need to send the entire peer_data  as both the channel and the peer_id are being queried in the method.

crates/consensus/authority/src/frost_task.rs
rwlockbg: Otherwise, I have removed the expect and handled the case safely, thanks.

crates/net/network/src/frost/manager.rs
rwlockbg: The todo is an old one, will remove it. All our messaging happens by sending an empty struct inside the request which the responder should populate and send back to us.

crates/net/network/src/frost/messages.rs
rwlockbg: @0xBEEFCAF3 I have resolved all unwraps (this was an audit finding) and now all errors are being handled properly. Please check and resolve.

crates/consensus/authority/src/wallet_state_sync.rs
```diff
@@ -0,0 +1,428 @@
...
+                        error!(target: "consensus::authority::sync_wallet_state", ?e, "Failed to send get wallet state command to frost manager");
+                    }
+                    // (re-)start the current response cycle
+                    current_response_cycle.write().await.replace(uuid);
+                }
+                CanonStateNotification::Reorg { old: _old, new: _new } => {
+                    warn!(target: "consensus::authority::snapshot_manager::run", "reorg detected, this should not happen");
+                    continue;
+                }
+            }
```
SatwikPrabhu, line 477: I have a feeling that there is a race condition here. A `CanonStateNotification::Commit` sets `current_response_cycle` here and starts a background process tied to this block's `uuid`. Lets say that a `CanonStateNotification::Reorg` is triggered soon after this. The current code just logs a warning and continues without invalidating the `current_response_cycle` or signal the background task to terminate or ignore results associated with the uuid which is now stale I think. The background task will thus continue and validates wallet state peer response messages because the `request_uuid` would still match the `uuid` in `current_response_cycle`. Wouldn't this eventually lead to a divergence of the btc server's view of the finalized pegouts from the canonical state?

Apologies if the above thought process is wrong!

SatwikPrabhu:
Also, maybe we can remove a bunch of the debug prints that I introduced during debugging. We probably don't need most of them.

crates/consensus/authority/src/wallet_state_sync.rs
scottmillner: we don't have reorgs bc cometbft has instant finality but we still need the match arm
and everytime a new commit comes in, the uuid is replaced with a new one
you still think it's a problem?
and thanks for asking!

SatwikPrabhu: Thanks for clarifying! I don't think its a problem now. This race condition is practically impossible. The code looks good to me!

crates/consensus/authority/src/wallet_state_sync.rs
rwlockbg: @lamafab The validation is here: https://github.com/botanix-labs/Macbeth/pull/693

bin/btc-server/src/bin/main.rs
```diff
@@ -611,6 +610,74 @@ where
...
+
+            while let Some(chunk_result) = stream.next().await {
+                info!(
+                    "get_finalized_pegout_ids stream task: Received chunk from DB stream: {:?}",
+                    chunk_result
+                );
+                match chunk_result {
+                    Ok((pegout_ids, chunk_index, total_chunks)) => {
+                        if let Some(telemetry) = telemetry.as_ref() {
+                            telemetry.update_pending_pegouts(pegout_ids.len() as i64);
```
arminsabouri: Is this the right telemetry function to call? This function returns all the finalized pegoutids but this telemetry method updates pending pegouts?

bin/btc-server/src/bin/main.rs
```diff
@@ -611,6 +610,74 @@ where
...
+                        if tx.send(Ok(batch)).await.is_err() {
+                            error!("get_finalized_pegout_ids stream task: Client disconnected, stopping stream.");
+                            continue;
+                        }
+
+                        // add a small delay between chunks
+                        tokio::time::sleep(tokio::time::Duration::from_millis(20)).await;
+                    }
+                    Err(e) => {
+                        error!("get_finalized_pegout_ids stream task: Error from DB stream: {}. Skipping chunk.", e);
```
arminsabouri, line 679: Will a client re-request a chunk?

bin/btc-server/src/bin/main.rs
```diff
@@ -940,19 +997,58 @@ where
...
+            info!(
+                "- Output {}: value={}, script_pubkey={:?}",
+                i, tx_output.value, tx_output.script_pubkey
+            );
+        }
+        info!(
+            "[finalize_signing] NOT calling add_tracked_tx again. Associated pegout requests found in DB were: {:?}",
+            pending_pegouts.iter().map(|p| p.id).collect::<Vec<_>>()
+        );
+        info!(
```
arminsabouri, line 1014: This is a lot of logs. Are all these needed?

bin/btc-server/src/pegout_scheduler/mod.rs
```diff
@@ -407,57 +419,193 @@ impl PegoutScheduler {
...
+            store_change_res?;
+            info!(
+                "PegoutScheduler::finalize_block: Attempting flush after storing change UTXOs for tx {}",
+                txid
+            );
+            let flush_change_res = self.db.flush(); // Flush after storing change UTXOs
+            info!(
+                "PegoutScheduler::finalize_block: Result of flush after storing change UTXOs for tx {}: {:?}",
+                txid,
+                flush_change_res
```
arminsabouri: Samething here. Were these used durring debugging and forgotten to clean up?

arminsabouri (changes requested):
I want to give this PR the attention it deserves and needs. It touches **CRITICAL** code that could lead to a real loss of funds if we are not careful. However, It seems like this PR needs a rebase at least and a through clean up. There are commits from main mixed in here and what seems like debugging logs still left in. Please clean these up and re-request a review

bin/btc-server/src/bin/main.rs
rwlockbg: sorry, was the wrong one. I fixed it.

bin/btc-server/src/bin/main.rs
rwlockbg: nope, I have removed the logs

bin/btc-server/src/pegout_scheduler/mod.rs
rwlockbg: Nope, I have removed all these logs

bin/btc-server/src/bin/main.rs
rwlockbg: I have added a mechanism on the calling side that wil lprint out if there are missing chunks and if all the stream completed successfully. Nevertheless, with grpc streams it is hard to retry the re-submission of a single chunk by native gprc means and cancelling the whole stream because of one chunk error is not ideal either.  I would say there are 2 sources of erros:
- chunk is good but sending it fails: In this case I have added a retry mechanism purely on the sending side
- chunk is broken due to ceborium data error: in this case there is little we can do as the data is most likely corrupt and persisted in such a way in the db (there is in fact little we can do here). The receiving side wont get all the chunks and thus the state will be incomplete which is not going to be counted towards the minimum superset we need (which is fine again)

rwlockbg:
> PS> I still need to add an integration test!

Integration tests are added too.

scottmillner:
@rwlockbg im refactoring the happy path int test

it's not running long enough for syncing to happen once the nodes have finalized outputs:
- the test checks they all have the same list and they do but that's not bc they synced
- it's bc they all added the outputs during `finalize_block`
- so it's good the test confirms this but i need to add a bit more to check syncing as well

also adding check that outputs aren't in finalized outputs list when signing

ill have this done today/tomorrow

Done!

scottmillner (approved):
Unit tests passing locally
Int tests passing locally
I've run everything locally and smoke tests pass

PR #638, opened 2025-03-28
Fix(audit): Revert burn amount instead of tx value for invalid pegouts

SatwikPrabhu:
Fix for https://github.com/botanix-labs/botanix/issues/958


SatwikPrabhu:
We may need to either update the invalid pegout int test or add a new int test for this case

scottmillner:
> We may need to either update the invalid pegout int test or add a new int test for this case

Thanks for creating this issue!

Yes, you're right! Please add/update the int tests.

For all audit tickets, we must have int tests that confirm the bug is fixed.

SatwikPrabhu:
I've modified the existing int test for invalid pegout instead of adding a new test as this could be easily made part of it. Let me know if it looks good

crates/primitives/src/botanix/mint_validation.rs
```diff
@@ -281,13 +282,14 @@ pub fn try_parse_burn_event(
     }
 
     Ok(Some(
-        PegoutData::new(btc_amount, destination, btc_network)
+        PegoutData::new(btc_amount, amount, destination, btc_network)
```
lamafab, line 285: Can't you just retrieve the `U256` (`amount`) from the `bitcoin::Amount` (`btc_amount`)? Here you're passing on the `amount` which is only needed in case of an error.

bin/test-suite/src/suite/consensus/invalid_transactions/test_invalid_pegout.rs
```diff
@@ -40,10 +40,24 @@ pub async fn invalid_pegout(
         botanix_eth_client.get_nonce(botanix_eth_client.get_sender_address()).await.unwrap();
     it_info_print!("Nonce before pegout: ", nonce_before);
 
-    // use empty pegout data
-    let pegout_data = ethers::core::types::Bytes::new();
-    let pegout_amount = Amount::from_btc(0.5).unwrap();
+    // Lets also test that when pegouts fail, the reverted amount
+    // is the actual burned amount. For this we need the burned amount
+    // to be different from the pegout amount. 
+    let burned_amount = Amount::from_btc(0.1).unwrap();   // the "true" burned amount
```
lamafab: This is not explicitly checked for in the test(?). It only checks this (relying on the gas price, which can be manipulated):

```rust
    // subtract tx costs from initial balance
    let tx_cost = tx_receipt.gas_used.unwrap() * tx_receipt.effective_gas_price.unwrap();
    it_info_print!("Tx cost: ", tx_cost);
    sender_address_initial_balance -= tx_cost;

    assert_eq!(sender_address_initial_balance, sender_address_final_balance);
```

Maybe I'm missing something?

lamafab (changes requested):

crates/primitives/src/botanix/mint_validation.rs
lamafab: It's essentially just this, in reverse:

```rust
    let amount =
        params[0].clone().into_uint().ok_or(ParseBurnEventError::InvalidLog("pegout amount"))?;
    let btc_amount = bitcoin::Amount::from_wei_floor(amount)
        .ok_or(ParseBurnEventError::InvalidLog("invalid amount"))?;
```

But maybe `*_floor` could be a problem?

SatwikPrabhu: > But maybe *_floor could be a problem?

Yes, exactly the int tests were failing due to round inconsistencies. So, I decided just to pass the eth amount directly.

bin/test-suite/src/suite/consensus/invalid_transactions/test_invalid_pegout.rs
SatwikPrabhu: Essentially the burned amount should have been reverted back because the pegin validation fails. So not sure, what we will achieve by checking this. Or maybe I am missing your point

lamafab: Well, it says `"Lets also test that when pegouts fail, the reverted amount is the actual burned amount"` - but that's never explicitly checked.

SatwikPrabhu: Yep. Thats being verified. Apologies. I should write the comments better. Before the fix, the `pegout_amount` = 0.5 btc would have been refunded and the initial balance would have been lesser than the final balance. So we dont need to add any additional check here. I'll update the comments though accordingly to be more clear.

scottmillner: First, this is a tough test to write so thanks for giving it a shot.  And your actual fix does look correct so thanks!

But this test isn't testing what we're trying to fix.  Not sure the intent of `pegout_data` here.  This is just the pegout version (have another look at `try_parse_burn_event`).  The only check on this field is that it's a single byte but this call will fail before then bc of the invalid destination. So just keep `pegout_data` as it was. And `burned_amount` has no impact here.  You're just putting random values in the `pegout_data`. TL;DR - the changes in this test have no impact on the result of the test. It's the same as before.

The root of the issue we're fixing is:
- if `burn` is called from a contract, that `msg.value` sent from the contract is the actual burn value. Currently we are assuming `let pegout_amount = transaction.value();` is the burn value.  But that is just `msg.value` sent from the originator of the overall tx. The contract to burn call is the internal tx.  

What do I mean by `overall tx`? An EOA makes a call to a contract that makes a call to `burn`.  The EOA is the originator and currently the value it sends with tx is what would be refunded.  The value sent from the contract is the actual burn amount.

So here's my idea (can do this for any fix that prevents an internal tx attack):
- write a very simple contract that when called, then calls `burn` but with a really low value
- chatgpt can easily do this for you
- add a helper function in the test-suite that deploys this contract
- in the int test, you'll call this contract with a really high value

Above is the exact attack described in the audit

SatwikPrabhu: > TL;DR - the changes in this test have no impact on the result of the test. It's the same as before.

Apologies. Thanks for pointing this out. I realize now yes the int test change did not do anything really. 

> So here's my idea (can do this for any fix that prevents an internal tx attack):
> 
> write a very simple contract that when called, then calls burn but with a really low value
> chatgpt can easily do this for you
> add a helper function in the test-suite that deploys this contract
> in the int test, you'll call this contract with a really high value

Thanks for the suggestion. I'll write a new separate int test following this approach.

scottmillner: im writing a new test so closing this comment

bin/test-suite/mint_attack_contract_abi.json
```diff
@@ -0,0 +1,75 @@
+{
+  "abi": [
+			{
+				"inputs": [],
+				"name": "mintingContract",
```
arminsabouri: Change name

bin/test-suite/mint_attack_contract_abi.json
```diff
@@ -0,0 +1,75 @@
...
...104af565b9050919050565b60006104ee826104d1565b9050919050565b6104fe816104e3565b82525050565b600060208201905061051960008301846104f5565b92915050565b6105288161023f565b82525050565b6105378161027d565b82525050565b610546816102b3565b82525050565b600082825260208201905092915050565b82818337600083830152505050565b6000601f19601f8301169050919050565b6000610589838561054c565b935061059683858461055d565b61059f8361056c565b840190509392505050565b600060a0820190506105bf600083018961051f565b6105cc602083018861052e565b6105d9604083018761053d565b81810360608301526105ec81858761057d565b90506105fb608083018461051f565b979650505050505050565b7f4e487b7100000000000000000000000000000000000000000000000000000000600052601260045260246000fd5b60006106408261027d565b915061064b8361027d565b92508261065b5761065a610606565b5b828204905092915050565b6000604082019050818103600083015261068181868861057d565b9050818103602083015261069681848661057d565b905095945050505050565b6106aa8161046f565b81146106b557600080fd5b50565b6000815190506106c7816106a1565b92915050565b6000602082840312156106e3576106e2610215565b5b60006106f1848285016106b8565b9150509291505056fea26469706673582212204a74ed7a5c39a0d0349b905487b1beb33fba675324cd40ccafe79f2c8ae943b964736f6c634300080d0033"
```
arminsabouri: Do you mind uploading the contract you used [here](https://github.com/botanix-labs/contracts) as a PR

bin/test-suite/src/suite/consensus/common/contracts/MintingAttack.sol
```diff
@@ -0,0 +1,42 @@
+// SPDX-License-Identifier: MIT
```
arminsabouri: Testing utlils should probably live as a part of the test suite or [here](https://github.com/botanix-labs/contracts)
Not in consensus common.

crates/chainspec/res/genesis/botanix_testnet.json
```diff
@@ -16,6 +16,10 @@
...
...104af565b9050919050565b60006104ee826104d1565b9050919050565b6104fe816104e3565b82525050565b600060208201905061051960008301846104f5565b92915050565b6105288161023f565b82525050565b6105378161027d565b82525050565b610546816102b3565b82525050565b600082825260208201905092915050565b82818337600083830152505050565b6000601f19601f8301169050919050565b6000610589838561054c565b935061059683858461055d565b61059f8361056c565b840190509392505050565b600060a0820190506105bf600083018961051f565b6105cc602083018861052e565b6105d9604083018761053d565b81810360608301526105ec81858761057d565b90506105fb608083018461051f565b979650505050505050565b7f4e487b7100000000000000000000000000000000000000000000000000000000600052601260045260246000fd5b60006106408261027d565b915061064b8361027d565b92508261065b5761065a610606565b5b828204905092915050565b6000604082019050818103600083015261068181868861057d565b9050818103602083015261069681848661057d565b905095945050505050565b6106aa8161046f565b81146106b557600080fd5b50565b6000815190506106c7816106a1565b92915050565b6000602082840312156106e3576106e2610215565b5b60006106f1848285016106b8565b9150509291505056fea26469706673582212204a74ed7a5c39a0d0349b905487b1beb33fba675324cd40ccafe79f2c8ae943b964736f6c634300080d0033"
```
arminsabouri: Testing vectors should not be deployed in the genesis block. They can be deployed in a normal fashion

arminsabouri (changes requested):
A few comments and concerns about the placment of test utils. Awesome work

bin/test-suite/mint_attack_contract_abi.json
scottmillner: that's defining the internal member:
```i
interface IMintable {
    function mint(
        address destination,
        uint256 amount,
        uint32 bitcoinBlockHeight,
        bytes calldata metadata,
        address refundAddress
    ) external;

    function burn(
        bytes calldata destination,
        bytes calldata data
    ) external payable returns (bool success);
}
```
which is how the contract instantiates: ` Mintable public constant mintingContract` to make calls to Minting.sol

so it's intended to represent Minting.sol so I named it `mintingContract`

the contract itself is name `MintAttackContract`

crates/chainspec/res/genesis/botanix_testnet.json
scottmillner: this is while i figure out why im having issue deploying it in the test-suite

this allows me to finish the int tests and see if the changes have fixed the audit problem
then ill circle back and figure out the deployment issue

but agreed this should not be merged with predeploy in the genesis block

scottmillner: ive already rolled back since i need to figure out the original deployment issue

bin/test-suite/mint_attack_contract_abi.json
scottmillner: done!
https://github.com/botanix-labs/Contracts/pull/1

bin/test-suite/src/suite/consensus/common/contracts/MintingAttack.sol
scottmillner: moved contract to contracts repo
https://github.com/botanix-labs/Contracts/pull/1

crates/ethereum/evm/src/execute.rs
```diff
@@ -290,12 +288,11 @@ where
             cumulative_gas_used += result.gas_used();
 
             // ***** Botanix specific checks ******
-            let pegout_amount = transaction.value();
             let mut pegins = vec![];
             let mut pegouts = vec![];
 
             let new_result = {
-                if result.is_success() && transaction.to() == Some(*MINT_CONTRACT_ADDRESS) {
+                if result.is_success() {
```
SatwikPrabhu, line 295: I think this change is not part of this PR

crates/ethereum/evm/src/execute.rs
```diff
@@ -322,12 +319,11 @@ where
...
-                                    Self::increment_balance_by_address(
-                                        *sender,
-                                        EthersU256::from_little_endian(pegout_amount.as_le_slice()),
-                                        &mut state,
-                                    );
+                                MintContractError::InvalidPegoutData(PegoutDataError::Invalid(
+                                    _,
+                                    amount,
+                                )) => {
+                                    Self::increment_balance_by_address(*sender, amount, &mut state);
```
SatwikPrabhu, line 326: Are these changes fmt changes?

crates/ethereum/evm/src/execute.rs
```diff
@@ -510,7 +506,7 @@ where
                     error: "No proofs found in pegin data".to_string(),
                     revert_address: pegin_data.account,
                     revert_amount: pegin_data.amount,
-                })
+                });
```
SatwikPrabhu, line 509: Same as https://github.com/botanix-labs/Macbeth/pull/638/files#r2029404293 for this and the next couple of places

SatwikPrabhu (commented):
I think the changes look good to me. Great work 🔥 
Just a minor comment about changes not related to this PR

crates/ethereum/evm/src/execute.rs
scottmillner: it needs to be included in this pr or we can't test the functionality since any calls from the attack contract won't be picked up bc the top level call isn't to Minting.sol

scottmillner (approved):

bin/test-suite/src/mint_attack_contract_abi.rs
lamafab, line 1: Can we get more info about this and the corresponding `.sol` source code?

lamafab (approved):

bin/test-suite/src/mint_attack_contract_abi.rs
scottmillner: it's in our contracts repo: https://github.com/botanix-labs/Contracts/blob/main/src/MintAttack.sol

SatwikPrabhu:
Approve from my end as well!

scottmillner:
runner disk space is full
ive run unit tests and int tests locally and all pass

PR #640, opened 2025-03-29
test (invalid pegout): use mint attack contract

SatwikPrabhu:
the scope off this pr changed since the original issue (validate all calls) was fixed in: https://github.com/botanix-labs/Macbeth/pull/638

this pr now just refactors `test_invalid_pegin` to use the mint attack contract so we ensure we are validating internal calls to Minting.sol and not just external ones. `test_invalid_pegout` already uses the attack contract

### DIFF
--- a/bin/test-suite/src/suite/consensus/common/botanix_client.rs
+++ b/bin/test-suite/src/suite/consensus/common/botanix_client.rs
@@ -184,6 +184,30 @@ impl BotanixEthClient {
         Ok(tx_receipt)
     }
 
+    pub async fn mint_attack(
+        &self,
+        destination: EtherAddress,
+        amount: ethers::core::types::U256,
+        bitcoin_block_height: u32,
+        metadata: ethers::core::types::Bytes,
+        refund_address: EtherAddress,
+    ) -> Result<Option<TransactionReceipt>, Error> {
+        let gas_price = self.http_client.get_gas_price().await.ok().unwrap_or_default();
+
+        let tx_receipt = self
+            .mint_attack_contract.as_ref()
+            .expect("mint attack contract exists")
+            .pass_through_mint(destination, amount, bitcoin_block_height, metadata, refund_address)
+            .gas_price(gas_price)
+            .gas(U256::from(1_000_000))
+            .send()
+            .await
+            .map_err(Error::Contract)?
+            .await
+            .map_err(Error::Provider)?;
+        Ok(tx_receipt)
+    }
+
     pub async fn burn(
         &self,
         destination: ethers::core::types::Bytes,
--- a/bin/test-suite/src/suite/consensus/invalid_transactions/test_invalid_pegin.rs
+++ b/bin/test-suite/src/suite/consensus/invalid_transactions/test_invalid_pegin.rs
@@ -1,3 +1,4 @@
+use crate::suite::consensus::common::events::SEND_AMOUNT;
 use std::{str::FromStr, time::Duration};
 
 use bitcoin::{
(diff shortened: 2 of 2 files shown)


SatwikPrabhu:
Working on the int test for this

crates/ethereum/evm/src/execute.rs
```diff
@@ -295,7 +295,7 @@ where
             let mut pegouts = vec![];
 
             let new_result = {
-                if result.is_success() && transaction.to() == Some(*MINT_CONTRACT_ADDRESS) {
```
arminsabouri: Won't this cause the panic below on L335? It seems like we need more narrow and specific validation and not remove the existing validation. Unfortunately I don't have a suggestion

arminsabouri (changes requested):

crates/ethereum/evm/src/execute.rs
scottmillner: Agreed, this code will panic as written

it's good practice to run the int tests locally before opening a pr just to make sure nothing broke
i like to run `frost_e2e_stable` since it tests pegins/pegouts and block building
then i run the whole suite

SatwikPrabhu: Thanks for pointing this out. I'll try to figure out how to solve this and make sure the tests pass. Btw the test is seemingly passing, but the ci fails
Screenshot 2025-03-31 at 07 21 40

arminsabouri: Whatever fix is deployed, a unit test should be accompanying it

SatwikPrabhu: @scottmillner Do you have an idea of how we would do the int test for this. Cause the current test suite seems to be supporting only calls to the mint contract. Would appreciate any pointers

scottmillner: @SatwikPrabhu make sure you cargo build reth and btc-server before running int tests so your changes are reflected

ill get back to you on how to write int tests

SatwikPrabhu: > @SatwikPrabhu make sure you cargo build reth and btc-server before running int tests so your changes are reflected

The above snapshot is from the ci run itself. The log says that the test passes but the job fails

SatwikPrabhu: > Won't this cause the panic below on L335?

The panic would happen if the call is not to a mint contract address. This check happens in `try_parse_mint_event`. Isnt that expected behavior?
Screenshot 2025-03-31 at 10 56 42Also, this is the test passing locally for me with my changes. So, I dont think the above would panic normally

arminsabouri: >The panic would happen if the call is not to a mint contract address. This check happens in try_parse_mint_event. Isnt that expected behavior?

No b/c you have removed out the mint contract address check. If i understand correctly, this code will try to parse logs from any result and produce an `MintContractError::InvalidLog`.

Logs coming from the minting contract should never be invalid. In that case we should panic.

arminsabouri: to int test this you would need to create a test contract that emmits some log and ensure you don't interfere with minting contract logs. Easier to unit test imo

SatwikPrabhu: > No b/c you have removed out the mint contract address check. If i understand correctly, this code will try to parse logs from any result and produce an MintContractError::InvalidLog.

I think currently we have this piece of code in both `try_parse_burn_event` and `try_parse_mint_event`:

`if log.address != *MINT_CONTRACT_ADDRESS {
        return Ok(None);
    }
`

So, if the log is not from a mint contract address, we wont return any burn/mint event. Should we just return an invalid log error here?

I'm not sure how we would produce the invalid log error for logs from the mint contract. Wouldnt the CI tests fail if this was the case?

scottmillner: apologies, you are correct:
the code will not panic and it looks like this would check all calls to the minting contract even if not at the top level (EOA) but a contract

in your other pr, i've described how we can write an int test for the specific attack:
https://github.com/botanix-labs/Macbeth/pull/638

we'll use the same approach for both

scottmillner: `So, if the log is not from a mint contract address, we wont return any burn/mint event. Should we just return an invalid log error here?`

`I'm not sure how we would produce the invalid log error for logs from the mint contract. Wouldnt the CI tests fail if this was the case?`

What @0xBEEFCAF3 is saying you also need to test a valid tx that emits a log that is not mint/burn and we don't reject it.

SatwikPrabhu: Aah gotcha. Maybe we can capture all these cases in a single test.

arminsabouri: yes @SatwikPrabhu you are right this check is redundant but we still need the minting contract address check in `botanix_mint_contract_checks`

SatwikPrabhu: What should we do if this mint contract address check fails?

crates/primitives/src/botanix/mint_validation.rs
```diff
@@ -139,17 +139,15 @@ fn topic_to_address(t: B256) -> Option<Address> {
 pub fn try_parse_mint_event(
     log: &revm_primitives::Log,
 ) -> Result<Option<PeginData>, ParseMintEventError> {
-    if log.address != *MINT_CONTRACT_ADDRESS {
```
scottmillner: we need to keep this check in:
- if we don't someone can deploy the exact same contract (at a different address) and pegin/pegout.

crates/ethereum/evm/src/execute.rs
arminsabouri: if its from the minting contract same behavior as before.

scottmillner (approved):

SatwikPrabhu:
Looks good to me. Approve from my side

lamafab (approved):

PR #641, opened 2025-03-29
fix(dkg): Check all participants before proceeding with `part2`

lamafab:
(_Text has been revised with Claude.AI_)

This PR offers a simpler, safer alternative to the proposed implementation of a third confirmation round using the coordinator as a broker.

## The Current Issue

Let's examine this critical piece of logic in the `main` branch:

```rust
// Retrieve round 1 packages from peers
// Here we don't check we have enough that should be done by the frost lib
// So we just propagate the error
let round1_packages = self.db.get_round1_dkg_packages().to_status()?;

let (round2_secret_package, round2_packages) =
    frost::keys::dkg::part2(round1_dkg.0.clone(), &round1_packages).to_status()?;
```

The comment suggests that the `frost` library enforces participation requirements. **However, the `frost` library only requires contributions from at least `min_signers` participants, NOT ALL `max_signers`** (minus one). This means the DKG protocol will proceed to `part2` as soon as the minimum threshold is reached, leaving some participants behind. This PR solves this issue by implementing a simple check that ensures all participants have submitted their round1 packages before proceeding to `part2`. This also ensures that `part3` can only be computed if all participants submitted their round2 packages.

@birozuru Please try your 11/15 member test again as soon as you get the chance.

### A Single, Third Confirmation Round Is Not Enough

A proper confirmation system would require two phases - _confirmation_ AND _acknowledgment_. Consider this scenario:

Alice, Bob, and Eve complete all DKG rounds and compute the aggregated public key, then send confirmations. Alice and Bob receive confirmations from everyone, but Eve never receives Alice's confirmation. **Without an acknowledgment round, Alice and Bob have no way of knowing that Eve is still waiting for a confirmation.**

### Network Reliability Assumptions

Our existing DKG protocol assumes reliable message delivery to all peers. It has no retry mechanism or way to request missing messages. Implementing a proper acknowledgment system would require redesigning the entire DKG process - something we'll need to do in the medium-to-long term anyway. Adding just a confirmation round without acknowledgment provides no real benefit, as it doesn't solve the core issue of ensuring all participants complete the DKG process successfully.

### Pragmatic Launch Considerations

We're launching with a permissioned set of multisig participants. This initial setup will be somewhat supervised, allowing us to manually verify that each participant has correctly completed the DKG process and saved their keys.

## Recommendation

I don't recommend making steep protocol changes on such a tight timeline, especially when we'll need to redesign the entire DKG process later. Additionally, at launch, **this DKG process happens just once**, with keys being retained long-term.

This PR provides an immediate safety improvement without introducing complex protocol changes.


scottmillner:
Regarding:

"Alice, Bob, and Eve complete all DKG rounds and compute the aggregated public key, then send confirmations. Alice and Bob receive confirmations from everyone, but Eve never receives Alice's confirmation. Without an acknowledgment round, Alice and Bob have no way of knowing that Eve is still waiting for a confirmation."

I was thinking the confirmations would only be sent to the coordinator to reduce p2p communication.  Once the coordinator has confirmations from all signers, dkg is done.

scottmillner (approved):

rwlockbg (approved):

arminsabouri:
> The comment suggests that the `frost` library enforces participation requirements. **However, the `frost` library only requires contributions from at least `min_signers` participants, NOT ALL `max_signers`** (minus one).

This is catagorically false statment, both the math and code enforce a n-1 commitments to be provided: https://github.com/ZcashFoundation/frost/blob/a1350ea18206a812975740207f90fb121883a5b3/frost-core/src/keys/dkg.rs#L405-L407. Perhaps you are talking about some other code and I am misunderstanding. 

>A proper confirmation system would require two phases - confirmation AND acknowledgment. Consider this scenario:
>Alice, Bob, and Eve complete all DKG rounds and compute the aggregated public key, then send confirmations. Alice and Bob receive confirmations from everyone, but Eve never receives Alice's confirmation. Without an acknowledgment round, Alice and Bob have no way of knowing that Eve is still waiting for a confirmation.

This is not quite what I had in mind. First the success cert which is a Vec of signed payloads over the agg threshold pk is provided by a cooridnator not p2p. Although p2p can be considered, its neglected in the original proposal in order to simplify the design not to enhance security. One trivial way to enforce acknowledgement is to enforce that block 1 of the chain must always include a NDD tx which is the success cert in full in plaintext. But this is also an enhancment. By virtue of signing the agg pk you are ack'ing the fact that you have finished DKG. Singers should persist their DKG results in DB with a the cert of success as an optional. If a signer does not have a success cert they can simply request it. So even if a signer has not recieved the success cert they can always do it after the fact. There is a lot of edge cases here but until they have a success cert they should treat the DKG results as non-existent.

>This PR offers a simpler, safer alternative to the proposed implementation of a third confirmation round using the coordinator as a broker.

I recognize your reasons for not wanting to change the protocol. A third confirmation round alone does not solve the core issue as you have pointed out. There are additional changes required such as a retry mechanism, versioning session with ids, sending all DKG related messages through a coordinator for all rounds of DKG and solving the edge cases that come with that.

BUT, we should also acknowledge that the changes in this PR are NOT an alternative to a third round. I think the change is fine b/c we get a specific error message instead of something ambigious from the FROST lib. 

This is a hard problem with many moving parts. Thanks for staying on this and proposing solutions.

lamafab:
> This is catagorically false statment, both the math and code enforce a n-1 commitments to be provided:

I just double-checked on my side and you're right; I made the mistake on taking AI's word on this, and not actually verifying the hypothesis myself. But then how is it even possible that @birozuru managed to create a 11-out-of-15 subset multisig? Or was he explicitly testing for that case where `part3` fails (which is an edge case, realistically speaking)? My understanding was that when not all Fed member spin up at the same time, then the minimum subset finalize the DKG, leaving the others behind. Considering that `part2` does enforce `round1_packages.len() == (secret_package.max_signers - 1) `, this should never have been possible in the first place(?)

> This is not quite what I had in mind [...] By virtue of signing the agg pk you are ack'ing the fact that you have finished DKG

I'm aware, but at this point I should just take that one week to implement everything correctly (per-member encryption channels, recovery mechanism, `dkg.rs` refactor, etc) instead of implementing a (somewhat hacky) "quick fix".

arminsabouri:
>My understanding was that when not all Fed member spin up at the same time, then the minimum subset finalize the DKG, leaving the others behind.

All N parties need to contribute r1 payloads for r2 to start. Similarly, you need n-1 r2 payloads from your peers to complete r2. An T:N setup doesn't mean that T-N parties were left behind and never finished (thats more applicable to signing then key setup). a T quorum just informs how each party should create their polynomial.

PR #642, opened 2025-03-31
fix/remove-tracked-tx-when-deeply-confirmed

scottmillner:


arminsabouri (approved):
Thanks for finding this 
Is there a unit test to update or a way to prevent  this in the future.

scottmillner:
> Thanks for finding this Is there a unit test to update or a way to prevent this in the future.

there is `test_finalize_block` that should test this but it was passing before
will see why it passed when it shouldn't have and refactor test

rwlockbg (approved):

SatwikPrabhu (approved):

arminsabouri (approved):

PR #644, opened 2025-03-31
fix(btc-server): only validate outputs up to upper bound

scottmillner:
Changing the validation logic to only enforce that a psbt pegout exists in the pending pegouts list. Not that all pending pegouts need to be in the psbt.  This is overly restrictive and all pegouts will eventually get processed if there's more than the upper bound allowed in a single pegout.

### DIFF
--- a/bin/btc-server/src/bin/main.rs
+++ b/bin/btc-server/src/bin/main.rs
@@ -29,6 +29,7 @@ use btcserverlib::{
     util::{
         btc_per_kb_to_sat_per_vb, deserialize_frost_peer_id, get_available_utxos,
         get_pegin_confirmation_depth, parse_eth_address, parse_signing_session_id, ParsingError,
+        UPPER_PEGOUT_BOUND
     },
     wallet::{
         self,
@@ -56,19 +57,6 @@ use btcserverlib::{
 };
 
 const JWT_HEADER_KEY: &str = "trace-proto-bin";
-
-/// The upper bound on pegouts in a single transaction. We use a _reasonable_
-/// number based on the following properties:
-///
-/// * Bitcoin's upper bound on a transaction is 100kb
-/// * 32 bytes required for an output (ie. pegout)
-/// * 110 bytes required for an input
-/// * We assume each output has an input
-///   * In practice, it's more likely that an individual input will map to multiple smaller outputs.
-///     But a larger output could also consume multiple inputs.
-///
-/// With a pegout bound of 500, we can conclude: (32 + 110) * 500 = 71_000
-const UPPER_PEGOUT_BOUND: usize = 500;
 
 macro_rules! already_exists {
     ($($arg:tt)*) => {{
--- a/bin/btc-server/src/bin/main.rs
+++ b/bin/btc-server/src/bin/main.rs
@@ -29,7 +29,7 @@ use btcserverlib::{
     util::{
         btc_per_kb_to_sat_per_vb, deserialize_frost_peer_id, get_available_utxos,
         get_pegin_confirmation_depth, parse_eth_address, parse_signing_session_id, ParsingError,
-        UPPER_PEGOUT_BOUND
+        UPPER_PEGOUT_BOUND,
     },
     wallet::{
         self,
@@ -849,6 +849,7 @@ where
 
         // We just signed for all pending pegouts lets start tracking them
         if cfg!(feature = "conflicting_input") {
+            info!("get_round2_signing_package: removing pending pegouts");
             let pending_pegouts = self.db.get_pending_pegouts().to_status()?;
             let pending_pegout_ids =
                 pending_pegouts.iter().map(|p| p.id).collect::<Vec<PegoutId>>();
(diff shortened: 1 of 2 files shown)


bin/btc-server/src/util.rs
```diff
@@ -400,7 +413,10 @@ pub(crate) fn validate_outputs(psbt: &Psbt, db: &database::Db) -> Result<(), Val
         return Err(ValidateOutputsError::DuplicateOutputs);
     }
 
-    for pegout_id in pending_pegout_ids.iter() {
+    // only enforce up to max number of pegouts
```
lamafab: Do note that `coord_pending_pegouts` - which applies the `max` param - first SORTS the pegouts in ascending order, based on height. This might not work as expected.

Slightly related question; what happens if new pending pegouts are detected before this `validate_psbt` function is called? I would not do this check at all, but rather only check whether the pegout actually exists.

lamafab (changes requested):

bin/btc-server/src/util.rs
lamafab: On line 395:

```rust
    let pending_pegouts = db.get_pending_pegouts()?;
```

You might want to use `db.coord_pending_pegouts(UPPER_PEGOUT_BOUND)` here.

scottmillner: `Slightly related question; what happens if new pending pegouts are detected before this validate_psbt function is called? I would not do this check at all, but rather only check whether the pegout actually exists.`

I believe we have a ticket to stop processing new pending pegouts once we start the psbt process but not sure. 
But you bring up a good point: should just check that pegout actually exists. The pending_pegouts list check is overly restrictive and error prone if pegout comes in after make_tx has started

scottmillner: And yes, my changes most likely don't have the intended effect bc of sorting.  

I'll use `let pending_pegouts = db.coord_pending_pegouts(UPPER_PEGOUT_BOUND)?;`

scottmillner: @lamafab I'm using `cord_pending_pegouts` now and only checking the psbt pegout exists in the pending pegouts list not that the psbt needs to include all the pending pegouts.

all pegouts should still be handled and we don't have to worry about a pegout coming in while were creating the psbt. thanks for the feedback!

lamafab (dismissed):

SatwikPrabhu (dismissed):

bin/btc-server/src/bin/main.rs
```diff
@@ -924,20 +913,27 @@ where
...
-        // tx now
+        // tx now.
+        // If pending pegouts is empty, it means the coordinator signed and removed pegouts and tracked the tx in round 2
         let pending_pegouts = self.db.get_pending_pegouts().to_status()?;
-        let pegout_ids = psbt
-            .pegout_ids()
-            .iter()
-            .map(|p| PegoutId::from_bytes(p).expect("values are 36 bytes"))
-            .collect::<Vec<PegoutId>>();
+        if !pending_pegouts.is_empty() {
```
lamafab: Just wondering; doesn't the existing implementation simply result in a no-op when `pending_pegouts` is empty? Or do those `db` operations behave weirdly if you pass on an empty `pending_pegouts` (and `pegout_ids`)? To me it looks like the logic does not change here.

bin/btc-server/src/util.rs
```diff
@@ -400,9 +418,14 @@ pub(crate) fn validate_outputs(psbt: &Psbt, db: &database::Db) -> Result<(), Val
         return Err(ValidateOutputsError::DuplicateOutputs);
     }
 
-    for pegout_id in pending_pegout_ids.iter() {
-        if !psbt_pegout_ids.contains(pegout_id) {
-            return Err(ValidateOutputsError::MissingPsbtPegout(*pegout_id));
+    // check psbt pegout exists in pending pegouts list
+    // if round 2, signers have already cleared the pending pegouts included
+    // in the psbt and tracked the tx
```
lamafab, line 423: Just to double check; in case round 2 fails, those pegouts go back into the pending pegouts pool, right?

lamafab (dismissed):

bin/btc-server/src/util.rs
scottmillner: - from the signers viewpoint, the psbt is finalized and they will track after signing round 2 (which has already happended at this point). They will not add it back.  Also, we will be removing the requirement to enforce the pending pegouts list once we have the finalized pegouts list and the conflicting input feature enabled
- from the coordinator viewpoint, at this point the psbt will be finalized and broadcast.  Only if the broadcast is successful will the pending pegouts be cleared.  

Round 2 here is actually the end of Round 2.

scottmillner: updated the comment

bin/btc-server/src/bin/main.rs
scottmillner: yep you're right, this is a no-op and the check has no impact
removed the check, thanks!

lamafab (dismissed):

scottmillner:
this was already approved but needed to run ci on another runner

PR #645, opened 2025-04-04
fix: fixed all expects and unwraps in abci.rs

rwlockbg:
https://github.com/botanix-labs/botanix/issues/974

From SigmaPrime:

"In abci.rs several .expect() and panic!()s are used instead of explicit error handling. For example, on line 1328, a panic occurs if transactions cant be decoded."


crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -558,22 +572,47 @@ where
...
         let client = self.storage.client.clone();
-        ResponseInitChain { app_hash: self.application_hash(&client), ..Default::default() }
+        let app_hash = match self.application_hash(&client) {
+            Ok(app_hash) => app_hash,
+            Err(e) => {
+                error!("Error getting application hash: {:?}", e);
+                return ResponseInitChain::default();
+            }
+        };
+        ResponseInitChain { app_hash, ..Default::default() }
```
arminsabouri, line 583: I'm not sure what the expected behavior here should be. It seems ok to panic! if we cannot read the DB when the chain is being init'd. Responding default app_hash may result in some odd behavior. Lmk what you think

bin/btc-server/src/util.rs
```diff
@@ -621,9 +621,9 @@ mod tests {
                 .unwrap_or_default()
         });
 
-        let diff = total_inputs.checked_sub(total_outputs).unwrap_or_default().to_sat()
-            / psbt.unsigned_tx.output.len() as u64
-            + 100;
+        let diff = total_inputs.checked_sub(total_outputs).unwrap_or_default().to_sat() /
+            psbt.unsigned_tx.output.len() as u64 +
+            100;
```
SatwikPrabhu: Noticed that there a bunch of formatting changes as part of this PR. Is that something we want to do?

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -670,8 +717,10 @@ where
             // read the lock and make sure we are not already syncing the snapshot we are being
             // offered
             if let Some(snapshot_sync_state_lock) = self.snapshot_sync_state_lock.as_ref() {
-                let snapshot_sync_state_lock =
-                    snapshot_sync_state_lock.read().expect("snapshot state sync locked");
+                let Ok(snapshot_sync_state_lock) = snapshot_sync_state_lock.read() else {
+                    error!("Snapshot state sync locked");
+                    return ResponseOfferSnapshot { result: SnapshotOfferResult::Reject as i32 };
+                };
```
SatwikPrabhu: Why not use a match statement here. This applies to many places in this file. I think we can be consistent by either using match arms or a let statement consistently. Dont have a particular preference for either

crates/consensus/authority/src/comet_bft/abci.rs
scottmillner: good point

this will cause an app hash mismatch error on the comet side since it'll have a different hash then default so a panic will happen no matter what either in abci or comet

im fine either way, explicitly panic! here or let comet throw the error

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -1457,10 +1584,23 @@ where
     /// docs: https://docs.cometbft.com/v0.38/spec/abci/abci++_methods#commit
     fn commit(&self) -> ResponseCommit {
         info!("commit request received");
-        let candidate_blocks = self.block_cache.write().expect("to get write lock");
+        let candidate_blocks = match self.block_cache.write() {
```
scottmillner, line 1630: we should panic! in `fn commit` and not return default when there's an error

we can't return to comet if we don't commit that block to the reth chain

```
driver_tx.send(ABCIDriverMessage::CommitBlock((sealed_block_with_context, commit_tx)))
```

if we do, comet will return app hash mismatch the next time we go to add a block

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -1542,7 +1685,13 @@ where
                             Arc::new(trie_updates.clone()),
                         );
 
-                        let db_rw = self.database_provider.provider_rw().expect("to get db rw");
```
scottmillner, line 1545: related to the comment about `fn commit`
if driver gets a commit message it has to add it to the canonical chain or panic

at this point, comet has already added it to it's chain so we're going to hit an app mismatch error anyway

there's no graceful way to recover from this scenario: poa node should crash and we'd have to do a manual rollback to a healthy state and sync with the network

rwlockbg: acknowledged

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -968,18 +1039,21 @@ where
         let (compressor_task_tx, compressor_task_rx) =
             tokio::sync::oneshot::channel::<Vec<BlockWithSenders>>();
         self.task_executor.spawn_blocking(Box::pin(async move {
-            let blocks_with_senders: Vec<BlockWithSenders> = compressor
-                .decode(request.chunk.as_ref())
-                .await
-                // TODO: handle the error
-                .expect("Failed to deserialize and decompress block with context");
-            let _ = compressor_task_tx.send(blocks_with_senders);
+            if let Ok(blocks_with_senders) = compressor.decode(request.chunk.as_ref()).await {
```
lamafab: What's the consequence of skipping that chunk and shouldn't that be communicated to the ABCI client via the `refetch_chunks` field? Also note that the return type has a `reject_senders` field, FYI:

```rust
pub struct ResponseApplySnapshotChunk {
    #[prost(enumeration = "response_apply_snapshot_chunk::Result", tag = "1")]
    pub result: i32,
    /// Chunks to refetch and reapply
    #[prost(uint32, repeated, tag = "2")]
    pub refetch_chunks: ::prost::alloc::vec::Vec<u32>,
    /// Chunk senders to reject and ban
    #[prost(string, repeated, tag = "3")]
    pub reject_senders: ::prost::alloc::vec::Vec<::prost::alloc::string::String>,
}
```

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -968,18 +1039,21 @@ where
...
 
         // await the response from the compressor task
-        let blocks_with_senders = compressor_task_rx
-            .blocking_recv()
-            .expect("to receive confirmation from snapshot manager");
+        let Ok(blocks_with_senders) = compressor_task_rx.blocking_recv() else {
+            error!("Failed receive confirmation from snapshot manager");
+            return ResponseApplySnapshotChunk {
+                result: ApplySnapshotResult::RetrySnapshot as i32,
+                refetch_chunks: vec![],
```
lamafab: Shouldn't the chunks be indicated in `refetch_chunks` so that the ABCI client knows about them?

lamafab (changes requested):

crates/consensus/authority/src/comet_bft/abci.rs
rwlockbg: Fixed

crates/consensus/authority/src/comet_bft/abci.rs
rwlockbg: agree with you, fixed

crates/consensus/authority/src/comet_bft/abci.rs
rwlockbg: Yes, nice observation. I have refactored that now everywhere.

crates/consensus/authority/src/comet_bft/abci.rs
rwlockbg: Thanks, I agree here. I have made it panic now.

bin/btc-server/src/util.rs
rwlockbg: They should all disappear once https://github.com/botanix-labs/Macbeth/pull/655 is merged.

crates/consensus/authority/src/comet_bft/abci.rs
rwlockbg: @lamafab I am not sure tbh, lets not make this the subject of this PR. What the auditors want is a cleanup of unwraps and expects. Surely, you have a valid point to consider here, but I would rather have this be the subject of an issue or a separate PR.

crates/consensus/authority/src/comet_bft/abci.rs
rwlockbg: @lamafab I am not sure tbh, lets not make this the subject of this PR. What the auditors want is a cleanup of unwraps and expects. Surely, you have a valid point to consider here, but I would rather have this be the subject of an issue or a separate PR.

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -1457,10 +1621,22 @@ where
     /// docs: https://docs.cometbft.com/v0.38/spec/abci/abci++_methods#commit
     fn commit(&self) -> ResponseCommit {
```
scottmillner, line 1628: i shouldve been more clear
we always have to panic in `fn commit` if there's an error

if we return `ResponseCommit::default();` that means we are actually going to commit the block to the canonical chain

if we return but that doesn't happen, there will be an app hash mismatch

scottmillner: ill push a commit fixing this

scottmillner (dismissed):

SatwikPrabhu (dismissed):

lamafab (dismissed):

scottmillner:
pr had 3 approvals before
last commit was just adding/editing comments
merging

PR #646, opened 2025-04-05
Fix(Pegins): Revert invalid pegin bitcoin block height

SatwikPrabhu:
While we explore whether a general fix explained in https://botanixlabs.slack.com/archives/C08HFGV9F1D/p1743698029799619, here is a fall back solution that is less ideal but should work to fix https://github.com/botanix-labs/botanix/issues/963


arminsabouri:
A couple broken tests. Since this fallback is messing with evm state stuff we should really consider bringing back the eth state test CI job.

scottmillner:
closing this for now as it's superceded by https://github.com/botanix-labs/Macbeth/pull/664

PR #647, opened 2025-04-07
feat: added checks on non-deterministic data and max tx size

rwlockbg:
https://github.com/botanix-labs/botanix/issues/975

From SigmaPrime:

"In prepare_proposal(). non_deterministic_data_bytes is being added to the txs but max_tx_bytes is not being checked. This can break CometBFT handling if non_deterministic_data_bytes size is significantly large and overall it exceeds max_tx_bytes limit"


scottmillner (dismissed):

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -1116,7 +1116,19 @@ where
                             .map(|tx| prost::bytes::Bytes::copy_from_slice(tx))
                             .collect::<_>();
                         info!("prepare_proposal number of txs: {:?}", txs.len());
+                        if txs.iter().any(|tx| tx.len() as i64 > request.max_tx_bytes) {
+                            error!(
+                                "Tx exceeds the max tx bytes allowed size {:?}",
+                                request.max_tx_bytes
+                            );
+                            return ResponsePrepareProposal { ..Default::default() };
```
arminsabouri: If one tx in the mempool is over the `max_tx_bytes` should we just ignore that tx and create a block anyways? 
Here it seems like we would never create a block with other txs until the problematic tx is pruned. Is that correct @rwlockbg

arminsabouri (changes requested):
I had a question about checking the tx size of non-NDD txs. 
Also some formatting changes are always included in these PRs. They seem unrelated. Maybe you are using a different rust fmt version

arminsabouri:
Additionally, I would ask if there is a way to unit test this as to avoid regressions in the future

crates/consensus/authority/src/comet_bft/abci.rs
scottmillner: ah nice catch!
yep looks like we'd just create empty blocks forever since we'd
`return ResponsePrepareProposal { ..Default::default() };`

lamafab: Additionally, if I see it correctly, none of those checks are enforced during the backing phase in `process_proposal`(?)

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -1116,7 +1116,19 @@ where
...
+                        }
 
+                        // check that the non-deterministic data is not larger than the max tx bytes
+                        if non_deterministic_data_bytes.len() as i64 > request.max_tx_bytes {
+                            error!("Non-deterministic data size: {:?} exceeds the max tx bytes allowed size {:?}", non_deterministic_data_bytes.len(), request.max_tx_bytes);
+                            return ResponsePrepareProposal { ..Default::default() };
+                        }
```
lamafab, line 1129: Afaik the created `non_deterministic_data_bytes` (in `prepare_proposal`) would only exceed this max if there's a bug on our side. If that happens, we'd permanently halt block production.

lamafab (changes requested):

crates/consensus/authority/src/comet_bft/abci.rs
rwlockbg: @lamafab Agree, but a check wont harm. Plus it is required because of hte audit.

crates/consensus/authority/src/comet_bft/abci.rs
rwlockbg: Thanks, that makes sense. I am now skipping those txs.

rwlockbg:
> I had a question about checking the tx size of non-NDD txs. Also some formatting changes are always included in these PRs. They seem unrelated. Maybe you are using a different rust fmt version

Yes, these are fmt changes. I really dont know where they come from and why is the master formatted for you but not for me. I will check that.

crates/consensus/authority/src/comet_bft/abci.rs
scottmillner: im going to push a commit where we panic! if NDD is larger than `max_tx_bytes` bc there's a critical bug if that happens and we should halt

scottmillner (approved):

rwlockbg:
> > I had a question about checking the tx size of non-NDD txs. Also some formatting changes are always included in these PRs. They seem unrelated. Maybe you are using a different rust fmt version
> 
> Yes, these are fmt changes. I really dont know where they come from and why is the master formatted for you but not for me. I will check that.

Already fixed, no more fmt changes, only the releveant ones are in here.

lamafab (approved):

SatwikPrabhu (approved):

PR #649, opened 2025-04-08
Fix/state sync get correct height

scottmillner:


arminsabouri (approved):
utACK

lamafab (approved):

crates/consensus/authority/src/snapshot_manager.rs
```diff
@@ -457,9 +445,10 @@ where
         }
 
         // check the block height vs. the last snapshot height
+        let mut state_lock = self.state_lock.write().expect("snapshot state sync locked");
```
SatwikPrabhu, line 448: Maybe update the above comment to reflect this update?

rwlockbg (approved):

SatwikPrabhu (approved):

scottmillner:
ran all int tests successfully locally

PR #651, opened 2025-04-08
chore(git): don't track btc-server rpc binary

scottmillner:


.gitignore
```diff
@@ -83,6 +83,9 @@ build_rs_cov.profraw
 bin/test-suite/*.txt
 cobertura.xml
 
+# btc-server rpc binary
+bin/btc-server/src/rpc/btc_server.bin
```
arminsabouri, line 87: Wasn't this used in service interface discovery or something. if so I dont think we have any need for something like that. In fact it may pose as a security concern.

arminsabouri (approved):

.gitignore
rwlockbg: We need this since we are using grpc reflection, i.e. automatic self-discovery. We cannot ignore this file @scottmillner @0xBEEFCAF3

lamafab: There must be another way... having some binary blob in a repository like this is a huge red flag.

rwlockbg: The binary is being build in the build.rs file I have implemented - https://github.com/botanix-labs/Macbeth/blob/main/bin/btc-server/build.rs @lamafab it is a matter of choice ... whoever is using the btc-servers needs to build it first and the bin must exist alongside with the binary file. We can ignore it for sure, as @scottmillner has suggested, I have no issues with that, but we shd tell devs to build btc-server at all costs so that the bin file gets generated.

rwlockbg (approved):

.gitignore
scottmillner: thanks for the comments, im going merge the pr

devs just need to `cd bin/btc-server cargo build` and the binary will be generated where it needs to be bc of `build.rs`

PR #652, opened 2025-04-09
Feature/pegouts enable conflicting inputs feature

scottmillner:
Overview of the changes (@0xBEEFCAF3 definitely would like to talk through them):

- Signers don't enforce pending pegouts (https://github.com/botanix-labs/Macbeth/pull/634 has finalized outputs)
- Signers track txs after round2
  - the tracked tx contains the pending pegouts in the psbt that the signers had in their db when signing.
  - the way syncing works, signers can have a superset of pending pegouts compared to the psbt
- Signers don't enforce psbt utxos must be in their db
- But if the utxos do exist in their db, signers will validate them


bin/btc-server/src/signer/mod.rs
```diff
@@ -21,21 +21,6 @@ pub async fn get_round1_signing_package(
     db: &Database,
     my_identifier: &Identifier,
 ) -> Result<Vec<(SigningNonces, SigningCommitments)>, SigningRound1Error> {
-    // self.db.get_key_package()?.ok_or(SigningRound1Error::MissingKeyPackage)?;
-    // // Check if have already provided nonces for the current session
```
scottmillner, line 25: this check is done right before calling this method and is leftover from a previous refactor

bin/btc-server/src/bin/main.rs
```diff
@@ -845,20 +848,39 @@ where
...
-            let pending_pegouts = self.db.get_pending_pegouts().to_status()?;
-            let pending_pegout_ids =
-                pending_pegouts.iter().map(|p| p.id).collect::<Vec<PegoutId>>();
-            self.add_tracked_tx(signed_tx.clone(), &pending_pegouts, SystemTime::now())
-                .await
-                .to_status()?;
-            self.db.remove_pending_pegout(&pending_pegout_ids).to_status()?;
-            self.db.flush().to_status()?;
+        let signed_tx =
+            psbt.clone().extract_tx().expect("just checked in get_round2_signing_package");
```
arminsabouri, line 852: There is a clone added here that was previously not there. Do we need it?

bin/btc-server/src/bin/main.rs
```diff
@@ -845,20 +848,39 @@ where
...
+        let pending_pegout_ids = pending_pegouts.iter().map(|p| p.id).collect::<Vec<PegoutId>>();
+        let psbt_pending_pegouts = pending_pegouts
+            .into_iter()
+            .filter(|p| psbt_pegout_ids.contains(&p.id))
+            .collect::<Vec<_>>();
+
+        self.add_tracked_tx(signed_tx.clone(), &psbt_pending_pegouts, SystemTime::now())
+            .await
+            .to_status()?;
+        self.db.reset_pending_pegouts(&[]).to_status()?;
```
arminsabouri, line 886: What if I have different pegouts than the one's I just signed for? Is there no value in keeping them?

bin/btc-server/src/signer/mod.rs
arminsabouri: Thanks for checking on this

arminsabouri (commented):
This is definetly on the right course. A few comments

bin/btc-server/src/bin/main.rs
scottmillner: yes, we are using psbt after this line in the for loop

bin/btc-server/src/bin/main.rs
scottmillner: My thought process:
- we're not enforcing the pending pegouts list
- this is ok with conflicting inputs enabled
- so only track the pegout ids in the psbt
- peers will sync on finalized pegouts (separate pr)

bin/btc-server/src/bin/main.rs
```diff
@@ -845,20 +848,43 @@ where
...
+        let pending_pegouts = self.db.get_pending_pegouts().to_status()?;
+        let psbt_pending_pegouts = pending_pegouts
+            .into_iter()
+            .filter(|p| psbt_pegout_ids.contains(&p.id))
+            .collect::<Vec<_>>();
+
+        self.add_tracked_tx(signed_tx.clone(), &psbt_pending_pegouts, SystemTime::now())
+            .await
+            .to_status()?;
+        self.db.reset_pending_pegouts(&[]).to_status()?;
```
SatwikPrabhu, line 886: Doesnt the above code clear all pending pegouts from the database if we provide an empty list as a parameter?

scottmillner: yes and that's the intent
comment for the method
```
        // Note: the coordinator determines the pending pegouts to include in the psbt.
        // Signers may or may not have the same pending pegouts depending on their liveliness.
        // When signers sync with the network they will add the pending pegouts to their
        // database but some of them may have already been honored when they were offline.
        // Signers need to track the pending pegouts included in the psbt and clear the pending
        // pegouts from the database.
```

signers don't enforce their pending pegouts list and they don't sync on pending pegouts (only finalized outputs)
but they do track the ones in their list that are included the tx given by the coordinator

once they've tracked they need to clear their list: if they have pending pegouts that weren't in the tx then it means they've already been honored. the coordinator always creates the psbt therefore it never misses pending pegouts. if the coordinator goes offline (no psbts happen) and it syncs when it comes back up

SatwikPrabhu: Apologies on missing that. Yes thats clear!

SatwikPrabhu (dismissed):

bin/btc-server/src/bin/main.rs
```diff
@@ -845,20 +848,43 @@ where
...
+        // Note: the coordinator determines the pending pegouts to include in the psbt.
+        // Signers may or may not have the same pending pegouts depending on their liveliness.
+        // When signers sync with the network they will add the pending pegouts to their
+        // database but some of them may have already been honored when they were offline.
+        // Signers need to track the pending pegouts included in the psbt and clear the pending
+        // pegouts from the database.
+
+        // Extract pegout ids from the psbt to store with the tx
+        if psbt.outputs.len() > UPPER_PEGOUT_BOUND {
+            return Err(badarg!("Too many pegouts in the psbt"));
```
rwlockbg, line 863: Nit: I would add here to the error message the len of the current psbt outputs vs the upper limit

bin/btc-server/src/coordinator/mod.rs
```diff
@@ -121,50 +121,46 @@ pub async fn make_tx(
...
+    let tracked_pegout_request_ids = tracked_txs
+        .iter()
+        .flat_map(|tx| tx.pegout_requests.iter().map(|p| p.id))
+        .collect::<HashSet<_>>();
+    println!("tracked_pegout_request_ids = {:?}", tracked_pegout_request_ids);
+
+    // Collect all pegout ids being retried.
+    let matching_pegouts_ids: Vec<&PegoutId> = outputs
+        .iter()
+        .filter(|(_, pegout_id)| tracked_pegout_request_ids.contains(pegout_id))
```
rwlockbg, line 133: Nit: there is also `filter_map` aggregator method which is more efficient than map and then filter.

rwlockbg (dismissed):
@scottmillner Nice work ,just 2 small nits. Also, there is a conflict atm, might want to resolve it.

bin/btc-server/src/bin/main.rs
lamafab: > // Signers need to track the pending pegouts included in the psbt and clear the pending
> // pegouts from the database.

Shouldn't that be happening in `finalize_signing`? Afaik the returned `SigningPackage` might be dropped.

bin/btc-server/src/coordinator/mod.rs
```diff
@@ -121,50 +121,46 @@ pub async fn make_tx(
 
     // if we are retrying pegouts, we need to add a conflicting input for each tracked tx
     // that honors each pegout
-    let mut conflicting_inputs: Result<Vec<Utxo>, CoordinatorError> = Ok(vec![]);
+    let tracked_pegout_request_ids = tracked_txs
```
lamafab, line 124: I think it would be nice having all those conflicting-input related changes in a single function, such that it can be tested independently from all the other things that `make_tx` does. Ideally in a way such that non-coordinator can reuse said logic for validation.

Just as an optional improvement recommendation (separate PR).

lamafab (dismissed):

bin/btc-server/src/bin/main.rs
scottmillner: `finalize_signing` is only used by the coordinator
signers do this in `get_round2_signing_package`

scottmillner:
already had 3 approvals
just had one small merge conflict to address

bin/btc-server/src/bin/main.rs
arminsabouri: I'm not sure this fixes the problem above. If I sign for a psbt with outputs, some of which I am tracking but I also am tracking other valid pegouts that seems to be a loss in information. And a sync in finalized pegouts will not recover that information

bin/btc-server/src/coordinator/mod.rs
```diff
@@ -121,50 +121,46 @@ pub async fn make_tx(
...
+        .iter()
+        .flat_map(|tx| tx.pegout_requests.iter().map(|p| p.id))
+        .collect::<HashSet<_>>();
+    println!("tracked_pegout_request_ids = {:?}", tracked_pegout_request_ids);
```
arminsabouri, line 128: use a macro from log not println!

bin/btc-server/src/util.rs
```diff
@@ -386,21 +375,14 @@ pub enum ValidateOutputsError {
...
-/// Check all pending pegouts are being settled in this tx
-/// and additional outputs are change outputs
-pub(crate) fn validate_outputs(
-    psbt: &Psbt,
-    db: &database::Db,
-    is_round_2: bool,
-) -> Result<(), ValidateOutputsError> {
+/// Check:
+/// - additional outputs are change outputs
+/// - there are no duplicate outputs
```
arminsabouri, line 380: Nit: check there are no duplicatoin pegout outputs

arminsabouri (commented):
I see this Pr is already merged but thought I would still leave my review. I have one comment about removing pending pegouts when it may not be neccecary. This one should definetly get some attention.  The rest are nitpicks

bin/btc-server/src/bin/main.rs
scottmillner: Thanks for bringing that up. I was thinking a sync in finalized pegouts would recover this information.

- Yes, signers will only track pegouts in the psbt they are aware of.
- This is loss of information on a per signer basis.
- But that pegout isn't lost as long as one of the signers does have it and tracks it when signing the psbt.
- The coordinator will always track it since it created the psbt so must've had it in it's pending pegouts list.
- Then it will be added to finalized outputs when the tracked tx is deeply confirmed.
- Signers sync on a superset of finalized outputs to check against when signing

arminsabouri: >The coordinator will always track it since it created the psbt so must've had it in it's pending pegouts list.

But this is a case where the coordinator has not tracked it since it wasnt included in the psbt that was just signed. 
Since all signers will also remove all pending pegouts after signing this information will be lost

PR #653, opened 2025-04-09
feat: added grpc streaming for pegout ids

rwlockbg:


bin/btc-server/src/bin/main.rs
```diff
@@ -612,16 +616,37 @@ where
...
-        }
-        let res = tonic::Response::new(rpc::GetFinalizedPegoutIdsResponse {
-            ids: finalized_pegout_ids.into_iter().map(|p| p.as_bytes().to_vec()).collect(),
+        let db = self.db.clone();
+        let telemetry = self.telemetry.clone();
+
+        let (tx, rx) = tokio::sync::mpsc::channel(10);
+        tokio::spawn(async move {
+            loop {
+                let finalized_pegout_ids = match db.get_finalized_pegout_ids().to_status() {
```
scottmillner: is there a way to not load all the ids into memory but incrementally?
can we do that with sled?
like using `skip()` and `take()`?

rwlockbg: done

rwlockbg: pls check the latest version

bin/btc-server/src/bin/main.rs
```diff
@@ -611,17 +613,47 @@ where
...
-            telemetry.update_pending_pegouts(finalized_pegout_ids.len() as i64);
-        }
-        let res = tonic::Response::new(rpc::GetFinalizedPegoutIdsResponse {
-            ids: finalized_pegout_ids.into_iter().map(|p| p.as_bytes().to_vec()).collect(),
+        let db = self.db.clone();
+        let telemetry = self.telemetry.clone();
+        let request = req.into_inner();
+
+        let (tx, rx) = tokio::sync::mpsc::channel(request.chunk_size as usize);
+        tokio::spawn(async move {
```
lamafab, line 624: Is it necessary to spawn a task here? All those `App` methods are async already and can run in parallel, you can embed the logic here directly.

lamafab (approved):

scottmillner:
General question:

We're streaming chunks of pegout ids, and we'll sync our db once we have gotten a response from min_signers.
But this mean we're just syncing on chunks and not the full set of data from peers?

There's no logic to know if a peer has sent all their chunks and the stream is closed. We just count that we have recieved a response from enough peers, and peers send chunks.

rwlockbg:
> General question:
> 
> We're streaming chunks of pegout ids, and we'll sync our db once we have gotten a response from min_signers. But this mean we're just syncing on chunks and not the full set of data from peers?
> 
> There's no logic to know if a peer has sent all their chunks and the stream is closed. We just count that we have recieved a response from enough peers, and peers send chunks.

THis has been fixed with the last commit .Please have a look @scottmillner

bin/btc-server/src/bin/main.rs
rwlockbg: I need to spawn a task here in order to receive from the stream and push each item into a channel that is being wrapped by tokio_stream::ReceiverStream and returned via the grpc response. These are the mechanics behind grpc unidirectional streaming. As you see the stream also needs to be pinned via: pin_mut!(stream) while it is sending the items till fully drained. Let me know if I am missing something

bin/btc-server/src/database/mod.rs
```diff
@@ -760,15 +760,30 @@ impl Db {
...
     pub fn get_finalized_pegout_ids_stream(
         &self,
         chunk_size: usize,
-    ) -> impl Stream<Item = Result<Vec<PegoutId>, Error>> + Send + '_ + Sync {
+    ) -> impl Stream<Item = Result<(Vec<PegoutId>, u64, u64), Error>> + Send + '_ + Sync {
         async_stream::stream! {
             let mut current_chunk = Vec::with_capacity(chunk_size);
             let mut count = 0;
+            let mut chunk_index: u64 = 0;
+            let num_chunks = match self.peek_finalized_pegout_ids() {
```
scottmillner, line 779: this is the num of pegout ids right? not num of chunks?

shouldn't we take the result of `self.peek_finalized_pegout_ids()` and divide by `chunk_size` to get `num_chunks`?

scottmillner: pushed a commit fixing this

scottmillner (approved):

PR #660, opened 2025-04-11
fix(state-sync): flags default is false

scottmillner:
If we set the default to true there's no way to pass false.
The expected type is a boolean and is based on whether the flag is present or not.  So set default to false and pass the flag if you want it true.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -739,7 +739,7 @@ start-poa-server-3:
 	--p2p-secret-key "${NODE_3_DIR}/discovery-secret" \
 	--port 30305 \
 	--abci-port=46658 \
-  --sync.enable_state_sync \
+    --sync.enable_state_sync \
 	--sync.enable_historical_sync
 
 start-non-fed-server-1:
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -734,7 +738,9 @@ start-poa-server-3:
 	--sync.num_snapshots_to_keep 3 \
 	--p2p-secret-key "${NODE_3_DIR}/discovery-secret" \
 	--port 30305 \
-	--abci-port=46658
+	--abci-port=46658 \
+  --sync.enable_state_sync \
```
SatwikPrabhu: Nit -- fix alignment

scottmillner: locally it is showing aligned so not sure why it's not here
ill try doing a commit here in the repo

scottmillner: that worked, done!

SatwikPrabhu (dismissed):

arminsabouri (dismissed):
cACK

scottmillner:
was already approved just added small change so state sync would happen in int tests
merging

PR #661, opened 2025-04-13
Feature/mainnet enable mainnet on startup

scottmillner:
This enables the mainnet chain to spin up by not passing the `--is-testnet` flag and generating mainnet genesis block. There are checks in `poa > mod > execute()` to make sure chain version matches btc-network and abci client checks that chain id in comet config matches reth chain spec chain id.

@birozuru 
These changes require the following changes in `docker-compose` on the boxes:
- explicity pass `--is-testnet` for poa/rpc node otherwise it will assume mainnet

When we internally test mainnet:
- pass `--btc-network=bitcoin` for btc-server
- update `.bitcoin.env` creds and url to point to mainnet bitcoind instance
- for cometbft `genesis.json` > `chain_id` needs to be `3737`


bin/reth/src/commands/poa/mod.rs
```diff
@@ -267,6 +268,23 @@ impl<Ext: clap::Args + fmt::Debug> PoaNodeCommand<Ext> {
         )?;
         let chain_arc = Arc::new(chain.clone());
 
+        // if mainnet, check btc-network is mainnet
+        if chain.chain.id() == BOTANIX_MAINNET_CHAIN_ID
```
arminsabouri: Looks like we just need one flag. Is there value in two?

crates/chainspec/res/genesis/botanix_mainnet.json
```diff
@@ -0,0 +1,18 @@
...
+  "nonce": "0x00",
+  "timestamp": "0x00",
+  "extraData": "0x000000000000000000000000000000000000000000000000000000000000000000000000000000000279be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f817980000000000000000000000000000000000000000",
+  "gasLimit": "0x1c9c380",
+  "difficulty": "0x0",
+  "mixHash": "0x0000000000000000000000000000000000000000000000000000000000000000",
+  "coinbase": "0x0000000000000000000000000000000000000000",
+  "alloc": {
+    "0x0Ea320990B44236A0cEd0ecC0Fd2b2df33071e78": {
+      "balance": "0x115EEC47F6CF7E35000000",
```
arminsabouri, line 11: Converted to decimal this is 
200000000000000000000000000

crates/chainspec/src/spec.rs
```diff
@@ -166,16 +179,39 @@ pub static BOTANIX_TESTNET: Lazy<Arc<ChainSpec>> = Lazy::new(|| {
...
+pub static BOTANIX_MAINNET: Lazy<Arc<ChainSpec>> = Lazy::new(|| {
+    let mut spec = ChainSpec {
+        chain: Chain::from_id(BOTANIX_MAINNET_CHAIN_ID),
+        genesis: serde_json::from_str(include_str!("../res/genesis/botanix_mainnet.json"))
+            .expect("Can't deserialize Botanix Testnet genesis json"),
+        genesis_hash: Some(BOTANIX_MAINNET_GENESIS),
+        paris_block_and_final_difficulty: Some((0, U256::from(0))),
+        hardforks: EthereumHardfork::botanix().into(),
+        deposit_contract: None, // only relevant for PoS chains
+        parent_confirmation_depth: 6,
```
arminsabouri: This is a huge point of bikeshedding. I generally don't think 6 is sufficient. It's mostly anecdotal that 6 is considered enough. And for most cases it is. Alpen and liquid both use 100 blocks.

crates/chainspec/src/spec.rs
```diff
@@ -166,16 +179,39 @@ pub static BOTANIX_TESTNET: Lazy<Arc<ChainSpec>> = Lazy::new(|| {
...
+        leader_selection_window: Some(20),
+        base_fee_params: BaseFeeParamsKind::Constant(BaseFeeParams::ethereum()),
+        max_gas_limit: ETHEREUM_BLOCK_GAS_LIMIT,
+        prune_delete_limit: 1700,
```
arminsabouri: Any idea what this is?

crates/chainspec/templates/botanix_mainnet_template.json
```diff
@@ -0,0 +1,18 @@
...
+    "nonce": "0x00",
+    "timestamp": "0x00",
+    "extraData": "{{edh}}",
+    "gasLimit": "0x1c9c380",
+    "difficulty": "0x0",
+    "mixHash": "0x0000000000000000000000000000000000000000000000000000000000000000",
+    "coinbase": "0x0000000000000000000000000000000000000000",
+    "alloc": {
+        "0x0Ea320990B44236A0cEd0ecC0Fd2b2df33071e78": {
+            "balance": "0x115EEC47F6CF7E35000000",
```
arminsabouri, line 11: Same issue

arminsabouri (changes requested):

bin/reth/src/commands/poa/mod.rs
scottmillner: i went back and forth on this and was thinking:
- if botanix is testnet but btc server is using mainnet, the user will be given a btc mainnet address to send funds to
- there's no explicit check to stop this so was just being super paranoid here with this extra check

crates/chainspec/res/genesis/botanix_mainnet.json
scottmillner: im getting 210000000000000000000000000

[Screenshot 2025-04-13 at 3 33 04 PM
](https://www.rapidtables.com/convert/number/hex-to-decimal.html?x=115EEC47F6CF7E35000000)

crates/chainspec/src/spec.rs
scottmillner: thanks for bringing this up, i meant to add a comment here:

@PhoenixCodeCraft @jeffgeudens thoughts from the product side?

crates/chainspec/templates/botanix_mainnet_template.json
scottmillner: see screenshot above on first comment

crates/chainspec/src/spec.rs
scottmillner: it's used to limit the number of state entries that can be removed in one go during a pruning operation
we're running default mode which is archive(no pruning)

im not sure why such a low value was used in comparison to the other chainspecs
for example:
- mainnet is 20000
- sepolia is 10000

if we don't remember why, should we use the mainnet value of 20000?

crates/chainspec/res/genesis/botanix_mainnet.json
arminsabouri: Robot lied to me. Looks good

crates/chainspec/src/spec.rs
arminsabouri: I think pruning is still an option on our node. Even though we don't use it someone else might.

bin/reth/src/commands/poa/mod.rs
```diff
@@ -267,6 +268,23 @@ impl<Ext: clap::Args + fmt::Debug> PoaNodeCommand<Ext> {
...
+        }
+        // if testnet, check btc-network is not mainnet
+        if chain.chain.id() == BOTANIX_TESTNET_CHAIN_ID
+            && rpc.btc_network == bitcoin::Network::Bitcoin
+        {
+            return Err(eyre::eyre!(
+                "Chains mismatch: Botanix is testnet and btc network is mainnet."
+            ));
+        }
+
```
rwlockbg: I would use  a match with conditionals here for making it easier to read. E.g. 

`match chain.chain.id() {
 s => if s ==  BOTANIX_MAINNET_CHAIN_ID && rpc.btc_network != bitcoin::Network::Bitcoin { ... }
 s => if s ==  BOTANIX_TESTNET_CHAIN_ID && rpc.btc_network == bitcoin::Network::Bitcoin { ... }
`

or even 
`match (chain.chian.id(), rpc.btc_network) {
( BOTANIX_MAINNET_CHAIN_ID, bitcoin::Network::Bitcoin) => { ... }
(BOTANIX_TESTNET_CHAIN_ID, bitcoin::Network::Bitcoin) => { ,,, }
`

rwlockbg (dismissed):

crates/chainspec/src/spec.rs
scottmillner: im setting it to 20000 to match mainnet

bin/reth/src/commands/poa/mod.rs
scottmillner: made it easier to read, thanks for the suggestions!

bin/reth/src/commands/poa/mod.rs
scottmillner: i made this more concise based on @rwlockbg suggestions

rwlockbg (dismissed):

crates/chainspec/src/spec.rs
scottmillner: spoke with @PhoenixCodeCraft and we're going with 18 but can increase if need be
that's a 3 hour wait so it balances user experience with the likelihood of a reorg

arminsabouri: This is a protocol property not something that is configurable for UX. Treating it as such may have adverse consequences. Today a 18 block re-org is unimagninable. @PhoenixCodeCraft talks about building the financial platform for the next 100 years. Its not inconceivable that miner incentives / the whole mining industries changes after many halvings or bc of the introduction of newer protocol upgrades. Its the liquidity of the network that would be unstable if a deep-reorg happens and consequently user funds. The whole thing falls apart. While on-chain wallets would function fine. I voiced my concern. I would prefer to play this on the safer side especially considering users will swap not pegin.

Also keep in mind that increasing this value is a hardfork. In an emergency situation it may not be as easy to hardfork as we think it is.

rwlockbg (approved):

crates/chainspec/src/spec.rs
scottmillner: Thanks for voicing concerns.

Yes, I understand this would be a hardfork and it definitely might not be easy.
Agreed, the future for the miner industries/incentives could/will look very different.  But in that future I question that our bridging mechanism will even be used? I imagine a hardfork will be needed anyway to adjust to that future.

PhoenixCodeCraft: Would this be a soft fork using @lamafab activation logic? The non-changeable aspect of the pegin does present difficulties. 

The problem we're facing is that the theoretical required block depth is dependent on network usage. If there's almost no adoption (<100BTC bridged), then 3 blocks suffices. With gigantic adoption, this grows well beyond. 

Some data points on which a decision can be based: 
1) The tBTC bridge has been live for a few years, has around 350M USD (or 4k BTC) in it and has a 6 confirmation depth. No attack has occured. Ofcourse, this does not mean people haven't considered it but still an interesting actually existing data point in the market. 
2) The 100 confirmation depth from Liquid was decided in 2018. Today, the hashrate has grown around 20x from 2018, meaning if you consider the hashrate a direct comp of economic cost for a reorg (meaning the hashrate represents the block rewards), using the 2018 Liquid calculations would result in a 5 block confirmation depth. 
3) Antpool and stake.fish will be Botanix node operators and have around 30% bitcoin hash rate. 
4) The cost of a reorg is about 300k USD today (see point (2) as well since block reward and hashrate are fairly correlated), meaning trying a 18 block reorg would cost around 5M USD if not successful. To shed some light, that means for a 32.55% hash rate of foundry.usa and a corresponding likelihood of 1.42% this attack would have to yield around 380M USD to theoretically be profitable. The cost of the destruction of brand, legal entity etc should yield an additional factor to this calculation.  

If you take an additional factor on the calculation of (4), the conservative corresponding block depth should be something like:
-  6 from 0-1000BTC bridged. 
- 12 from 1000-10.000 bridged. 
- 18 from 10.000 to 100.000 bridged. 
etc etc

Meaning, the required depth is very much a function of the potential prize. Up until 10.000 BTC I believe the 18 block confirmation depth is a very conservative number. 

We don't have a swap provider ready to solve the current UX problems of long wait times. If a swap provider exists with sufficient liquidity that provides 1 block confirmation for users at a cost, there should be no reason to go beyond conservative to almost impossible with 100 block confirmation. 

Long story short, I believe at some point we'll have to change the confirmation depth. @lamafab had an interesting idea where we could already bake in before starting the genesis that first A botanix blocks confirmation depth is X, after A blocks this goes to Y, after B blocks, this goes to Z which basically bakes in the growth from the network. 

For now, I believe 18 is a very conservative number.

SatwikPrabhu (approved):

PR #662, opened 2025-04-14
Feature/pass block fee recipient address flag

scottmillner:
Instead of enforcing block builder fees go the cometbft address (which uses h160 and isn't the correct EVM address), allow node operators to pass the block fee recipient address on start up.


SatwikPrabhu (dismissed):
Code looks good to me

crates/cli/util/src/parsers.rs
```diff
@@ -154,4 +201,61 @@ mod tests {
...
+    }
+
+    #[test]
+    fn test_parse_ethereum_address_with_prefix() {
+        // A valid address provided with the "0x" prefix.
+        let address_str = "0x43C8bDCb9AFeBB1D834A7de18CC214a6FD1632d9";
+        let parsed = parse_ethereum_address(address_str).expect("Should parse valid address");
+
+        // The expected address is the same as if the prefix weren't there.
+        let trimmed = "0x43C8bDCb9AFeBB1D834A7de18CC214a6FD1632d9";
```
lamafab: This is not trimmed? 🤔

crates/consensus/authority/src/comet_bft/non_deterministic_data.rs
```diff
@@ -23,6 +24,7 @@ pub(crate) struct NonDeterministicData {
     pub(crate) version: u16,
     pub(crate) bitcoin_block_hash: bitcoin::hash_types::BlockHash,
     pub(crate) aggregated_public_key: secp256k1::PublicKey,
+    pub(crate) block_fee_recipient_address: Address,
```
lamafab: Shouldn't there be a version bump or at least backwards compatibility handling? Afaik this assumes all nodes update to the same version at the same time.

lamafab: Additionally, why have this in the NDD when it's already available in the EDH?

rwlockbg (dismissed):

crates/consensus/authority/src/comet_bft/non_deterministic_data.rs
scottmillner: ah yep nice catch, need versioning/backwards compatibility

this is how it gets in the EDH

the first tx in txs sent to comet is ndd which contains `block_fee_recipient`
then when comet gives the txs to reth node to build the block and append to chain, reth node will extract the block_fee_recipient and put in the EDH

crates/cli/util/src/parsers.rs
scottmillner: ah yep, just forget to remove `0x`, thanks!

crates/consensus/authority/src/comet_bft/non_deterministic_data.rs
scottmillner: added new v1 and have legacy support when historical syncing on testnet
mainnet will only use v1 for building and syncing

rwlockbg (dismissed):

lamafab (dismissed):

crates/cli/util/src/parsers.rs
```diff
@@ -154,4 +201,61 @@ mod tests {
 
         assert!(parse_duration_from_secs_or_ms("5ns").is_err());
     }
+
+    #[test]
+    fn test_parse_ethereum_address_no_prefix() {
+        // A valid address with 40 hex characters (without the "0x" prefix)
+        let address_str = "0x43C8bDCb9AFeBB1D834A7de18CC214a6FD1632d9";
```
lamafab: This **does** have the `0x` prefix :)

scottmillner: oh my, removed, thanks!

PR #663, opened 2025-04-14
DKG Redesign with Acknowledgements & Retry Timers and Authentication+Encryption

lamafab:
(_This description has been revised with Claude.AI_)

## DKG State Machine Redesign

This PR introduces a complete redesign of the DKG protocol implementation, replacing the previous round-based approach with a more robust state machine that handles the entire DKG protocol lifecycle with proper timeout handling and automatic message resending.

- Creates a dedicated `DkgStateMachine` component that handles the entire DKG protocol lifecycle
- Creates a dedicated `dkg::encryption` module that handles the message authentication and encryption
- Implements proper timeout handling and automatic resending of messages
- Adds comprehensive unit tests for each phase of the protocol
- Uses a more efficient message exchange pattern between participants

dkg_workflow

## RPC Interface Simplification

- **Removed RPC interfaces**:
	- `GetRound1DkgPackage`
	- `GetRound1DkgPackages`
	- `NewRound1DkgPackage`
	- `GetRound2DkgPackage`
	- `NewRound2DkgPackage`

- **Added RPC interfaces**:
	- `GetDkgPayloads`
	- `NewDkgPayload`

## Protocol Structure Changes

- Consolidated multiple payload variants into a single DKG payload type with routing information. This change is also in preparation for the future encryption mechanism and TEE implementation.
- Added a new `DkgPayloads` structure that includes a `timeout` field and a list of `payloads`
- Implements a coordinator-based approach where one node orchestrates the DKG process.
- Adds support for a multi-phase protocol with proper acknowledgments and retry mechanisms.

## Encryption & Authentication

Added a three-round handshake protocol with proper authentication using static and ephemeral keys:
   - Round 1: Initial authenticated commitments with ephemeral keys
   - Round 2: Encrypted package exchange with derived symmetric keys
      - Uses ChaCha20Poly1305 for symmetric encryption
   - Round 3: Verifiable distributed key aggregation
## POA Node Integration

- On the POA node side, implemented a simple `DkgRunnerTask` that interfaces with the `btc-server`.
- The `btc-server` now dictates when the next RPC call should happen by returning a timeout value.
- This timeout-based approach enables handling of retransmissions without constant polling.

## Configuration Updates

- Added command-line parameters for coordinator and federation members configuration

## Test Suite

The test-suite has been adjusted to reflect those changes.


scottmillner:
@birozuru 
this pr adds two more flags for `btc-server`:
- these flags already exist for reth node but needed in `btc-server` now
- when we release this next monday, we can hop on a call but just wanted to give heads up

```
--federation-config-path "${NODE_1_DIR}/federation.toml" \
--p2p-secret-key "${NODE_1_DIR}/discovery-secret" \
```

arminsabouri:
> @birozuru this pr adds two more flags for `btc-server`:
> 
> * these flags already exist for reth node but needed in `btc-server` now
> * when we release this next monday, we can hop on a call but just wanted to give heads up
> 
> ```
> --federation-config-path "${NODE_1_DIR}/federation.toml" \
> --p2p-secret-key "${NODE_1_DIR}/discovery-secret" \
> ```

@lamafab Why does btc server need access to the reth p2p secret. This doesnt seem right

bin/btc-server/src/bin/main.rs
```diff
@@ -295,12 +298,34 @@ where
...
+        // Prepare the federation config.
+        // TODO: Handle error
+        let raw = std::fs::read_to_string(&config.federation_config_path).unwrap();
+        let federation = FederationTomlConfig::from_str(&raw).unwrap();
+
+        // Prepare our secret key.
+        let raw = std::fs::read_to_string(&config.p2p_secret_key).unwrap();
+        let sanitzed_key = raw.chars().filter(|c| c.is_ascii_hexdigit()).collect::<String>();
+        // TODO: Handle error
+        let secret_key = sanitzed_key.as_str().parse::<secp256k1::SecretKey>().unwrap();
```
arminsabouri: Typo sanitized

arminsabouri: Also reminder to fix unwrap

lamafab:
> @lamafab Why does btc server need access to the reth p2p secret. This doesnt seem right

We want to make sure that the Dkg payloads that we receive are from the correct federation member. The `--p2p-secret-key` is used to sign the payloads, while the corresponding public key in the federation Toml is used to verify the signature. This PR just reuses the two structures, we technically do not need the other entries in the federation Toml.

bin/btc-server/src/dkg/mod.rs
```diff
@@ -0,0 +1,1489 @@
...
+pub enum Stage {
+    /// Active exchange of round1 packages between participants.
+    RoundOneActive,
+    /// Active exchange of round2 packages between participants.
+    RoundTwoActive,
+    /// Active exchange of round3 packages (aggregated public keys).
+    RoundThreeActive,
+    /// The DKG process was aborted.
+    Aborted,
+    /// DKG protocol finalizxed successfully.
```
scottmillner: nit: typo `finalizxed` > `finalized`

bin/btc-server/src/dkg/mod.rs
```diff
@@ -0,0 +1,1489 @@
...
+
+// TODO: Hide behind `#[cfg(test)]`?
+impl From<frost::Identifier> for Target {
+    fn from(id: frost::Identifier) -> Self {
+        Target(id)
+    }
+}
+
+mod sealed_round1_pkg {
+    use super::*;
```
scottmillner, line 43: nit: mod `sealed_round1_pkg`contains `SealedRoundThreeSignature` and impl which confused me for a second

lamafab: oh right, forgot to rename that one :) will update.

bin/btc-server/src/dkg/mod.rs
```diff
@@ -0,0 +1,1489 @@
...
+    /// participant identifiers and configuration parameters.
+    ///
+    /// When created, the state machine behavior differs based on whether this participant
+    /// is the coordinator:
+    ///
+    /// - If this participant is the coordinator (`my_frost_id == coordinator`), the state machine
+    ///   immediately enters the `RoundOneActive` stage and generates round1 packages for all other
+    ///   participants. These initial packages can be retrieved by calling `send()`.
+    ///
+    /// - If this participant is not the coordinator, the state machine enters the `RoundOnePending`
```
lamafab: Just a note; the `RoundOnePending` variant is deprecated. We now just use the `pending: bool` field. Will update.

bin/btc-server/src/dkg/mod.rs
```diff
@@ -0,0 +1,1489 @@
...
+            // intended recipient.
+            if !self_is_coordinator {
+                // We're not the coordinator; just drop the package.
+                return Ok(());
+            }
+
+            debug_assert!(!*pending);
+
+            // Note that we do not explicitly enforce that the package is
+            // opaque; we allow forwarding unencrypted packages.
```
lamafab: Deprecated comment as well.

scottmillner:
Went through all the changes (mainly for my own understanding than to give constructive feedback)
- awesome work!!
- great tests! makes it easy to reason about the dkg flow and encryption: again, very nice!
- all unit tests passed locally
- all integration tests pass locally

the clippy job is failing with a few issues:
- do you mind cleaning those up and then ✅ 

excited to see this in action next week!!

bin/btc-server/src/dkg/encryption.rs
```diff
@@ -0,0 +1,603 @@
...
+            my_frost_id,
+            my_static_sec,
+            my_eph_sec,
+            my_eph_pub,
+            fed_members,
+            eph_keys,
+            round1_commits: BTreeMap::new(),
+        };
+
+        Ok(this)
```
arminsabouri: Unneccecary placeholder variable `this`.

bin/btc-server/src/dkg/encryption.rs
```diff
@@ -0,0 +1,603 @@
...
+}
+
+/// Authentication and encryption layer for round one of the DKG protocol.
+///
+/// Handles the authentication of round one packages using ephemeral keys and
+/// signatures, establishing a secure communication channel between
+/// participants.
+#[derive(Clone)]
+pub struct RoundOneAuth {
+    bytes: [u8; 64],
```
arminsabouri: This could use a more decriptive name.

bin/btc-server/src/dkg/encryption.rs
```diff
@@ -0,0 +1,603 @@
...
+        // ephemeral key and the round1 package.
+        let mut t = self.transcript.clone();
+        t.append_message(b"frost_id", self.my_frost_id.serialize().as_slice());
+        t.append_message(b"eph_pub", my_eph_pub.serialize().as_slice());
+        t.append_message(b"round1_package", package.serialize()?.as_slice());
+        t.challenge_bytes(b"round1_commit", commit);
+        std::mem::drop(t);
+
+        // Create the signature
+        let msg = secp256k1::Message::from_digest_slice(&commit[..32]).expect("valid size");
```
arminsabouri: Why slice the `bytes`?

bin/btc-server/src/dkg/encryption.rs
```diff
@@ -0,0 +1,603 @@
...
+        let signature = secp256k1::Secp256k1::new().sign_ecdsa(&msg, &self.my_static_sec);
```
arminsabouri: Use the global SECP context when you can. Recreating the context is costly

bin/btc-server/src/dkg/encryption.rs
```diff
@@ -0,0 +1,603 @@
...
+            let mut t = self.transcript.clone();
+            t.append_message(b"ss1", ss1.secret_bytes().as_slice());
+            t.append_message(b"ss2", ss2.secret_bytes().as_slice());
+            t.challenge_bytes(b"sym_key1", key1.as_mut_slice());
+            t.challenge_bytes(b"sym_key2", key2.as_mut_slice());
+            std::mem::drop(t);
+
+            // The participant with the lower Frost ID uses the first key as the
+            // sending key and the second key as the receiving key. The other
+            // participant does the opposite.
```
arminsabouri, line 312: I dont understand why we need two shared secrets?

bin/btc-server/src/dkg/encryption.rs
```diff
@@ -0,0 +1,603 @@
...
+
+            // Generate the symmetric keys using the shared secrets.
+            let mut key1 = Zeroizing::new([0; 64]);
+            let mut key2 = Zeroizing::new([0; 64]);
+
+            let mut t = self.transcript.clone();
+            t.append_message(b"ss1", ss1.secret_bytes().as_slice());
+            t.append_message(b"ss2", ss2.secret_bytes().as_slice());
+            t.challenge_bytes(b"sym_key1", key1.as_mut_slice());
+            t.challenge_bytes(b"sym_key2", key2.as_mut_slice());
```
arminsabouri, line 307: There might be something going on here that I don't understand. WE are Zeroizing two keys and then submitting it to the transcript?

bin/btc-server/src/dkg/encryption.rs
```diff
@@ -0,0 +1,603 @@
...
+            debug_assert_ne!(&self.my_frost_id, fed_id);
+            if &self.my_frost_id < fed_id {
+                ss1 = secp256k1::ecdh::SharedSecret::new(&fed_eph, &self.my_static_sec);
+                ss2 = secp256k1::ecdh::SharedSecret::new(&fed_static, &self.my_eph_sec);
+            } else {
+                ss1 = secp256k1::ecdh::SharedSecret::new(&fed_static, &self.my_eph_sec);
+                ss2 = secp256k1::ecdh::SharedSecret::new(&fed_eph, &self.my_static_sec);
+            }
+
+            // Generate the symmetric keys using the shared secrets.
```
arminsabouri: **Raw DH Outputs Are Not Directly Suitable for Symmetric Keys** 

You need to use a KDF or XOF. You cannot use raw points directly. They need to be uniform over a key space.

bin/btc-server/src/dkg/encryption.rs
```diff
@@ -0,0 +1,603 @@
...
+        &mut self,
+        target: &Target,
+        package: &round2::Package,
+    ) -> Result<(Vec<u8>, u64), Error> {
+        // Retrieve the symmetric key for the target.
+        let entry = self.symmetric_keys.get_mut(&target.0).ok_or(Error::NotAFedMember)?;
+
+        let sending_nonce = entry.nonce;
+        let mut s = [0; 12];
+        s[..8].copy_from_slice(&sending_nonce.to_le_bytes());
```
arminsabouri: Why slice?

bin/btc-server/src/dkg/encryption.rs
```diff
@@ -0,0 +1,603 @@
...
+    ) -> Result<round2::Package, Error> {
+        // Prepare the encryption key.
+        let entry = self.symmetric_keys.get(&initiator.0).ok_or(Error::NotAFedMember)?;
+
+        let receiving_key = entry.receiving.as_slice();
+        let cipher = ChaCha20Poly1305::new_from_slice(receiving_key.as_ref()).expect("valid size");
+
+        let mut s = [0; 12];
+        s[..8].copy_from_slice(&nonce.to_le_bytes());
+        let nonce = Nonce::from_slice(&s);
```
arminsabouri: Since this is common operation perhaps we wrap this a nonce struct? And impl Into the libraries None

bin/btc-server/src/dkg/encryption.rs
```diff
@@ -0,0 +1,603 @@
...
+    ///
+    /// # Returns
+    ///
+    /// A signature on the public key package for verification by other
+    /// participants
+    pub fn commit_round3(
+        &mut self,
+        package: &frost::keys::PublicKeyPackage,
+    ) -> Result<secp256k1::ecdsa::Signature, Error> {
+        let commit = &mut self.bytes;
```
arminsabouri: This is intialized to empty bytes above. This should be a commitment to the entire transcript so far (or in our bastardized version of chillDKG a commitment to the agg pk) Does this get updated somewhere?

bin/btc-server/src/dkg/encryption.rs
arminsabouri: Why ecdsa over schnorr?

bin/btc-server/src/dkg/encryption.rs
```diff
@@ -0,0 +1,603 @@
...
+                ss1 = secp256k1::ecdh::SharedSecret::new(&fed_eph, &self.my_static_sec);
+                ss2 = secp256k1::ecdh::SharedSecret::new(&fed_static, &self.my_eph_sec);
+            } else {
+                ss1 = secp256k1::ecdh::SharedSecret::new(&fed_static, &self.my_eph_sec);
+                ss2 = secp256k1::ecdh::SharedSecret::new(&fed_eph, &self.my_static_sec);
+            }
+
+            // Generate the symmetric keys using the shared secrets.
+            let mut key1 = Zeroizing::new([0; 64]);
+            let mut key2 = Zeroizing::new([0; 64]);
```
arminsabouri: it seems like we only use 32 bytes but initialize 64 bytes? Do we ever use the other 32 bytes somewhere?

bin/btc-server/src/dkg/encryption.rs
```diff
@@ -0,0 +1,603 @@
...
+    pub fn commit_round3(
+        &mut self,
+        package: &frost::keys::PublicKeyPackage,
+    ) -> Result<secp256k1::ecdsa::Signature, Error> {
+        let commit = &mut self.bytes;
+
+        self.transcript.append_message(b"round3_package", package.serialize()?.as_slice());
+        self.transcript.challenge_bytes(b"round3_commit", commit);
+
+        let challenge: [u8; 32] = commit[..32].try_into().expect("valid size");
```
arminsabouri: Why does commit need to be > 32 bytes if we only use the last 32 bytes?

arminsabouri (changes requested):
I took a look at encryption.rs and had a couple comments.
Main points were:
* Use of plain ec points for symetric key material 
* re-creating secp contexts where we could be using the global context
* Seems like there are alot of random slicing going on. Need some clarification on that 

But those comments are besides the point. My concern and where I would approach NACK this is: this work conflates the reth p2p key with the FROST dkg protocol -- which has its own keys suitable for our r2 and r3 authentication purposes.

Reth already has a ECIES implementation. Meaning messages to and from fed members are authenticated and encrypted. And you also are also able to authenticate if you are communicating with a real fed members by virtue of the fed.toml. 

Our FROST DKG protocol does not need access to these keys as they already generating session keys for the key generation ceremony. This is the whole premise of chill DKG. Its a all-batteries-included protocol where you don't need access to external keys.

Public commitments (G*a_0) are shared in round1 along side a proof of ownership.  By the time peers initiate round 2 they already have n-1 public commitments which can be used to NI-ECDH and use the result in a authenticated encryption scheme (poly-chacha) like you have done. Similarly for round3, peers will sign the result of the dkg session using their partial sk -- which anyone can verify as they took part in round 1. In essence, for round2 we are wrapping an encrypted message using hpke in an existing hpke protocol using different keys. Whereas, in this impl we are wrapping encrypted messages using the same set of keys. This creates a strong coupling that I can't seem to justify. Especially since the alternative is using the existing DKG keys which are already accessible.

bin/btc-server/src/dkg/encryption.rs
lamafab: The challenge bytes generated by `merlin` are 64-bytes, meanwhile the `secp256k1::Message` wants a 32-byte slice.

bin/btc-server/src/dkg/encryption.rs
lamafab: Since we use ephemeral keys - for forward secrecy - we want to use the ephemeral key material to generate the symmetric keys. We could technically do a single `DH(eph, static)` operation, but doing it with all keys is common practice.

See for example: https://noiseprotocol.org/noise.html#pattern-modifiers

We're closest to this setup, which even does four DH operation in total:

```
KK:
  -> s
  <- s
  ...
  -> e, es, ss
  <- e, ee, se
```

In our code, we only do `es` and `se`, since both parties already have the ephemeral key (`e`) before creating the handshake (distributed and authenticated in round1, before encryption is required).

bin/btc-server/src/dkg/encryption.rs
lamafab: No, `challenge_bytes` writes bytes **to** the given variable.

This just generates two reproducible challenges - in this context, two symmetric keys - where it takes the transcript state with an additional message (`sym_key1` and `sym_key2`) and KDF'es the output to `key1` and `key2`, respectively. Both parties can only reproduce this state if they can compute the shared secrets.

bin/btc-server/src/dkg/encryption.rs
lamafab: > You need to use a KDF or XOF.

Yes, that's what `challenge_bytes` does.

bin/btc-server/src/dkg/encryption.rs
lamafab: The ChaCha20 cipher wants a 12-byte nonce, meanwhile `u64` only has 8 bytes.

bin/btc-server/src/dkg/encryption.rs
lamafab: It's just a placeholder. I initially created this inner `bytes` field to avoid new allocations, so it can be reused by future calls - but it's essentially useless. I will remove this since it might lead to confusion.

bin/btc-server/src/dkg/encryption.rs
lamafab: It's more common and we don't specifically need Schnorr ¯\_(ツ)_/¯ I guess we could switch to it, up for debate. I don't have a strong opinion here.

lamafab:
@0xBEEFCAF3 

> Reth already has a ECIES implementation. Meaning messages to and from fed members are authenticated and encrypted. And you also are also able to authenticate if you are communicating with a real fed members by virtue of the fed.toml.

The main issue is the coordinator, **who forwards messages on behalf of someone else**. While the individual endpoints are encrypted and authenticated, the coordinator can by design manipulate the packages. This is a typical MITM-attack problem. So simply relying on the transport layer is not sufficient; the Frost packages - **or at least the first** round1 - MUST be authenticated with the P2P key individually. Without that P2P key, there is nothing tying that round1 package to the intended federation member.

Additionally, this allows us to complete the DKG process via gossip, which might or might not be useful for DynaFed (however that will turn out).

> Similarly for round3, peers will sign the result of the dkg session using their partial sk -- which anyone can verify as they took part in round 1.

Sure, for round3 we could technically skip the signature generation using the P2P key, and just rely on a partial key for that purpose. But we do need the P2P key for the round1 signature.

And on this point:

> Seems like there are alot of random slicing going on. Need some clarification on that

~The slicing is there because `merlin` generates 64-bytes challenges (KDF outputs) with `challenge_bytes`~, meanwhile both `secp256k1::Message` (for signature creation) and `ChaCha20Poly1305::new_from_slice` (for encryption) want 32-byte slices. We could use another KDF here that converts that 64-byte challenge into 32 bytes, but since that challenge already is a KDF output, slicing is fine here.

EDIT: Actually, `challenge_bytes` **can** write to a 32-byte slice directly 🤔 not sure why I thought it has to be 64-bytes. Will fix this!

bin/btc-server/src/dkg/encryption.rs
lamafab: Updating this to generate 32-byte keys directly.

bin/btc-server/src/dkg/encryption.rs
lamafab: Updating this to generate 32-byte keys directly.

bin/btc-server/src/dkg/encryption.rs
lamafab: The `bitcoin` module does not enable the `"global-context"` feature for `secp256k1`, so now I'm just initializing this context once and saving it internally. Fixed.

bin/btc-server/src/dkg/encryption.rs
lamafab: Done.

lamafab:
## Changes

- Renamed `RoundOneAuth` to `DkgHandshakeManager`
- Renamed `RoundTwoAuth` to `SecureChannelManager`
- Renamed `RoundThreeAuth` to `KeyVerificationManager`
- Renamed module `sealed_round1_pkg` to `sealed_pkg` to better reflect its purpose
- Created helper function `integer_to_serialized_nonce()` to eliminate duplicate code
- Removed mutable `bytes` field arrays from auth types, using local variables instead
- Cached `secp256k1::Secp256k1` context in each manager to avoid repeated creation
- Removed slicing by using properly sized buffers
- Added proper error handling in main.rs (replaced unwrap calls with proper error propagation)
- Other tiny cleanups such as typos or deprecated docs
- Executed `make fmt`

bin/btc-server/src/dkg/encryption.rs
arminsabouri: Its more common but for all the wrong reasons. Schnorr is always prefered

arminsabouri:
> @0xBEEFCAF3
> 
> > Reth already has a ECIES implementation. Meaning messages to and from fed members are authenticated and encrypted. And you also are also able to authenticate if you are communicating with a real fed members by virtue of the fed.toml.
> 
> The main issue is the coordinator, **who forwards messages on behalf of someone else**. While the individual endpoints are encrypted and authenticated, the coordinator can by design manipulate the packages. This is a typical MITM-attack problem. So simply relying on the transport layer is not sufficient; the Frost packages - **or at least the first** round1 - MUST be authenticated with the P2P key individually. Without that P2P key, there is nothing tying that round1 package to the intended federation member.

Isn't that the whole point of the transcript in this PR? If a coord manipulates, drops or equivocates any message in R1, its detectable in the final rounds. This is also the design of chillDKG. IIRC the coord is only semi trusted bc there is no way to detect sybils not because they can pull of MITM attacks. In order to prevent sybils we need some identity registration. A way to tie fed.toml long term identities to the round 1 public commitments. It could be simple. Round 1 messages could be the package of all R1's from all peers along side their Frost Ids (which for simplicity could just be a hex encoding of their federation public keys -- this would solve the federation operator UX problem cc @birozuru ) and a signature over their round 1 commitments, their frost id, their fed pk (we dont need this if the frost id is the fed pk obv), and any other metadata like session ids if such a thing exists in this design). This logic can be handled entirely in the FROST task. This way we seperate application concerns and more importantly keys.

arminsabouri:
I have many files to still review. In the mean time there are some CI failures as well as conflicts that could use some resolving. 

In the future please try to make each commit atomic and small. 
For example fmt changes or compile errors from bad rebases should just be applied in the commit history not a seperate commit. This speeds up the review process and allows us to rebase and merge -- for this PR I would just squash and merge commit at this point.

arminsabouri:
One more thing, I think this branch needs a rebase. Seeing some changes from what I think is from main. e.g `bin/test-suite/multi_mint_helper_abi.json`

bin/btc-server/proto/btc_server.proto
```diff
@@ -188,9 +182,15 @@ message GetTrackedTxsResponse {
 message Empty {}
 
 // Frost things
+message DkgPayloads {
+    uint64 timeout = 1;
```
rwlockbg, line 202: Do we actually need to have the timeout in the payload - is not this something that can be set across all nodes to be equal via an argument ? It looks very weird to have a timeout param in the proto payload ...

bin/btc-server/src/dkg/encryption.rs
```diff
@@ -0,0 +1,589 @@
...
+        let my_eph_pub = self.my_eph_pub.clone();
+
+        // Compute the challenge bytes to be signed; notably, we commit the
+        // ephemeral key and the round1 package.
+        let mut t = self.transcript.clone();
+        t.append_message(b"frost_id", self.my_frost_id.serialize().as_slice());
+        t.append_message(b"eph_pub", my_eph_pub.serialize().as_slice());
+        t.append_message(b"round1_package", package.serialize()?.as_slice());
+        t.challenge_bytes(b"round1_commit", &mut commit);
+        std::mem::drop(t);
```
rwlockbg, line 177: is there no Drop implemented for the transcript, why is manual dropping necessary here ?

bin/btc-server/src/dkg/mod.rs
```diff
@@ -0,0 +1,1488 @@
...
+// TODO: Hide behind `#[cfg(test)]`?
+impl From<frost::Identifier> for Initiator {
+    fn from(id: frost::Identifier) -> Self {
+        Initiator(id)
+    }
+}
+
+/// Wrapper type for FROST identifiers used in the DKG protocol.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, PartialOrd, Ord, Serialize, Deserialize)]
+pub struct Target(frost::Identifier);
```
rwlockbg, line 33: Maybe rename it to FrostTarget or MessageTarget ? Target sounds a bit unclear to me ...

bin/btc-server/src/dkg/mod.rs
```diff
@@ -0,0 +1,1488 @@
...
+    ///
+    /// A new `DkgStateMachine` instance
+    pub fn new(
+        my_frost_id: frost::Identifier,
+        my_static_sec: secp256k1::SecretKey,
+        coordinator: frost::Identifier,
+        members: BTreeMap<frost::Identifier, secp256k1::PublicKey>,
+        config: Config,
+    ) -> Result<Self, Error> {
+        if !members.contains_key(&my_frost_id) {
```
rwlockbg, line 399: Just wondering if it would not make sense to actually add more info to these err messages. e.g. min_signers X cannot be greater than max_signers Y for a quick scan and seeing the problem if it comes to one ?

bin/btc-server/src/dkg/mod.rs
```diff
@@ -0,0 +1,1488 @@
...
+    /// # Returns
+    ///
+    /// An optional `Duration` until the next timeout event. If `None`, there
+    /// are no pending timeout events.
+    pub fn timeout(&self, now: Instant) -> Option<Duration> {
+        match &self.state {
+            StageState::RoundOneActive { out_round1_packages, .. } => out_round1_packages
+                .values()
+                .filter(|e| e.is_some())
+                .filter_map(|e| e.as_ref().expect("must be available").timer)
```
rwlockbg, line 569: Cannot we do both ops in filter_map only ?

bin/test-suite/build.rs
```diff
@@ -18,13 +18,10 @@ fn main() {
...
-    Abigen::new(
-        "MultiMintHelperContract",
-        "multi_mint_helper_abi.json",
-    )
-    .expect("Error reading multi mint helper contract json abi")
-    .generate()
-    .expect("Error generating multi mint helper contract rust definitions")
-    .write_to_file("./src/multi_mint_helper_abi.rs")
-    .expect("Error writing multi mint helper contract rust file");
+    Abigen::new("MultiMintHelperContract", "multi_mint_helper_abi.json")
```
rwlockbg: These 2 files are not part of this PR, right ? Maybe we need to discard them I guess ...

rwlockbg:
@lamafab Might need to resolve several conflicts in the PR btw.

rwlockbg:
@lamafab Very well written code and tests, it all makes total sense to me. I left a few small comments but overall it is indeed very well done and ready to go! Thanks for the hard work in here.

bin/btc-server/proto/btc_server.proto
lamafab: This `timeout` is primarily used by the caller of the `new_dkg_payload` (btc-server) API. This message does not get gossiped directly to the network, but rather the nested individual payloads inside the array.

See the [`frost_task.rs`](https://github.com/botanix-labs/Macbeth/blob/e4247cb7193531970f74b0e940722b0dce474a8c/crates/consensus/authority/src/frost_task.rs#L802C48-L802C63) as an example:

```rust
let resp = self.btc_server.new_dkg_payload(req).await.unwrap();

if let Ok(resp) = self.btc_server.get_public_key(client::Empty {}).await {
	self.metrics.created_agg_pub_keys.increment(1);

	// decode the public key and assign it to the self variable
	let public_key_package =
		secp256k1::PublicKey::from_str(&resp.publickey).unwrap();
	storage.aggregate_public_key = Some(public_key_package);
}

// Update timeout at which point the btc-server should be called again.
timeout = Duration::from_millis(resp.timeout);

self.gossip_payloads(resp.payloads).await;
```

bin/btc-server/src/dkg/encryption.rs
lamafab: The `std::mem::drop()` function here serves purely for clarification purposes; the transcript is only used temporarily to create a commit - usually to be signed/verified - and that this short term `t` transcript is not appended to the internal `self.transcript`.

bin/btc-server/src/dkg/mod.rs
lamafab: Yes, I could do `MessageTarget` (or `MsgTarget`)

bin/btc-server/src/dkg/mod.rs
lamafab: I remember trying this, but there was some _flattening_ issue, I believe. Will revisit this again.

bin/test-suite/build.rs
lamafab: Yeah, the linting and formatting are a constant pain...

scottmillner (dismissed):
per offline discussion:
- @0xBEEFCAF3 recommendations (btc server not using reth p2p secret) will be implemented in a separate pr

bin/btc-server/proto/btc_server.proto
rwlockbg: all good

bin/test-suite/build.rs
rwlockbg: no worries, I get them too. Resolving it for now.

rwlockbg (dismissed):
@lamafab @scottmillner Approved, good to go, nice piece of work. It just needs rebasing!

bin/btc-server/src/dkg/mod.rs
rwlockbg: no worries, not important.

bin/btc-server/src/dkg/mod.rs
```diff
@@ -0,0 +1,1488 @@
...
+        else {
+            // On error, we abort the entire DKG process.
+            //
+            // TODO (lamafab): We should probably do some extra things here,
+            // such as communicating this information to the coordinator/peers.
+            //
+            // TODO (lamafab): Explicitly test this condition.
+            self.state = StageState::Aborted;
+            return Ok(());
+        };
```
SatwikPrabhu, line 1242: Do we expect this TODO to be critical? Since I think lack of abort communication could lead to other nodes stalling indefinitely.

bin/btc-server/src/dkg/mod.rs
```diff
@@ -0,0 +1,1488 @@
...
```
SatwikPrabhu, line 1368: Same as https://github.com/botanix-labs/Macbeth/pull/663/files#r2074091762

crates/consensus/authority/src/frost_task.rs
```diff
@@ -802,3 +739,141 @@ where
...
+                Ok(Some(dkg)) => {
+                    let req = client::DkgPayload {
+                        sender: dkg.sender,
+                        // TODO: comment
+                        recipient: dkg.recipient,
+                        payload: dkg.data,
+                    };
+
+                    let resp = self.btc_server.new_dkg_payload(req).await.unwrap();
+
```
SatwikPrabhu, line 711: I see a bunch of unwraps in this file. These might be relatively a non-issue, but maybe its better to gracefully handle errors instead of panicking?

SatwikPrabhu (dismissed):
Great work @lamafab 🔥 Enjoyed going through this thoroughly. Sorry for the delay in my review though. I've left a couple of questions which I think can be addressed later after this is merged. For now I think this looks good. Awesome!

crates/consensus/authority/src/frost_task.rs
lamafab: Yes, you're right. Will follow up on that.

PR #664, opened 2025-04-15
Fix(audit): Revert tx state changes that has invalid pegins/pegouts

SatwikPrabhu:
Fix for https://github.com/botanix-labs/botanix/issues/980


SatwikPrabhu:
The int tests and unit tests are passing locally for me. Currently considering adding another int test to test the case of a block with two transactions: one with valid pegins/pegouts and one with an invalid pegin/pegouts.
@scottmillner Let me know if you can think of any other test scenarios

bin/test-suite/src/suite/consensus/invalid_transactions/test_invalid_pegout.rs
```diff
@@ -64,7 +64,7 @@ pub async fn invalid_pegout(
     // use empty pegout data
     let pegout_data = ethers::core::types::Bytes::new();
     let pegout_amount = Amount::from_btc(0.5).unwrap();
-    let actual_pegout_amount = (pegout_amount / 2).to_wei();
+    // let actual_pegout_amount = (pegout_amount / 2).to_wei();
```
arminsabouri: If its not used and not important to leave as comment, lets remove

crates/ethereum/evm/src/execute.rs
```diff
@@ -263,6 +261,15 @@ where
 
             self.evm_config.fill_tx_env(evm.tx_mut(), transaction, *sender);
 
+            // Store original sender account info before transaction execution.
+            // This is used later to partially revert the state if botanix specific validation
+            // fails.
+            let mut original_sender_info = match evm.db_mut().basic(*sender) {
+                Ok(Some(info)) => info,
+                Ok(None) => AccountInfo::default(),
+                Err(_) => panic!("Error getting sender account"),
```
arminsabouri: A panic raises some eyebrows. Under what conditions would this error occur? 
If it ephemeral it should retry and and err should be returned. 
If it irrecoverable it still seems to be that an err should be returned.

crates/ethereum/evm/src/execute.rs
```diff
@@ -302,45 +309,42 @@ where
                         Ok((new_pegins, new_pegouts)) => {
                             pegins.extend(new_pegins);
                             pegouts.extend(new_pegouts);
+                            drop(original_sender_info);
```
arminsabouri: why drop here if this eventually falls out of scope

crates/ethereum/evm/src/execute.rs
```diff
@@ -302,45 +309,42 @@ where
...
+                            original_sender_info.balance = post_tx_sender_info.info.balance;
+
+                            // Clear all other state changes introduced by the transaction.
+                            state.clear();
+
+                            // Re-insert the sender's account with the original info but updated
+                            // nonce and balance.
+                            let reverted_account = Account {
+                                info: original_sender_info,
+                                storage: HashMap::new(), // No storage changes.
```
scottmillner: just confirming that state is just the diff, yeah?

as in wiping the storage here only clears the diff and not all the Account's storage?

scottmillner:
looking good! 
utAck

scottmillner:
> The int tests and unit tests are passing locally for me. Currently considering adding another int test to test the case of a block with two transactions: one with valid pegins/pegouts and one with an invalid pegin/pegouts. @scottmillner Let me know if you can think of any other test scenarios

i think we should add test cases for all the audit findings
you can do that in this pr or a separate one, whichever you prefer

rwlockbg (dismissed):

bin/test-suite/src/suite/consensus/invalid_transactions/test_invalid_pegin.rs
```diff
@@ -69,23 +69,24 @@ async fn generate_invalid_pegin_metas(
...
+    //         )
+    //         .expect("valid public key"),
+    //         tx: pegin_tx.clone(),
+    //         merkle_proof: pmt.clone(),
+    //         block_headers: headers.clone(),
+    //     },
+    //     ref_block_hash: B256::from_slice(&[0; 32]),
+    // })];
+    // invalid_pegin_meta_cases.push((invalid_ref_hash_meta, "Invalid reference hash for v1"));
+    // TODO(satwik): The above test case is not passing anymore. Need to debug.
```
arminsabouri: Can you explain why this isn't passing anymore? Are we not getting a invalid refhash error anymore or a different error. This would be useful context for the TODO comment.

crates/ethereum/evm/src/execute.rs
arminsabouri: If the orignial sender has state wouldn't this wipe it out?

crates/ethereum/evm/src/execute.rs
```diff
@@ -302,45 +309,42 @@ where
...
+                            // Crucially, retain the sender's nonce and balance from after the EVM
+                            // execution success. This ensures the sender pays gas
+                            // and their nonce increments, even though the core logic of the
+                            // transaction (e.g., minting/burning) is effectively undone.
+                            let post_tx_sender_info =
+                                state.get(sender).expect("Sender account to exist").clone();
+
+                            original_sender_info.nonce = post_tx_sender_info.info.nonce;
+                            original_sender_info.balance = post_tx_sender_info.info.balance;
+
```
arminsabouri: Just to clarify the orignial sender is who exactly? The entity that called the contract or the recepiented of the mint? i think it should be the recepient of the mint. and it that case do we need to change their nonce?

arminsabouri (commented):
I'm glad this was possible. i mostly have questions. Not really reqs for change. The one error assert that is being commented out is concerning. 

Great finds and over all work @SatwikPrabhu

crates/ethereum/evm/src/execute.rs
SatwikPrabhu: Yep. The `info` field is represents what the updated state would contain and the `storage` has the diffs (not the account's storage) indeed.

SatwikPrabhu:
> i think we should add test cases for all the audit findings
> you can do that in this pr or a separate one, whichever you prefer

I have already started working on the int tests, which I was planning to push later today. That said, I think it might make sense to push this first and the int tests in another PR to keep things moving.

crates/ethereum/evm/src/execute.rs
SatwikPrabhu: Yeah. I was unsure as well as to what to do here. That's because an error here means a `BlockExecutionError` and that means the block is rejected. That said, it might make sense to have a retry mechanism here. 
This is an error when there is no account info for a sender address whose transaction was included in a block. I think a panic is warranted here?

crates/ethereum/evm/src/execute.rs
SatwikPrabhu: Good point! Will update.

crates/ethereum/evm/src/execute.rs
SatwikPrabhu: This is just the diff in the storage

bin/test-suite/src/suite/consensus/invalid_transactions/test_invalid_pegin.rs
SatwikPrabhu: The all zeros hash is apparently valid now and hence the pegin is not invalid anymore. Havent identified the cause yet. Will update the comment for better info

crates/ethereum/evm/src/execute.rs
SatwikPrabhu: Yes. It is the recipient of the mint. I think we need to change the nonce because otherwise would run into the transaction replay scenario. Moreover, as highlighted before, this is the revert behavior in the evm as well right. I dont see why we should deviate from it.

scottmillner: origininal sender is **NOT** the recipient nor is there a concept of recipient here

if we think this through: who sends the pegin currently from the dapp? the botanix hot wallet sends pegins on behalf of users...this is just like any external tx on ethereum...any account can "broadcast" a tx as long as they pay for it

origininal sender is `tx.origin` which is not necessarily the caller of the contract (this is what the mint attack contract does)

`tx.origin` is the account that initiated the external tx

otherwise @SatwikPrabhu comment is accurate:

normal evm revert behavior (which we need to keep intact to not deviate from the protocol):
- storage state remains the same
- sender (who initiated the external tx) nonce increments
- sender pays gas
- events/logs are not included in the reciept

this is just like if you had a tx reverted on ethereum and our behavior should be the same

bin/test-suite/src/suite/consensus/invalid_transactions/test_invalid_pegout.rs
SatwikPrabhu: Done. The test has also been updated according to the revert logic

crates/ethereum/evm/src/execute.rs
SatwikPrabhu: Added a revert logic and returning an error now

crates/ethereum/evm/src/execute.rs

bin/test-suite/src/suite/consensus/invalid_transactions/test_invalid_pegin.rs
SatwikPrabhu: Updated the todo. Will open a follow up issue to fix it

SatwikPrabhu: Ok. I did some digging up on this and opened a follow up issue for this: https://github.com/botanix-labs/botanix/issues/996

Here's the cause on why the test was passing before this PR and not now:
In `botanix_mint_contract_checks`, we were not returning an error when we failed to find a block corresponding to an invalid hash. In this case, we just used the default bitcoin checkpoint that we would have used in pegin v0. But this does match with the block headers list and we would eventually hit https://github.com/botanix-labs/Macbeth/blob/52ddca9c30ee5206b3789b72dd081afe26d11617/crates/primitives/src/botanix/peg_contract.rs#L135 and invalidate the pegin and the test passes. 
In this branch however, when invalid pegins before this invalid ref hash case fails, the pegin block height is reset (due to the clearing the storage changes) and the above condition passes for this specific edge case and the pegin is deemed valid and the test fails. 
So, I guess the fix for this is properly return an error when we do not find a block corresponding to the hash instead of using the default bitcoin checkpoint. Hope I am not missing anything.

crates/ethereum/evm/src/execute.rs
```diff
@@ -263,6 +260,34 @@ where
...
             self.evm_config.fill_tx_env(evm.tx_mut(), transaction, *sender);
 
+            // Store original sender account info before transaction execution.
+            // This is used later to partially revert the state if botanix specific validation
+            // fails.
+            let mut original_sender_info = {
+                const MAX_RETRIES: u8 = 3;
+                let mut attempts = 0;
+
+                loop {
```
arminsabouri: I dont understand the reasoning behind this retry. The revm executor is functional. Meaning it is init with state, and suggests updates.  If sender information doesnt exist when we are executing a transaction retrying shouldnt change that. If behavior does change after retrying something else is really wrong.

arminsabouri (changes requested):

crates/ethereum/evm/src/execute.rs
arminsabouri: Or I dont understand how the revm crate works

bin/test-suite/src/suite/consensus/invalid_transactions/test_invalid_pegin.rs
scottmillner: ah nice catch there
will return None if not found and use checkpoint from storage

crates/ethereum/evm/src/execute.rs
scottmillner: yep agreed, retying doesn't make since
this is deterministic so nothing will change during the retry
please remove

scottmillner: ill make this change

SatwikPrabhu: Sorry for the delay. Am returning an error now without retries. Let me know if it looks good

crates/ethereum/evm/src/execute.rs
```diff
@@ -263,6 +263,25 @@ where
 
             self.evm_config.fill_tx_env(evm.tx_mut(), transaction, *sender);
 
+            // Store original sender account info before transaction execution.
+            // This is used later to partially revert the state if botanix specific validation
+            // fails.
+            // If the sender is not found in the state, we need to error because there is no balance
+            // to subtract from. This shouldn't happen because the tx would have failed.
+            let sender_db_error =
+                BlockExecutionError::Internal(InternalBlockExecutionError::Other(
```
scottmillner, line 272: I didn't create new variants for `BlockExecutionError` bc there are several helper functions on variants and they would all need to be updated.  So went with the verbose way of using `Internal`.

scottmillner (dismissed):
nice job! 🔥

scottmillner (approved):

PR #667, opened 2025-04-15
added amount validation to validate_psbt_by_output

rwlockbg:
Fixes issue: https://github.com/botanix-labs/botanix/issues/989

### DIFF
--- a/bin/btc-server/src/wallet/psbt.rs
+++ b/bin/btc-server/src/wallet/psbt.rs
@@ -261,14 +261,25 @@ pub trait PsbtExt: BorrowMut<Psbt> {
 
     /// Get the fee per output for this PSBT.
     /// Self only needs to be mutable so it can be included in this trait.
-    fn fee_per_output(&self, num_outputs: u64) -> Result<Amount, Box<dyn std::error::Error>> {
+    fn fee_per_output(&self, num_outputs: u64) -> Result<Amount, PsbtFeePerOutputError> {
         // calculate fee per output which is shared across all outputs
         let psbt: &Psbt = self.borrow();
         let fee: Amount = psbt.fee()?;
-        Ok(fee / num_outputs)
+        let fee_per_output: Amount =
+            fee.checked_div(num_outputs).ok_or(PsbtFeePerOutputError::DivideByZero)?;
+        Ok(fee_per_output)
     }
 }
 impl PsbtExt for Psbt {}
+
+/// Errors that can occur when calculating the fee per output for a PSBT.
+#[derive(Debug, thiserror::Error)]
+pub enum PsbtFeePerOutputError {
+    #[error("Failed to calculate fee: {0}")]
+    FeeError(#[from] bitcoin::psbt::Error),
+    #[error("Division by zero error")]
+    DivideByZero,
+}
 
 /// Errors that can occur during the conversion from a PSBT to
 /// a vector of signing packages for Frost signature generation and aggregation.
--- a/bin/btc-server/src/wallet/psbt.rs
+++ b/bin/btc-server/src/wallet/psbt.rs
@@ -2,7 +2,7 @@ use std::{borrow::BorrowMut, collections::BTreeMap};
 
 use bitcoin::{
     psbt::{raw::ProprietaryKey, Input as PsbtInput, Output as PsbtOutput, Psbt},
-    OutPoint, TapSighash, TapSighashType, TxOut,
+    Amount, OutPoint, TapSighash, TapSighashType, TxOut,
 };
 use bitcoin_hashes::Hash;
 use frost_secp256k1_tr as frost;
@@ -257,6 +257,15 @@ pub trait PsbtExt: BorrowMut<Psbt> {
             ret.push(signing_package);
         }
         Ok(ret)
+    }
+
+    /// Get the fee per output for this PSBT.
+    /// Self only needs to be mutable so it can be included in this trait.
+    fn fee_per_output(&self, num_outputs: u64) -> Result<Amount, Box<dyn std::error::Error>> {
+        // calculate fee per output which is shared across all outputs
+        let psbt: &Psbt = self.borrow();
+        let fee: Amount = psbt.fee()?;
+        Ok(fee / num_outputs)
     }
 }
 impl PsbtExt for Psbt {}
--- a/crates/consensus/authority/src/utils.rs
+++ b/crates/consensus/authority/src/utils.rs
@@ -405,10 +405,48 @@ pub fn extract_pegout_ids(psbt: &Psbt) -> Vec<PegoutId> {
 pub fn validate_psbt_by_output(
     psbt: &Psbt,
     destination: &Address,
-    _amount: Amount,
+    amount: Amount,
 ) -> Result<(), PsbtValidationError> {
-    // TODO: to ensure the correct amount is being sent we need a psbt ext function to get the fee
-    // per output and compare it to the amount
+    let absolute_fee = match psbt.fee() {
+        Ok(fee) => fee,
+        Err(e) => match e {
+            bitcoin::psbt::Error::NegativeFee => {
+                return Err(PsbtValidationError::FailedToValidatePsbtByIds(String::from(
+                    "Psbt has negative fee",
+                )))
+            }
+            bitcoin::psbt::Error::FeeOverflow => {
+                return Err(PsbtValidationError::FailedToValidatePsbtByIds(String::from(
+                    "Psbt has fee overflow",
+                )))
+            }
+            _ => {
+                return Err(PsbtValidationError::FailedToValidatePsbtByIds(String::from(
+                    "Psbt fee is corrupted",
+                )))
+            }
+        },
+    };
+    if absolute_fee.to_sat() <= 0 {
+        return Err(PsbtValidationError::FailedToValidatePsbtByIds(String::from(
+            "Psbt fee is zero",
+        )));
+    }
+    if amount.to_sat() <= 0 {
+        return Err(PsbtValidationError::FailedToValidatePsbtByIds(String::from("Amount is zero")));
+    }
+    if psbt.outputs.len() <= 0 {
+        return Err(PsbtValidationError::FailedToValidatePsbtByIds(String::from(
+            "Missing psbt outputs",
+        )));
+    }
+    let fee_per_output = absolute_fee / psbt.outputs.len() as u64;
+    if fee_per_output >= amount {
+        return Err(PsbtValidationError::FailedToValidatePsbtByIds(String::from(
+            "Fee per output is greater than the actual psbt amount",
+        )))
+    }
+
     debug!(target: "consensus::authority::frost_task::validate_psbt_by_ids", "Validating {} outputs in psbt", psbt.outputs.len());
     match psbt.clone().extract_tx() {
         Ok(transaction) => {
--- a/crates/consensus/authority/src/utils.rs
+++ b/crates/consensus/authority/src/utils.rs
@@ -9,7 +9,7 @@ use bitcoin::{
 use btcserverlib::{
     extended_client::{BtcServerExtendedApi, GrpcClientError},
     pegout_id::PegoutId,
-    wallet::psbt::PsbtOutputExt,
+    wallet::psbt::{PsbtExt, PsbtOutputExt},
 };
 use client::{MakeTxRequest, PendingPegout, ScriptBuf, SigningPackage, TxOut, Utxo};
 use futures_util::Future;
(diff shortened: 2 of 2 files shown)


crates/consensus/authority/src/utils.rs
```diff
@@ -405,10 +405,43 @@ pub fn extract_pegout_ids(psbt: &Psbt) -> Vec<PegoutId> {
...
+    };
+    if absolute_fee.to_sat() <= 0 {
+        return Err(PsbtValidationError::FailedToValidatePsbtByIds(String::from(
+            "Psbt fee is zero",
+        )));
+    }
+    if amount.to_sat() <= 0 {
+        return Err(PsbtValidationError::FailedToValidatePsbtByIds(String::from("Amount is zero")));
+    }
+    let fee_per_output = absolute_fee / psbt.outputs.len() as u64;
```
SatwikPrabhu: Maybe we should add a check for division by 0 check here just in case?

crates/consensus/authority/src/utils.rs
```diff
@@ -405,10 +405,43 @@ pub fn extract_pegout_ids(psbt: &Psbt) -> Vec<PegoutId> {
...
+    if fee_per_output >= amount {
+        return Err(PsbtValidationError::FailedToValidatePsbtByIds(String::from(
+            "Fee per output is greater than the actual psbt amount",
+        )))
```
SatwikPrabhu: I'm not following why this solves the issue. Wasnt the issue that the the psbt can be signed with a much higher value than the amount to the destination? This could still happen here if the absolute fee is low.. Maybe I am missing something. 
Also, do we want to avoid the cases where the fee per output is higher than the amount. Why is that not valid?

crates/consensus/authority/src/utils.rs
rwlockbg: good point, I have now added a check for that. Thanks.

crates/consensus/authority/src/utils.rs
rwlockbg: @SatwikPrabhu I dont understand your point here - here we verify what the auditors are asking for. There was also a TODO in the code which is now implemented (cases where the fee per output is higher than the amount) - pls correct me if I am wrong. Where else do you imagine this to be done ?

scottmillner: i see what satwik is saying:

there's really two parts to this fix:
- the fee check that @rwlockbg you have done which addresses the comment that was there
- loop through outputs and verify the amount being pegged out is correct

the second bullet point is what addresses the audit finding:

```
"validate_psbt_by_output() doesn’t validate the amount of the pegout . A malicious coordinator can perform burn action with relatively small value. But sign a psbt with arbitrarily large pegout value to steal bitcoin"
```

that's saying we aren't checking the actual amount being pegged out

- we need to loop through the outputs and calculate the total amount and check that equals the `psbt.amount` - `psbt.fee`

SatwikPrabhu: @scottmillner Thanks a lot understanding and elaborating on my point. Yes this is what I meant. @rwlockbg Apologies. I should have been more clear in my comment and recognized that what you did is also a part of the fix.

rwlockbg: resolved

crates/consensus/authority/src/utils.rs
```diff
@@ -993,6 +1050,40 @@ mod tests {
...
+        );
+
+        let result = validate_psbt_by_output(&psbt, &destination, incorrect_amount);
+
+        // assert that validation fails
+        assert!(result.is_err());
+
+        // check that the error message contains the expected reason
+        if let Err(err) = result {
+            assert!(err.to_string().contains("outputs + fee != amount"));
```
lamafab: > `assert!(err.to_string().contains("outputs + fee != amount"));`

🥲

_You're getting promoted to a JS developer_

crates/consensus/authority/src/utils.rs
lamafab: Why is `fee_per_output` needed at all?

```rust
    let fee_per_output = absolute_fee / psbt.outputs.len() as u64;
    if fee_per_output >= amount {
        return Err(PsbtValidationError::FailedToValidatePsbtByIds(String::from(
            "Fee per output is greater than the actual psbt amount",
        )))
    }
```

This check makes no sense to me(?)

lamafab (dismissed):

crates/consensus/authority/src/utils.rs
SatwikPrabhu: Should we compare the absolute fee with the amount or the fee per output?

crates/consensus/authority/src/utils.rs
scottmillner: them fighting words!!

crates/consensus/authority/src/utils.rs
rwlockbg: @SatwikPrabhu This was the TODO that @0xBEEFCAF3 had written. I just converted it into code. It is not a bad check to have imo. @scottmillner ?

SatwikPrabhu (dismissed):

SatwikPrabhu:
I guess you have some clippy fixes :)

crates/consensus/authority/src/utils.rs
rwlockbg: @lamafab I will take the JS developer joke as an insult :)

crates/consensus/authority/src/utils.rs
scottmillner: @rwlockbg just realized the intent of `validate_psbt_by_output` is just to validate a single output for a single pegout id

and the comment is saying that all pegouts should share the fee
as in, we can't send users exactly what they pegged out for bc the psbt has a fee which should be spread across all users otherwise there's less sats in the multisig than there should be 

ill refactor this since it's my original logic and should've noticed before
sorry for that!

crates/consensus/authority/src/utils.rs
scottmillner: i was wrong...see my comment above

crates/consensus/authority/src/utils.rs
rwlockbg: @scottmillner No worries. Yes, btw. all of your tests are based on creating a single output single pegout id tx.

bin/btc-server/src/wallet/psbt.rs
```diff
@@ -258,6 +258,15 @@ pub trait PsbtExt: BorrowMut<Psbt> {
...
         Ok(ret)
     }
+
+    /// Get the fee per output for this PSBT.
+    /// Self only needs to be mutable so it can be included in this trait.
+    fn fee_per_output(&self, num_outputs: u64) -> Result<Amount, Box<dyn std::error::Error>> {
+        // calculate fee per output which is shared across all outputs
+        let psbt: &Psbt = self.borrow();
+        let fee: Amount = psbt.fee()?;
+        Ok(fee / num_outputs)
```
lamafab: This assumes that a `num_outputs != 0` check is done before this method is called; not an assumption I would make.

crates/consensus/authority/src/utils.rs
```diff
@@ -648,6 +670,22 @@ mod tests {
         psbt
     }
 
+    fn create_psbt_without_fee(num_inputs: usize, address: &bitcoin::Address) -> Psbt {
+        let tx = create_tx(num_inputs, address);
+
+        let input_needed = tx.output.iter().map(|o| o.value.to_sat()).sum::<u64>();
+        let value_per_input = input_needed / num_inputs as u64;
```
lamafab: Same case here; but this is inside a test so a possible division by zero is acceptable, imo.

crates/consensus/authority/src/utils.rs
```diff
@@ -476,8 +493,13 @@ pub async fn validate_psbt_by_ids(
                 PsbtValidationError::FailedToValidatePsbtByIds(String::from("Failed to get pegout data from burn event"))
             })?;
 
-        // check if a corresponding output exists in the psbt
-        validate_psbt_by_output(psbt, &destination, amount)?;
+        // check if a corresponding output exists in the psbt and is for the right amount
+        let fee_per_output = psbt.fee_per_output(pegout_ids.len() as u64)
```
lamafab: This does not need to be re-computed on each iteration(?)

crates/consensus/authority/src/utils.rs
```diff
@@ -417,8 +416,26 @@ pub fn validate_psbt_by_output(
...
+                                return Err(PsbtValidationError::FailedToValidatePsbtByIds(
+                                    String::from(
+                                        "The output value does not match the expected amount",
+                                    ),
+                                ));
+                            }
+                        },
+                        None => {
+                            return Err(PsbtValidationError::FailedToValidatePsbtByIds(
+                                String::from("Calculating expected amount caused an overflow"),
```
lamafab: Technically that's an "underflow" :)

crates/consensus/authority/src/utils.rs
```diff
@@ -401,14 +401,13 @@ pub fn extract_pegout_ids(psbt: &Psbt) -> Vec<PegoutId> {
         .collect()
 }
 
-/// Validate psbt contains the correct output
+/// Validate psbt contains the correct output and amount including the shared fee
 pub fn validate_psbt_by_output(
```
lamafab, line 405: Overall, I'd avoid embedding that many conditions in this function and I'd just take advantage of early returns, for example

```rust
/// Validate psbt contains the correct output
pub fn validate_psbt_by_output(
    psbt: &Psbt,
    destination: &Address,
    amount: Amount,
    fee_per_output: Amount,
    debug!(target: "consensus::authority::frost_task::validate_psbt_by_ids", "Validating {} outputs in psbt", psbt.outputs.len());

    let Ok(transaction) = psbt.clone().extract_tx() else {
        error!(target: "consensus::authority::frost_task::validate_psbt_by_ids", "Failed to extract transaction from psbt {:?}", e);
        return Err(PsbtValidationError::FailedToValidatePsbtByIds(String::from(
            "Failed to extract transaction from psbt",
        )));
    };

    let Some(output) = transaction
        .output
        .iter()
        .find(|output| output.script_pubkey == destination.script_pubkey())
    else {
        error!(target: "consensus::authority::frost_task::validate_psbt_by_ids", "Failed to find matching output in psbt");
        return Err(PsbtValidationError::FailedToValidatePsbtByIds(String::from(
            "Failed to find matching output in psbt",
        )));
    };

    let Some(expected_amount) = amount.checked_sub(fee_per_output) else {
        return Err(PsbtValidationError::FailedToValidatePsbtByIds(String::from(
            "Calculating expected amount caused an underflow",
        )));
    };

    if output.value == expected_amount {
        Ok(())
    } else {
        Err(PsbtValidationError::FailedToValidatePsbtByIds(String::from(
            "The output value does not match the expected amount",
        )))
    }
}
```

It's cleaner.

crates/consensus/authority/src/utils.rs
scottmillner: bring it up with the crate developers 😆 

```
/// Checked subtraction.
    ///
    /// Returns [None] if overflow occurred.
    pub fn checked_sub(self, rhs: Amount) -> Option<Amount> {
        self.0.checked_sub(rhs.0).map(Amount)
    }
```

but yep, will change to underflow

crates/consensus/authority/src/utils.rs
scottmillner: no, so we're looping through the pegout ids for a single psbt

the fee is for the psbt and then should be evenly distributed for all the outputs in the psbt (that each represent a pegout id)

but this did point out that the calculation should be outside the loop since it's only done once

bin/btc-server/src/wallet/psbt.rs
scottmillner: done!

crates/consensus/authority/src/utils.rs
scottmillner: oooh much cleaner, thanks!!

crates/consensus/authority/src/utils.rs
scottmillner: done!

crates/consensus/authority/src/utils.rs
scottmillner: done!

scottmillner (dismissed):

lamafab (dismissed):

scottmillner (approved):

scottmillner:
pr had another approval
had to rebase and fix conflicts

logic didn't change: merging

crates/consensus/authority/src/utils.rs
```diff
@@ -999,13 +1042,14 @@ mod tests {
         let destination = bitcoin::Address::from_str("mrpkDJFJdNGA22FaxCWw6T9oXogXfHU1rh")
             .expect("valid address")
             .assume_checked();
-        let mut psbt = create_psbt(1, &destination);
+        let mut psbt = create_psbt_without_fee(1, &destination);
```
SatwikPrabhu, line 1045: Should we use `create_psbt` here instead to properly test the fee subtraction?

scottmillner: it gets tested in the other unit tests 
it's just the way the helper function `create_tx` works doesn't follow how we create psbts
which causes an issue in this test

PR #669, opened 2025-04-15
retry if consensus checkpoint fails

rwlockbg:
Fixed issue: https://github.com/botanix-labs/botanix/issues/988


SatwikPrabhu (dismissed):

lamafab (commented):
The `tokio-retry` crate is not an "official" tokio project, FYI. Should we include it for logic that's very easy to implement ourselves? It's basically just a loop with a `Err(_)` check and `tokio::time::sleep(_)`

rwlockbg:
> The `tokio-retry` crate is not an "official" tokio project, FYI. Should we include it for logic that's very easy to implement ourselves? It's basically just a loop with a `Err(_)` check and `tokio::time::sleep(_)`

@lamafab Nice point, I have now re-written the logic using our own built retry_exec. Pls check agian.

scottmillner (approved):

PR #673, opened 2025-04-17
Fix(Pegins): Return an error to invalidate pegin v1 with invalid ref block hash

SatwikPrabhu:
Fix for https://github.com/botanix-labs/botanix/issues/996

Also moved the test case for invalid ref block hash in `test_invalid_pegin` to the front of the invalid pegin list as it essentially replicates the edge case encountered in https://github.com/botanix-labs/Macbeth/pull/664#discussion_r2048381145

### DIFF
--- a/bin/test-suite/src/suite/consensus/invalid_transactions/test_invalid_pegin.rs
+++ b/bin/test-suite/src/suite/consensus/invalid_transactions/test_invalid_pegin.rs
@@ -35,11 +35,30 @@ async fn generate_invalid_pegin_metas(
     headers: Vec<Header>,
     provider: Provider<Http>,
 ) -> Vec<(Vec<PeginMeta>, &'static str)> {
-    let mut invalid_pegin_meta_cases = Vec::new();
-
     // Destructure pegin data for easier access
     let (pegin_tx1, vout1, pmt1) = pegin1_data;
     let (pegin_tx2, vout2, pmt2) = pegin2_data;
+    
+    let mut invalid_pegin_meta_cases = Vec::new();
+
+    // Create invalid pegin meta with invalid reference hash for v1
+    let invalid_ref_hash_meta = vec![PeginMeta::V1(PeginMetaV1 {
+        inner: PeginMetaV0 {
+            version: PEGIN_META_VERSION_V1,
+            outpoint: bitcoin::OutPoint::new(pegin_tx1.compute_txid(), vout1),
+            address: eth_account.clone(),
+            aggregate_publickey: secp256k1::PublicKey::from_str(
+                gateway_address_response.aggregate_public_key.as_str(),
+            )
+            .expect("valid public key"),
+            tx: pegin_tx1.clone(),
+            merkle_proof: pmt1.clone(),
+            block_headers: headers.clone(),
+        },
+        ref_block_hash: B256::from_slice(&[0; 32]),
+    })];
+    invalid_pegin_meta_cases.push((invalid_ref_hash_meta, "Invalid reference hash"));
+
     // Create invalid pegin meta with empty headers list
     let empty_headers_meta = vec![PeginMeta::V0(PeginMetaV0 {
         version: PEGIN_META_VERSION_V0,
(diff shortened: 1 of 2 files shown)


lamafab (dismissed):
Interesting; what is the current behavior if the block hash is not found?

rwlockbg (dismissed):

crates/ethereum/evm/src/execute.rs
```diff
@@ -489,6 +489,13 @@ where
                                 })?;
                             bitcoin_checkpoint = package.bitcoin_checkpoint;
                         }
+                        else {
+                            return Err(MintContractError::InvalidPeginData {
+                                error: "No block found for reference block hash".to_string(),
```
arminsabouri: Any reason this error message is different that the one above?
https://github.com/botanix-labs/Macbeth/pull/673/files#diff-7fbb0c1e65c92e8879aa308fee2538af673a2cd6d70f9b5b1103cf4d6554604dL473

They seem to indicate the same thing? I think the above should represent that an internal db error occured and this one is no block found. 

If they are truly the same case (no block found) then we should consolidate error messages.

arminsabouri (changes requested):
One nitpick about error messaging

SatwikPrabhu:
> Interesting; what is the current behavior if the block hash is not found?

The current behavior uses the bitcoin checkpoint that would have been used in a v0 pegin corresponding to the tip of the bitcoin blockchain and would have assumed that the block headers list would also have extended upto the tip.

crates/ethereum/evm/src/execute.rs

rwlockbg (dismissed):

bin/test-suite/src/suite/consensus/invalid_transactions/test_invalid_pegin.rs
```diff
@@ -38,6 +38,24 @@ async fn generate_invalid_pegin_metas(
...
+                gateway_address_response.aggregate_public_key.as_str(),
+            )
+            .expect("valid public key"),
+            tx: pegin_tx.clone(),
+            merkle_proof: pmt.clone(),
+            block_headers: headers.clone(),
+        },
+        ref_block_hash: B256::from_slice(&[0; 32]),
+    })];
+    invalid_pegin_meta_cases.push((invalid_ref_hash_meta, "Invalid reference hash for v1"));
```
arminsabouri: Nit: refrence hash implies v1. 

```suggestion
    invalid_pegin_meta_cases.push((invalid_ref_hash_meta, "Invalid reference hash"));
```

crates/ethereum/evm/src/execute.rs
```diff
@@ -471,7 +471,7 @@ where
                         if let Some(block) = provider
                             .find_block_by_hash(hash, reth_provider::BlockSource::Any)
                             .map_err(|_| MintContractError::InvalidPeginData {
-                                error: "Reference block hash not found".to_string(),
+                                error: "Database error fetching reference block hash".to_string(),
```
arminsabouri: Shouldnt this be a panic condition instead of an event that causes a revert? If the database is corrupt and we can't validate pegins then I would think we expect. @scottmillner can you check my reasoning on this?

SatwikPrabhu: This would cause a revert with the logic introduced in https://github.com/botanix-labs/Macbeth/pull/664 right?

bin/test-suite/src/suite/consensus/invalid_transactions/test_invalid_pegin.rs
SatwikPrabhu: Done!

crates/ethereum/evm/src/execute.rs
arminsabouri: The question is whether or not it should

scottmillner: Thanks for pointing this out @0xBEEFCAF3 

I think the answer is this:
- We only want to revert when the sender submitted a tx that fails validation
- Not if we have an io error (db, call to btc-server, etc) bc that's something wrong on our end not with the tx

So, I agree we should panic if we have a database error

scottmillner (approved):

PR #674, opened 2025-04-17
fix (pegin): don't allow duplicate proofs

SatwikPrabhu:
Fix for https://github.com/botanix-labs/botanix/issues/962

### DIFF
--- a/bin/test-suite/src/suite/consensus/invalid_transactions/test_invalid_pegin.rs
+++ b/bin/test-suite/src/suite/consensus/invalid_transactions/test_invalid_pegin.rs
@@ -28,27 +28,29 @@ use crate::{
 
 // Helper function to generate invalid pegin metas
 async fn generate_invalid_pegin_metas(
-    pegin_tx: bitcoin::Transaction,
-    vout: u32,
+    pegin1_data: (&bitcoin::Transaction, u32, &PartialMerkleTree), // (tx, vout, pmt)
+    pegin2_data: (&bitcoin::Transaction, u32, &PartialMerkleTree), // (tx, vout, pmt)
     gateway_address_response: GatewayAddressResponse,
     eth_account: Address,
-    pmt: PartialMerkleTree,
     headers: Vec<Header>,
     provider: Provider<Http>,
 ) -> Vec<(Vec<PeginMeta>, &'static str)> {
     let mut invalid_pegin_meta_cases = Vec::new();
 
+    // Destructure pegin data for easier access
+    let (pegin_tx1, vout1, pmt1) = pegin1_data;
+    let (pegin_tx2, vout2, pmt2) = pegin2_data;
     // Create invalid pegin meta with empty headers list
     let empty_headers_meta = vec![PeginMeta::V0(PeginMetaV0 {
         version: PEGIN_META_VERSION_V0,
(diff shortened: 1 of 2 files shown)


crates/primitives/src/botanix/mint_validation.rs
```diff
@@ -207,6 +208,13 @@ pub fn try_parse_mint_event(
                     error!("Failed to parse pegin meta: {:?}", err);
                     err
                 })?;
+            if !outpoints.insert(proof.outpoint().clone()) {
+                return Err(ParseMintEventError::InvalidPeginData {
+                    error: PeginDataError::Invalid("Duplicate outpoint hence invalid pegin meta"),
```
arminsabouri: Just an error description would do. The internal variants shouldnt justify why they exists. 

```suggestion
                    error: PeginDataError::Invalid("Duplicate outpoint"),
```

arminsabouri (dismissed):
utACK. Core logic looks correct to me. 
Didnt look over the test changes too deeply

lamafab (dismissed):

rwlockbg (dismissed):

SatwikPrabhu:
> Didnt look over the test changes too deeply

They mainly add a second pegin tx so that our existing test cases having two proofs do not become invalid because of the same outpoint.

crates/primitives/src/botanix/mint_validation.rs
SatwikPrabhu: Done!

rwlockbg (approved):

arminsabouri:
> just realized an implication of making this change
> 
> there is no way to tell if indeed the proof is a duplicate or the user pegged in for the same amount multiple times in the same block (output is just destination and amount from our viewpoint)
> 


Can't we use the outpoint to distinguish duplicates?

scottmillner:
> > just realized an implication of making this change
> > there is no way to tell if indeed the proof is a duplicate or the user pegged in for the same amount multiple times in the same block (output is just destination and amount from our viewpoint)
> 
> Can't we use the outpoint to distinguish duplicates?

yes we can! my brain just hadn't gotten there yet

i removed my comment bc i wasn't thinking about this incorrectly

thanks!

scottmillner (approved):

bin/test-suite/src/suite/consensus/invalid_transactions/test_invalid_pegin.rs
```diff
@@ -324,21 +389,39 @@ pub async fn invalid_pegin(
...
+    // Get block info containing both transactions
     let conf_block_info = bitcoind_rpc.get_block_info(&conf_hash).expect("valid txids");
     it_info_print!("Block info", conf_block_info);
-    let pmt = PartialMerkleTree::from_txids(&conf_block_info.tx, &[false, true]);
+    let tx_indices_in_block = conf_block_info
+        .tx
+        .iter()
+        .map(|txid| txid == &pegin_txid1 || txid == &pegin_txid2)
+        .collect::<Vec<_>>();
+    let pmt = PartialMerkleTree::from_txids(&conf_block_info.tx, &tx_indices_in_block);
```
scottmillner, line 400: nit: we're only using this in the print statement?
bc we create separate pmts for both pegins

scottmillner:
nice job! 🔥

PR #675, opened 2025-04-18
added database persistance and optimizations for wallet state sync

rwlockbg:


crates/storage/db-api/src/models/wallet_sync.rs
```diff
@@ -0,0 +1,154 @@
...
+
+    /// Sets the uuid of the wallet state sync record.
+    pub fn set_uuid(&mut self, uuid: B256) {
+        self.uuid = uuid;
+    }
+
+    /// Adds a data chunk to the wallet state sync record if it doesn't already exist.
+    /// Returns `true` if the data chunk was added, `false` if it was already present.
+    pub fn add_data_if_not_exists(&mut self, data_chunk: Bytes) -> bool {
+        let unique_set: BTreeSet<Bytes> = self.data.iter().cloned().collect();
```
lamafab: Is this expensive `.iter().cloned().collect()` operation necessary here? Why not just `iter().any(...)`?

lamafab (approved):
The database interfaces look fine, but regarding the changes in `crates/consensus/authority/src/wallet_state_sync.rs` I cannot comment since I'd have to revisit  the primary #634 PR again - which I (we) will have to do soon anyway :)

scottmillner (approved):
utAck

crates/storage/db-api/src/models/wallet_sync.rs
scottmillner: @rwlockbg feel free to make the above optimization and merge without reapproval

rwlockbg: done

PR #676, opened 2025-04-19
test(invalid-pegin-v1): create headers list if needed

scottmillner:


scottmillner:
i dont think this actually resolves the issue i was seeing:
- sometimes the test errors with `invalid block header sequence`
- the above seems to be bc the confirmation tx isn't actually on chain which doesn't make sense
- will circle back

PR #679, opened 2025-04-21
fix(validate_outputs): validate change output `scriptPubkey` explicitly

lamafab:
Fixes: https://github.com/botanix-labs/botanix/issues/1002

**Btw**, I do wonder whether it's a good idea to assume that the ordering of `psbt.outputs` corresponds to the ordering of `psbt.unsigned_tx.output`? It looks like the [`Psbt::deserialize`](https://docs.rs/bitcoin/latest/bitcoin/struct.Psbt.html#method.deserialize) method primarily cares about the `unsigned_tx` and constructs the other fields on creation, meaning that ordering must be guaranteed by us internally. But maybe our code makes some other risky assumptions about the PSBT handling.

Just something to keep in mind for the upcoming Pegout `value` (amount) validation PR. I will study this further, those changes presented here might need to be updated based on that upcoming PR.

### DIFF
--- a/bin/btc-server/src/util.rs
+++ b/bin/btc-server/src/util.rs
@@ -444,9 +444,10 @@ pub(crate) fn validate_outputs(
         let agg_pk = public_key_package.verifying_key().to_secp_pk().expect("valid secp pk");
         let expected_script_pubkey = generate_taproot_change_scriptpubkey(&agg_pk);
 
-        let tx = &psbt.unsigned_tx;
-        let has_correct_change = tx.output[idx].script_pubkey == expected_script_pubkey;
+        let change_output =
+            psbt.unsigned_tx.output.get(idx).ok_or(ValidateOutputsError::InvalidChangeOutput)?;
 
+        let has_correct_change = change_output.script_pubkey == expected_script_pubkey;
         if !has_correct_change {
             return Err(ValidateOutputsError::InvalidChangeOutput);
         }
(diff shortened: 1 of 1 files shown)


scottmillner:
> Fixes: [botanix-labs/botanix#1002](https://github.com/botanix-labs/botanix/issues/1002)
> 
> **Btw**, I do wonder whether it's a good idea to assume that the ordering of `psbt.outputs` corresponds to the ordering of `psbt.unsigned_tx.output`? It looks like the [`Psbt::deserialize`](https://docs.rs/bitcoin/latest/bitcoin/struct.Psbt.html#method.deserialize) method primarily cares about the `unsigned_tx` and constructs the other fields on creation, meaning that ordering must be guaranteed by us internally. But maybe our code makes some other risky assumptions about the PSBT handling.
> 
> Just something to keep in mind for the upcoming Pegout `value` (amount) validation PR. I will study this further, those changes presented here might need to be updated based on that upcoming PR.

https://github.com/botanix-labs/Macbeth/pull/667
that's the pr that check output values

I don't think the ordering matters much (it does in this pr since you're getting the change output index)
But generally, as long as all the outputs correspond to L2 pegout tx ids, we make sure there are no duplicates using the outpoint,  and there is only one change output going to the multisig does the order matter?

scottmillner (approved):

SatwikPrabhu (approved):

PR #682, opened 2025-04-22
removed unused (stale) healthcheck request frost protocol message + fmt

rwlockbg:
This PR removes the stale and unused health-check request message for the frost protocol and also does fmt on the project which seems not to have been done previously.


arminsabouri (approved):
Ack changes in crates/net/network/src/frost/messages.rs

SatwikPrabhu (approved):
Looks good to me

scottmillner (approved):

PR #683, opened 2025-04-22
fixed decode_message method in order to prevent panic on malformed in…

rwlockbg:
…puts

Audit issue: https://github.com/botanix-labs/botanix/issues/1003

### DIFF
--- a/crates/net/network/src/frost/messages.rs
+++ b/crates/net/network/src/frost/messages.rs
@@ -312,11 +312,19 @@ impl FrostProtoMessage {
 
     /// Decodes a Frost protocol message from the given message buffer.
     pub fn decode_message(buf: &mut &[u8]) -> Option<Self> {
+        // Check if buffer is empty
         if buf.is_empty() {
             return None;
         }
+
+        // Safely get message ID
         let id = buf[0];
+        if buf.len() < 1 {
+            return None;
+        }
         buf.advance(1);
+
+        // Match message type
         let message_type = match id {
             0x00 => FrostProtoMessageId::Round1Dkg,
             0x01 => FrostProtoMessageId::Round2Dkg,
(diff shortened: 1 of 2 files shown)


lamafab (dismissed):
Btw, I do wonder though why we use this custom message protocol in the first place? Why not just rely on protobuf so we don't have to individually handle all those things? This is already something I noticed in my DKG changes.

rwlockbg:
> Btw, I do wonder though why we use this custom message protocol in the first place? Why not just rely on protobuf so we don't have to individually handle all those things? This is already something I noticed in my DKG changes.

After I discussed this matter long before you joined, the reth team advised us to implement the messaging using this way following some examples which they have had developed as a reference. Surely we could use ptotobuf structs if that is what you mean ? but I am not sure what the overall implications of this would be. I guess we could create a ticket for that and explore after mainnet release ? @scottmillner

crates/net/network/src/frost/messages.rs
```diff
@@ -335,11 +335,19 @@ impl FrostProtoMessage {
 
     /// Decodes a Frost protocol message from the given message buffer.
     pub fn decode_message(buf: &mut &[u8]) -> Option<Self> {
+        // Check if buffer is empty
```
SatwikPrabhu, line 315: Nit. Do we need this comment? Seems self explanatory

crates/net/network/src/frost/messages.rs
```diff
@@ -335,11 +335,19 @@ impl FrostProtoMessage {
...
         if buf.is_empty() {
             return None;
         }
+
+        // Safely get message ID
         let id = buf[0];
+        if buf.len() < 1 {
+            return None;
+        }
```
SatwikPrabhu, line 324: Isnt this check redundant to the earlier check at line 339?

crates/net/network/src/frost/messages.rs
```diff
@@ -335,11 +335,19 @@ impl FrostProtoMessage {
...
         buf.advance(1);
+
+        // Match message type
```
SatwikPrabhu, line 327: Nit. Same as https://github.com/botanix-labs/Macbeth/pull/683/files#r2054680655

crates/net/network/src/frost/messages.rs
scottmillner: this is causing lint job to fail

SatwikPrabhu: The clippy job is suggesting to change this to `buf.is_empty()`. I think this check can be removed altogether

rwlockbg: fixed

crates/net/network/src/frost/messages.rs
rwlockbg: removed

crates/net/network/src/frost/messages.rs
rwlockbg: removed

crates/net/network/src/frost/messages.rs
```diff
@@ -315,8 +315,14 @@ impl FrostProtoMessage {
         if buf.is_empty() {
             return None;
         }
+
+        // Safely get message ID
         let id = buf[0];
+        if buf.is_empty() {
+            return None;
+        }
```
SatwikPrabhu: I think this block can be removed as `buf` cant be empty at this point and we already check this at line 315

rwlockbg: fixed

SatwikPrabhu: I still see the check hasnt been removed.

rwlockbg: sorry, I removed it now @SatwikPrabhu

SatwikPrabhu (dismissed):

scottmillner (dismissed):

scottmillner (approved):

PR #684, opened 2025-04-22
properly handling pending state connection for violated peers

rwlockbg:
Fixes audit finding: https://github.com/botanix-labs/botanix/issues/1005

### DIFF
--- a/crates/net/network/src/frost/manager.rs
+++ b/crates/net/network/src/frost/manager.rs
@@ -1,5 +1,5 @@
 use super::{FrostPeerCommand, FrostProtocolEvent, PeerMessageResponse};
-use crate::{session::Direction, NetworkHandle};
+use crate::{frost::ConnectionEstablishedStatus, session::Direction, NetworkHandle};
 use frost_secp256k1_tr as frost;
 use futures::{Future, StreamExt};
 use reth_network_peers::PeerId;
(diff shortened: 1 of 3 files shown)


crates/net/network/src/frost/manager.rs
```diff
@@ -215,6 +215,29 @@ impl FrostManager {
...
+    ) -> bool {
+        if sender.is_closed() {
+            return false;
+        }
+        let msg = format!(
+            "Received FrostProtocolEvent::ConnectionEstablished event from peer {:?}. Error: {:?}",
+            peer_id, message
+        );
+        if sender.send(status).is_err() {
+            warn!(
```
lamafab, line 232: Nitpick: Why not just be quiet here and return `false` - removing that `warn!`? You can probably remove this method entirely and respond directly, ignoring any errors.

lamafab (dismissed):
Looks good! 🔥

crates/net/network/src/frost/manager.rs
rwlockbg: I just kept it for informational purposes. If the sending errors, at least we would know that. If you insist, I can remove it

crates/net/network/src/frost/manager.rs
```diff
@@ -215,6 +215,29 @@ impl FrostManager {
         self.authorities.contains_key(peer_id)
     }
 
+    fn respond_with_status(
+        sender: tokio::sync::oneshot::Sender<ConnectionEstablishedStatus>,
+        status: ConnectionEstablishedStatus,
+        message: &str,
```
SatwikPrabhu, line 221: I just noticed that the `message` parameter can be constructed directly from the `status` instead of it being a parameter. I mean this is not a strong opinion from my side. The existing code is also perfectly fine. Up to you

rwlockbg: I made it more flexible the way it is now so we can pass longer and more complex error messages if needed. I prefer to keep it as it is, but thanks Satwik.

SatwikPrabhu (dismissed):

SatwikPrabhu (dismissed):

scottmillner (dismissed):

scottmillner (approved):

PR #685, opened 2025-04-23
Test(pegins): Verify revert for multiple invalid pegins

SatwikPrabhu:
Test to verify fix for https://github.com/botanix-labs/botanix/issues/961


contracts/src/test/MultiMintHelper.sol
```diff
@@ -0,0 +1,49 @@
+// SPDX-License-Identifier: MIT
```
scottmillner: we have a separate Contracts repo
can you remove this directory and contract
and add the contract to
https://github.com/botanix-labs/Contracts

SatwikPrabhu: Yes. Apologies. Will do that!

scottmillner (approved):
awesome test!!!
🔥 🔥 🔥

PR #687, opened 2025-04-23
Fix/btnx 01 and btnx 17 gas calculations

scottmillner:
This refactors the tx cost for the mint call.
It also prefunds the sidecar hot wallet with 1000 sats so the first pegin can be done.  Otherwise there's no way to initially move funds to the chain.

related audit issues:
https://github.com/botanix-labs/botanix/issues/944
https://github.com/botanix-labs/botanix/issues/998


crates/chainspec/res/genesis/botanix_mainnet.json
```diff
@@ -7,9 +7,12 @@
     "mixHash": "0x0000000000000000000000000000000000000000000000000000000000000000",
     "coinbase": "0x0000000000000000000000000000000000000000",
     "alloc": {
+        "0x1eb1CE1116AB95b0D3ADc205CB9605f50d3CA51f": {
```
arminsabouri: is the mainnet key already generated? I recognize this address, and I think its the testnet hotwallet address. Is that right?

scottmillner: it's the kms key
yes it's used for the testnet hotwallet address but was thinking we could use it for both since it's in kms

@ronaldstoner @birozuru thoughts?
about having one or two keys?

arminsabouri: Its in KMS yes but I think the mainnet hotwallet key should have tighter IAM policies. For instance if someone were to take control of the testnet sidecar instance they shouldnt be able to sign mainnet botanix-eth txs

ronaldstoner: I agree that we need separate keys and should be using new keys for mainnet (with much tighter IAM policies).

We need to test the new key and address prior to mainnet, just to ensure everything is flowing correctly, locked down, and we can move funds.

scottmillner: via offline discussion,
we're creating separate keys
once i have the new key ill update this pr

rwlockbg:
@scottmillner Cannot really comment on the sc changes as I am not a solidity expert, but wonder if there is not a way to not hardcode the gas but retrieve it rather from the evm, this way ensuring that no magic numbers will be in the contract. Ignore my comment if it makes no sense please.

scottmillner:
> @scottmillner Cannot really comment on the sc changes as I am not a solidity expert, but wonder if there is not a way to not hardcode the gas but retrieve it rather from the evm, this way ensuring that no magic numbers will be in the contract. Ignore my comment if it makes no sense please.

thanks for the comment
yes, we could do this in the validation where we know the final tx cost and we would directly increment the `refundReceiver` and decrement the `destination`

but the tx cost refund might not be the actual full amount bc of the priority fee attack that this helps resolve: we need to use `block.basefee`and ive gone back and forth on what the best approach is

as far as magic numbers, that's why i added that comment explaining how i got the number
but yeah do get it's a hardcoded amount

i'll bring this up in a standup to get everyone's feedback, thanks!

crates/chainspec/res/genesis/botanix_mainnet.json
scottmillner: udpated with separate mainnet hot wallet address

scottmillner:
> > @scottmillner Cannot really comment on the sc changes as I am not a solidity expert, but wonder if there is not a way to not hardcode the gas but retrieve it rather from the evm, this way ensuring that no magic numbers will be in the contract. Ignore my comment if it makes no sense please.
> 
> thanks for the comment yes, we could do this in the validation where we know the final tx cost and we would directly increment the `refundReceiver` and decrement the `destination`
> 
> but the tx cost refund might not be the actual full amount bc of the priority fee attack that this helps resolve: we need to use `block.basefee`and ive gone back and forth on what the best approach is
> 
> as far as magic numbers, that's why i added that comment explaining how i got the number but yeah do get it's a hardcoded amount
> 
> i'll bring this up in a standup to get everyone's feedback, thanks!

we had offline discussion in standup meeting and decided to keep as is: using hardcoded gas amount

contracts/test/Minting.t.sol
```diff
@@ -80,4 +80,24 @@ contract MintingTest is Test {
...
+
+    // -- Gas Benchmarking Tests --
+    /// Used to help determine the `txCost` of the mint() function
+    /// which is sent to the refundAddress.
+
+    /// Measure the total gas used by mint() with 160 bytes of metadata
+    /// That is 2 bitcoin headers worth of metadata.
+    /// run `forge test --gas-report` to see the gas used
+    function testFullMintCost() public {
+        uint256 length = 160;
```
SatwikPrabhu, line 92: Maybe the length can be initialized to a predefined constant?

scottmillner: yep definitely could...my thinking:
- since im only using that value once, I'll leave it inline
- if i ever need to use it more than once, I'll pull it out into a const
thanks for the comment

SatwikPrabhu (dismissed):

contracts/src/Minting.sol
```diff
@@ -3,29 +3,12 @@ pragma solidity 0.8.13;
...
-    // > pay 8 times the number of 32 byte words written to the chain. Because
-    // > this region is stored in memory before being emitted, the memory
-    // > expansion cost must be accounted for also.
-    uint constant BASE_GAS_MINT_EVENT = 1157;
+    /// Gas used when calling mint() function.
+    /// This value is based on forge and live testing for a cold EOA destination with 160 bytes of metadata.
+    /// This is not intended to be a precise value but a rough estimate.
+    /// If the destination is a contract account, the gas used will be higher.
+    /// It is the caller's responsibility to check the destination before calling mint().
+    uint constant mintGasUsed = 150_000;
```
lamafab: Nitpick; should be `UPPER_CASE`, no?

lamafab (dismissed):
I'll have to double-check this tomorrow again.

contracts/src/Minting.sol
scottmillner: you sir are correct! much obliged
change has been made

SatwikPrabhu (approved):

rwlockbg (approved):

PR #689, opened 2025-04-24
fix(validate_psbt_by_ids): validate destination against corresponding pegout `scriptPubkey`

lamafab:
Fixes https://github.com/botanix-labs/botanix/issues/1008

This is the simplest way I could figure out without doing a (major) refactor, because it leaves some preferable improvements.

Changes:

* `validate_psbt_by_ids` (in `reth-authority-consensus`): As of now on `main`, this function goes through all pegout Ids, retrieves the corresponding pegout data from the database and checks if each pegout `scriptPubkey` **exists somewhere in the output list**. With this PR, we now keep track of the _position_ of the retrieved pegout Id, and validate the `scriptPubkey` at the corresponding _position_ directly.

* `validate_psbt` (in `btc-server`): This was already fixed in #679. Respectively, we make sure that there is **at most one non-pegout output**, and we validate that its `scriptPubkey` is a valid change destination.

Notably, we do partial PSBT validations in different places - for no clear reason(?). I've expanded the relevant docs:

```rust
/// Validates a PSBT by verifying its pegout outputs against data from the
/// receipt database.
///
/// This function extracts pegout IDs from the PSBT, retrieves the corresponding
/// pegout data from the database, and validates each output's destination and
/// amount.
///
/// # Warning
///
/// This function only validates pegout outputs. It does NOT validate:
/// * Non-pegout outputs (usually change outputs)
/// * Duplicate pegout IDs
/// * Conflicting inputs
/// * Change outputs
///
/// These responsibilities are handled by the `btc-server`.
pub async fn validate_psbt_by_ids(
    client: impl ReceiptProvider + Clone,
    btc_network: bitcoin::Network,
    psbt: &Psbt,
    // ...
}
```

 In my opinion, the entire PSBT should be validated in full in one place. Right now there are a lot of moving parts, which makes PSBT validation error prone and harder to test.

### DIFF
--- a/crates/consensus/authority/src/utils.rs
+++ b/crates/consensus/authority/src/utils.rs
@@ -382,7 +382,7 @@ pub fn is_known_minting_contract(
     Ok(())
 }
 
-#[derive(Debug, thiserror::Error)]
+#[derive(Debug, thiserror::Error, PartialEq, Eq)]
 /// Represents errors that can occur during psbt validation
 pub enum PsbtValidationError {
     #[error("Failed to validate psbt by ids: {0}")]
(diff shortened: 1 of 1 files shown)


rwlockbg (approved):
lgtm

SatwikPrabhu (approved):

scottmillner (approved):
🔥 
and agreed we should move all these checks into one place
and thanks for the clear documentation in the comments!

PR #690, opened 2025-04-24
Feature/btc yield send block fees to fee receiver contract

scottmillner:
This refactors the block fees split to include the FeeReceiver
This is from the Repo's README (it's a private repo)

```
sBTC is an upgradeable Staked Bitcoin Contract. It is designed to function on Botanix as a first implementation for their native staking system. It was designed with the goal of being

Upgradeable (for future chain upgrades that are already planned)
ERC4626 Compatible (For yield integration elsewhere)
Synthetix staking based (to prevent front-running of rewards)
In order to meet this, Openzeppelin's ERC4626 Upgradeable was chose as the base implementation, and staking logic from [apxEth](https://github.com/dinero-protocol/pirex-eth-contracts/blob/master/src/AutoPxEth.sol) was taken. Additionally, a separate FeeReceiver contract was created to receive rewards from the Consensus layer, rather than relying on balance being added directly to the contract for ease of accounting. The only FeeReceiver being able to call notifyRewards was changed for initialization simplification.
```
The `FeeReceiver` is a contract specifically designed to receive block fees directly at the protocol level.  The fees are periodically swept into the yield contract.


crates/consensus/common/src/utils.rs
```diff
@@ -11,13 +11,19 @@ pub fn unix_timestamp() -> u64 {
...
+/// 1. 50% to LST `FeeReceiver`
+/// 2. 40% to Botanix
+/// 3. 10% to the Block Fee Recipient (determined by the node operator)
+pub const fn block_fees_split(total_block_fees: u128) -> (u128, u128, u128) {
+    // 50% to LST `FeeReceiver`
+    let fee_receiver_fees = total_block_fees * 50 / 100;
+    // 40% to Botanix
+    let botanix_fees = total_block_fees * 40 / 100;
+    // 10% to the Block Fee Recipient (determined by the node operator)
+    let block_fee_recipient_fees = total_block_fees * 10 / 100;
```
PhoenixCodeCraft: I believe we can run into rounding errors here where total of the three splits doesn't equal the block reward. This could result in a depeg.

In the previous split this wasn't an issue since you subtract from total. 

Solution would be to have one of the three = the total - (sum of the other 2).

scottmillner: ah yep, thanks!

scottmillner: done!

crates/consensus/authority/src/lib.rs
```diff
@@ -88,8 +88,9 @@ impl AuthorityConsensus {
         // Determine the parent gas limit, considering elasticity multiplier on the London fork.
         let parent_gas_limit =
             if self.chain_spec.fork(EthereumHardfork::London).transitions_at_block(header.number) {
-                parent.gas_limit *
-                    self.chain_spec
+                parent.gas_limit
+                    * self
+                        .chain_spec
```
rwlockbg, line 93: Woudl it not be better if we use `checked_mul` or some safe math everywhere and apply defaults when we do `unwrap_or_default` - purely from safety perspective as I believe that gas is something that could vary and might blow up on occasions (just guessing) ?

crates/primitives-traits/src/constants/mod.rs
```diff
@@ -130,6 +130,9 @@ pub const BOTANIX_MAINNET_GENESIS: B256 =
 pub const BOTANIX_TESTNET_GENESIS: B256 =
     b256!("3797638175875c37cefa72ef546db685e43c81ba4af8238b48a495f98d61588d");
 
+/// LST `FeeReceiver` address: `0xDAfdb73cF5f14417fc5aEd97486704dC7C6ee0a4`
+pub const LST_FEE_RECEIVER: &str = "DAfdb73cF5f14417fc5aEd97486704dC7C6ee0a4";
```
rwlockbg: Dont we want to set this dynamically ?

crates/revm/src/state_change.rs
```diff
@@ -41,19 +43,29 @@ pub fn post_block_balance_increments(
...
+        && block_fee_recipient != Address::ZERO
+        && chain_spec.botanix_fee_recipient.is_some()
+    {
+        let (fee_receiver_fees, botanix_fees, block_fee_recipient_fees) =
+            utils::block_fees_split(fees);
 
+        // FeeReceiver fees
+        *balance_increments
+            .entry(Address::from_str(LST_FEE_RECEIVER).expect("FeeReceiver to exist"))
+            .or_default() += fee_receiver_fees;
```
rwlockbg: maybe again safe math and either use wrapped or checked addition ?

crates/consensus/common/src/utils.rs
```diff
@@ -11,13 +11,19 @@ pub fn unix_timestamp() -> u64 {
...
+/// 3. 10% to the Block Fee Recipient (determined by the node operator)
+pub const fn block_fees_split(total_block_fees: u128) -> (u128, u128, u128) {
+    // 50% to LST `FeeReceiver`
+    let fee_receiver_fees = total_block_fees * 50 / 100;
+    // 40% to Botanix
+    let botanix_fees = total_block_fees * 40 / 100;
+    // 10% to the Block Fee Recipient (determined by the node operator)
+    let block_fee_recipient_fees = total_block_fees - fee_receiver_fees - botanix_fees;
+
+    (fee_receiver_fees, botanix_fees, block_fee_recipient_fees)
```
rwlockbg: Maybe use safe math here as well (to prevent over/underflows) ?

crates/primitives-traits/src/constants/mod.rs
```diff
@@ -130,6 +130,9 @@ pub const BOTANIX_MAINNET_GENESIS: B256 =
 pub const BOTANIX_TESTNET_GENESIS: B256 =
     b256!("3797638175875c37cefa72ef546db685e43c81ba4af8238b48a495f98d61588d");
 
+/// LST `FeeReceiver` address: `0xDAfdb73cF5f14417fc5aEd97486704dC7C6ee0a4`
```
arminsabouri: Do we need this?

crates/primitives-traits/src/constants/mod.rs
arminsabouri: Magic hex. How was this value obtained? A short description as a comment would help others understand when they stumble onto this value

crates/revm/src/state_change.rs
```diff
@@ -41,19 +43,29 @@ pub fn post_block_balance_increments(
...
-    let address = block_fee_recipient_address.unwrap_or(Address::ZERO);
+    let block_fee_recipient = block_fee_recipient_address.unwrap_or(Address::ZERO);
 
-    if fees > 0 && address != Address::ZERO && chain_spec.botanix_fee_recipient.is_some() {
-        let (botanix_fees, builder_fees) = utils::block_fees_split(fees);
+    if fees > 0
+        && block_fee_recipient != Address::ZERO
+        && chain_spec.botanix_fee_recipient.is_some()
+    {
+        let (fee_receiver_fees, botanix_fees, block_fee_recipient_fees) =
```
arminsabouri: Nit: variable naming

fee_receiver_fees => lst_fee_receiver

bin/test-suite/src/context/mod.rs
```diff
@@ -7,6 +7,7 @@ use crate::{
 };
 
 pub const BOTANIX_FEE_RECEIPIENT: &str = "0xb8c03cb8C9bAC79c53926E3C66344C13452105f5";
+pub const LST_FEE_RECEIVER: &str = "0xDAfdb73cF5f14417fc5aEd97486704dC7C6ee0a4";
```
arminsabouri, line 10: This const has a 0x prefix while the other const in this file does not. Is that ok?

crates/primitives-traits/src/constants/mod.rs
arminsabouri: I think this should be coming from a fed.toml

scottmillner: this should never change and should be known ahead of time
so wasn't sure what benefit we get by putting it in the federation.toml?

scottmillner: but i guess if there ever was an issue with that contract updating the federation.toml would be easier?

crates/primitives-traits/src/constants/mod.rs
scottmillner: i added the comment to stay the same as the rest of the file:
- all hashes/addresses in the file have a similar comment but yeah nothing gained from it really

bin/test-suite/src/context/mod.rs
scottmillner: the const above it has a 0x
and it's ok, test works and passes

crates/consensus/authority/src/lib.rs
scottmillner: this is outside the scope of this pr but something we could look at

crates/primitives-traits/src/constants/mod.rs
arminsabouri: Part of the reason for fed.toml was to allow anyone to run a federation. And what lives in that file is configurables. It seems like we've kinda departed from that product direction. But if we do want to follow that paradigm then `LST_FEE_RECEIVER` should be optional value.

crates/primitives-traits/src/constants/mod.rs
arminsabouri: All good

crates/primitives-traits/src/constants/mod.rs
scottmillner: ah yep, that makes sense
ill add it to the fed.toml, thanks!

scottmillner: passing in federation.toml now

scottmillner: removed this const as it's now being passed from the federation.toml
added comment in chainspec explaining what the contract address is

crates/revm/src/state_change.rs
scottmillner: done!

crates/consensus/common/src/utils.rs
scottmillner: done!

crates/revm/src/state_change.rs
rwlockbg: nice!

rwlockbg (approved):

PR #693, opened 2025-04-29
feat: added finalized pegout ids L2 check

rwlockbg:


crates/consensus/authority/src/utils.rs
```diff
@@ -298,6 +300,130 @@ pub(crate) async fn epoch_pegouts(
...
+                match client.receipts_by_block(BlockHashOrNumber::Number(block.header.number)) {
+                    Ok(Some(receipts)) => {
+                        for receipt in receipts {
+                            if !receipt.success {
+                                continue;
+                            }
+                            for (index, log) in receipt.logs.into_iter().enumerate() {
+                                let hash = block.header.hash_slow();
+                                if let Ok(Some(_p)) = try_parse_burn_event(&log, btc_network) {
+                                    let mut tx_hash_array = [0u8; 32];
```
scottmillner: this isn't the block header hash

this is the TxHash (aka txid):

```
/// A transaction hash is a keccak hash of an RLP encoded signed transaction.
#[doc(alias = "TransactionHash")]
pub type TxHash = B256;
```

the index is used to find the correct log in the tx receipt

i think you could do:
- `client.transactions_by_block()`
- `client.receipts_by_block()`
- i believe the txs and receipts should be returned in the same order, if so
  - then zip them (just grab the txHash off the Tx) to associate the TxHash with the Receipt:
  ```
  // 3. zip them (if they’re returned in the same order)
  for (tx, receipt) in txs.into_iter().zip(receipts.into_iter()) {...}
  ```

crates/consensus/authority/src/utils.rs
```diff
@@ -298,6 +300,130 @@ pub(crate) async fn epoch_pegouts(
...
+    match client.block_by_number(block) {
+        Ok(Some(block)) if bloom_contains_pegout(block.header.logs_bloom) => {
+            match client.receipts_by_block(BlockHashOrNumber::Number(block.header.number)) {
+                Ok(Some(receipts)) => {
+                    for receipt in receipts {
+                        if !receipt.success {
+                            continue;
+                        }
+                        for (index, log) in receipt.logs.into_iter().enumerate() {
+                            let hash = block.header.hash_slow();
```
scottmillner: see comment above...needs to be TxHash not the block header hash

crates/consensus/authority/src/utils.rs
scottmillner: and this method isn't being used anywhere?

crates/storage/db-api/src/models/wallet_sync.rs
```diff
@@ -88,13 +115,18 @@ impl WalletStateSyncRecord {
...
+    /// Adds a data chunk with its block number to the wallet state sync record if it doesn't
+    /// already exist. Returns `true` if the data or block was added, `false` if it was already
+    /// present.
+    pub fn add_data_if_not_exists(&mut self, data_chunk: Bytes, block_number: u64) -> bool {
         if self.data.iter().any(|data| data == &data_chunk) {
             return false;
         }
+        if self.blocks.iter().any(|block| block == &block_number) {
+            return false;
+        }
```
SatwikPrabhu, line 127: This checks whether `data_chunk` OR `block_number` is present. Should we check for the presence of the pair i.e., `data_chunk` AND `block_number`?

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -4066,7 +4066,9 @@ impl<TX: DbTxMut + DbTx> WalletStateSyncWriter for DatabaseProvider<TX> {
             .map(|(_, record)| record);
 
         if let Some(mut wallet_state_sync_record) = wallet_state_sync_record {
-            wallet_state_sync_record.append_data_chunks(data);
+            for (block, data_chunk) in data {
+                wallet_state_sync_record.append_data_with_block(data_chunk, block);
+            }
             self.tx.put::<tables::WalletStateSyncs>(peer_id, wallet_state_sync_record)?;
         }
         Ok(())
```
SatwikPrabhu, line 4074: Noticed that we do not return an error here if we dont the `wallet_state_sync_record` is `None` and we silently succeed. Is this fine?

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -3871,7 +3871,7 @@ impl<TX: DbTx> WalletStateSyncReader for DatabaseProvider<TX> {
     fn get_minimum_superset(
```
SatwikPrabhu, line 3871: Maybe not part of this PR, but the name of this method is slightly misleading. The method name suggests we would return the minimum superset that meets the count threshold. But the method just returns the union of all the records if the `min_required_criterion` is met.

SatwikPrabhu:
I think a lot of the functionalities need to be unit tested. But that can be done in the future.

crates/storage/db-api/src/models/wallet_sync.rs
```diff
@@ -103,6 +135,20 @@ impl WalletStateSyncRecord {
         self.data.iter().cloned().collect()
     }
 
+    /// Converts the blocks to a set of unique blocks.
+    pub fn blocks_to_set(&mut self) -> BTreeSet<u64> {
```
scottmillner: this isn't used anywhere?

crates/storage/db-api/src/models/wallet_sync.rs
rwlockbg: The way the db works is that we cannot have a collection of both block_number and data_chunk at the same time. The way its organized right now wont allow anyone to remove a single item so that it is only found in one of them and not in the other. I prefer to keep it that way as the incoming tuple is either going to present in both block number and data or missing at all.

crates/storage/provider/src/providers/database/provider.rs
rwlockbg: This is the same pattern that has been applied elsewhere in the other methods on db queries where `seek_exact` has been used. The problem here is that all the db interfaces either return system errors or just empty - both on success and not found. I have just followed this pattern. (cf. pub type ProviderResult<Ok> = Result<Ok, ProviderError>;) If we were to change this now, this might break the api design and start feeling weird. The idea behind it is - you search if you find your thing, you modify it or return it (in many methods there is Option<T> being returned btw.) and you dont fail on non-found items. Only in system errors.

crates/storage/provider/src/providers/database/provider.rs
rwlockbg: we return the union of all records that meet the min criterion which is the minimum superset - is not it ? Do you want me to rename it ?

crates/storage/db-api/src/models/wallet_sync.rs
rwlockbg: nahe, I just added it for completeness. Not needed, have removed the unused api methods. Thanks.

crates/consensus/authority/src/utils.rs
rwlockbg: Nope, I was thinking of using it, but then I changed my approach. Now I have fully removed get_epoch_pegouts_ids. Thanks

crates/consensus/authority/src/utils.rs
rwlockbg: thanks, Scott. Fixed! Test passes like butter.

bin/btc-server/proto/btc_server.proto
```diff
@@ -77,6 +77,11 @@ message PendingPegout {
     uint64 height = 4;
 }
 
+message FinalizedPegout {
+    bytes id = 1;
+    uint64 block_number = 2;
```
arminsabouri: I would be more verbose with naming here since we have 2 blockchains.
`botanix_block_height`

crates/consensus/authority/src/utils.rs
```diff
@@ -298,6 +300,79 @@ pub(crate) async fn epoch_pegouts(
     Ok(pegouts)
 }
 
+/// Returns all pegouts ids in a given block
+///
+/// # Arguments
+///
+/// * `block` - The current block number
```
arminsabouri: Is this the current tip or any block height. Seems like conflicting comments in this function

scottmillner (approved):

crates/consensus/authority/src/utils.rs
rwlockbg: yep, thanks for this, fixed.

bin/btc-server/proto/btc_server.proto
rwlockbg: thanks for pointing this out. I have renamed it.

crates/storage/provider/src/providers/database/provider.rs
SatwikPrabhu: This is honestly a nit 😅 . I found the minimum superset to mean that its the minimal/smallest set of records that meet the min critereon.

SatwikPrabhu (approved):

PR #694, opened 2025-04-30
fix(authority): non-deterministic bitcoin checkpoint validation

shumkov:
This PR fixes https://github.com/botanix-labs/botanix/issues/982

## What was done?
- Introduced a checkpoints chain to keep historical and weak checkpoints as well.
   - Maintain chain consistency (prev hash =  block hash).
- Validate a proposed checkpoint against historical, strong and week checkpoints.
- Improved Bitcoin checkpoint sync process:
   - Sync multiple checkpoints
   - Fixed blocking IO usage in the main thread pool that could interfere with other tokio tasks. Run the sync process in a blocking thread and release it while waiting for the next sync. 
   - Trigger sync when a new Bitcoin block arrives (using Bitcoin ZMQ events) if zmq socket address is provided. Otherwise, sync is triggered every 5 seconds
- Use the same mockall version among the crates

The checkpoint sync process via ZMQ (optional) requires Bitcoin node to expose the hashblock ZMQ socket:
```
-zmqpubhashblock=tcp://0.0.0.0:28332
```
Additional configuration from the poa node is required too:
```
--bitcoind.zmq.hashblock-address=tcp://[HOST]:28332
```

## How Has This Been Tested?
- Added unit tests for various pegin amounts
- Manually tested with 0.0005 BTC pegin transactions
- Verified error messages are clear and helpful

## Breaking Changes
None

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Introduced a comprehensive Bitcoin checkpoint management system with configurable confirmation depths.
  - Added asynchronous synchronization of Bitcoin checkpoints using ZMQ and RPC for real-time updates.
  - Added a new configuration option for Bitcoin daemon ZMQ hashblock address.

- **Improvements**
  - Updated Botanix chain specs and related components to use the new Bitcoin checkpoint confirmation depth.
  - Enhanced consensus and client modules to utilize the Bitcoin checkpoints chain for improved verification.
  - Increased integration job timeout to 15 minutes for more stable CI.

- **Dependency Updates**
  - Added `bitcoincore-zmq`, `arc-swap`, and reorganized `mockall` dependencies to support new features.

- **Bug Fixes**
  - Improved error handling for stale blocks and RPC errors during Bitcoin checkpoint synchronization.

- **Refactor**
  - Replaced single checkpoint handling with a thread-safe checkpoint chain across modules.
  - Renamed confirmation depth fields for consistency and clarity.

- **Style**
  - Minor formatting and readability improvements.

- **Tests**
  - Added extensive tests covering checkpoint chain behavior, synchronization, and error scenarios.


bin/reth/src/commands/poa/mod.rs
```diff
@@ -437,62 +433,25 @@ impl<Ext: clap::Args + fmt::Debug> PoaNodeCommand<Ext> {
...
-                        );
-                        info!("Async bitcoin Tip {}:{}", height, tip_hash,);
-                        *bitcoin_block_header.write().await =
-                            Some((header, finalized.height as u32));
-                        last_tip = tip_hash;
-                    }
-                    tokio::time::sleep(SLEEP).await;
-                }
-            }),
+        // TODO: Magic numbers. Shall I add it configuration?
```
shumkov: @scottmillner @SatwikPrabhu what do you think guys?

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -1420,22 +1421,13 @@ where
...
-        };
-
         // check non-deterministic data: btc block hash and aggregate public key
-        if bitcoin_checkpoint_block_hash != non_deterministic_data.bitcoin_block_hash {
-            warn!("Bitcoin block hash mismatch");
+        if !self.bitcoin_checkpoints.contains_by_hash(non_deterministic_data.bitcoin_block_hash) {
+            warn!(
+                available_checkpoints = %self.bitcoin_checkpoints,
+                expected = %non_deterministic_data.bitcoin_block_hash,
+                "Bitcoin block hash mismatch"
```
lamafab: nitpick; error message is misleading. It should say something along the lines of the bitcoin hash not being present in the valid checkpoint range.

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -1964,6 +1956,9 @@ mod tests {
         let transaction_pool =
             RethPool::eth_pool(validator.clone(), blob_store, TxPoolArgs::default().pool_config());
 
+        let bitcoin_checkpoints_chain =
+            BitcoinCheckpointsChain::try_new(1, 0, 0).expect("create a valid chain");
```
lamafab, line 2419: Shouldn't this be passed on from `poa/mod.rs`? You create the same `BitcoinCheckpointsChain` twice, with different parameters(?).

crates/consensus/authority/src/bitcoin_checkpoint/chain.rs
```diff
@@ -0,0 +1,729 @@
...
+    /// ## Returns
+    /// Ok(()) if the checkpoint was added, or an error
+    ///
+    /// ## Errors
+    /// * `StaleBlockAdded` - If the checkpoint doesn't connect to the current chain
+    pub fn push(&self, checkpoint: BitcoinCheckpoint) -> Result<(), BitcoinCheckpointError> {
+        let checkpoints = self.checkpoints.load_full();
+
+        if let Some(recent) = checkpoints.back() {
+            if checkpoint.header.prev_blockhash != recent.hash {
```
lamafab, line 121: What if a checkpoint gets missed? In your `syncer.rs`, you (correctly) only apply the number of blocks that you actually need to retain (`blocks_to_sync`), but that last block (`from_height`) might no contain the `prev_blockhash` to the last cached block.

```rust
// In: crates/consensus/authority/src/bitcoin_checkpoint/syncer.rs

// How many blocks we need to sync (limited by the chain size)
let chain_size_limit = self.checkpoints_chain.size_limit() as u64;
let blocks_to_sync = need_to_sync.min(chain_size_limit);

let top_confirmed_height = tip_height - (lowest_confirmation_depth - 1);

// We push from oldest to newest, so we start `blocks_to_sync`−1 below the top.
let from_height = top_confirmed_height - (blocks_to_sync - 1);
let to_height = top_confirmed_height;

let mut synced_checkpoints = Vec::new();
for height in from_height..=to_height {
	let confirmed_hash = map_rpc_error!(self.rpc, get_block_hash(height))?;
	let header = map_rpc_error!(self.rpc, get_block_header(&confirmed_hash))?;

	// Create, report and push the checkpoint
	let bitcoin_checkpoint = BitcoinCheckpoint::new(header, height as u32);

	synced_checkpoints.push(SyncedCheckpointInfo::from(&bitcoin_checkpoint));

	self.checkpoints_chain.push(bitcoin_checkpoint)?;
}
```

I see a potential issue here that when a gap occurs, such as a `bitcoind` interruption, the _push_ results in a `BitcoinCheckpointError::StaleBlockAdded { .. }`, with no proper recovery. This would require the node operator to restart the node.

crates/consensus/authority/src/bitcoin_checkpoint/chain.rs
```diff
@@ -0,0 +1,729 @@
...
+        }
+
+        let checkpoints = self.checkpoints.load();
+        if checkpoints.is_empty() {
+            return None;
+        }
+
+        // Translate depth to index from back first
+        let lowest_confirmation_depth = *self.confirmation_window.start(); // e.g. 4
+        let index_from_back = depth - lowest_confirmation_depth;
```
lamafab, line 179: Can this underflow? I'd be very careful with direct arithmetic operations, consider using `saturating_sub` or `checked_sub` for this.

crates/consensus/authority/src/bitcoin_checkpoint/chain.rs
```diff
@@ -0,0 +1,729 @@
...
+            let result = chain.push(invalid_checkpoint);
+
+            assert!(matches!(
+                result,
+                Err(BitcoinCheckpointError::StaleBlockAdded {
+                    expected_prev_block_hash,
+                    added_prev_block_hash
+                })
+                if expected_prev_block_hash == checkpoint1_hash && added_prev_block_hash == BitcoinBlockHash::all_zeros()
+            ));
```
lamafab, line 370: Uhm, this `assert!` structure is kind of odd?

crates/consensus/authority/src/bitcoin_checkpoint/chain.rs
```diff
@@ -0,0 +1,729 @@
...
+    /// * `weak_checkpoints_count` - How many checkpoints before the strong confirmation depth to keep (depth < strong)
+    ///
+    /// ## Returns
+    /// A new `BitcoinCheckpointsChain` or an error if the parameters are invalid
+    ///
+    /// ## Errors
+    /// * `ZeroStrongConfirmationDepth` - If strong confirmation depth is zero
+    /// * `WeakCheckpointsCountTooBig` - If weak checkpoints count is greater than strong confirmation depth
+    /// * `ChainParamsTooLarge` - If the parameters would cause numeric overflow
+    pub fn try_new(
```
lamafab, line 52: I think it's acceptable to simply call this `new`, even though it might fail.

crates/consensus/authority/src/bitcoin_checkpoint/syncer.rs
```diff
@@ -0,0 +1,425 @@
...
+    /// This method runs in a loop and never returns, so it should be run in a separate task or thread.
+    pub async fn sync(self) {
+        // We need interior mutability so that we can move `self` into spawn_blocking
+        let syncer = Arc::new(Mutex::new(self));
+
+        loop {
+            // Sync bitcoin checkpoints to the tip in blocking task
+            let syncer_clone = Arc::clone(&syncer);
+            let result = tokio::task::spawn_blocking(move || {
+                let mut syncer = syncer_clone.lock().unwrap();
```
SatwikPrabhu: Maybe use `expect()` here for a more clear panic message in case of lock poisoning?

crates/consensus/authority/src/bitcoin_checkpoint/chain.rs
```diff
@@ -0,0 +1,729 @@
...
+    // We update every 10 mins, so we use `ArcSwap` to make reads lock-free and fast.
+    checkpoints: ArcSwap<VecDeque<BitcoinCheckpoint>>,
+
+    /// Range of confirmation depths to keep in the window
+    /// front=lowest, back=highest
+    confirmation_window: std::ops::RangeInclusive<usize>,
+
+    /// Size of the confirmation window (precalculated for efficiency)
+    confirmation_window_size: usize,
+}
```
lamafab, line 35: Could you clarify why you chose to implement it this way? Specifically the usage of `RangeInclusive`... couldn't you just rely on `checkpoints.len()` and `confirmation_window_size` in the logic? And why distinguish between `historical_checkpoints_count` and `weak_checkpoints_count`? This part kind of confuses me.

bin/reth/src/commands/poa/mod.rs
lamafab: This must be equal for all participants, so it should either be hard-coded in the binary or specified in the `ChainSpec`. Additionally; I think we agreed that it should retain the last `chain.parent_confirmation_depth + 18` of blocks in memory?

crates/consensus/authority/src/bitcoin_checkpoint/syncer.rs
```diff
@@ -0,0 +1,425 @@
...
+
+            match result {
+                Ok(synced_checkpoints) => {
+                    tracing::info!(
+                        ?synced_checkpoints,
+                        "Asynced task synced {} bitcoin checkpoints",
+                        synced_checkpoints.len()
+                    )
+                }
+                Err(e) => tracing::warn!("Async task failed to sync bitcoin checkpoints: {e}"),
```
lamafab: As mentioned in a previous comment, there's is no way to "fill the gap" after a `bitcoind` interruption. That would forever print an error, forcing the operator to restart the node.

lamafab (commented):
Very clean code! 🔥

I'll have to do a second review of the `chain.rs` code after you clarify my question about the inner structures.

crates/consensus/authority/src/bitcoin_checkpoint/chain.rs
```diff
@@ -0,0 +1,729 @@
...
+        }
+
+        let confirmations_from = strong_confirmation_depth - weak_checkpoints_count;
+
+        let confirmations_to = strong_confirmation_depth
+            .checked_add(historical_checkpoints_count)
+            .ok_or(BitcoinCheckpointError::ChainParamsTooLarge {
+                strong_confirmation_depth,
+                historical_checkpoints_count,
+                weak_checkpoints_count,
```
SatwikPrabhu, line 75: Nit: Do we need to display `weak_checkpoints_count` in this error variant?

crates/consensus/authority/src/bitcoin_checkpoint/chain.rs
```diff
@@ -0,0 +1,729 @@
...
+            })?;
+
+        let confirmation_window_size = confirmations_to
+            .checked_sub(confirmations_from) // width of the window
+            .and_then(|v| v.checked_add(1)) // +1 for inclusive range
+            .ok_or(BitcoinCheckpointError::ChainParamsTooLarge {
+                strong_confirmation_depth,
+                historical_checkpoints_count,
+                weak_checkpoints_count,
```
SatwikPrabhu, line 84: As far as I understand, the window size does not depend on `strong_confirmation_depth` right?

SatwikPrabhu:
As mentioned by @lamafab , very clean and clear code. Enjoyed reviewing this PR 🔥 

Left a couple of nitpicks from my side in addition to the comments left by @lamafab. Will do another rereview to see if I have missed anything!

bin/reth/src/commands/poa/mod.rs
shumkov: Do we expect fed nodes to have such a big difference in bitcoin tip view?

crates/consensus/authority/src/bitcoin_checkpoint/syncer.rs
shumkov: Yeah, let's do it. Technically, this has never happened while we aren't calling the `sync` method twice.

crates/consensus/authority/src/bitcoin_checkpoint/syncer.rs
shumkov: Make sense. We need to clean up the chain and fill it immediately.

crates/consensus/authority/src/comet_bft/abci.rs
shumkov: I will change wording

crates/consensus/authority/src/comet_bft/abci.rs
shumkov: This is a test.

crates/consensus/authority/src/bitcoin_checkpoint/chain.rs
shumkov: You are right, we need to clean up and refill the chain from scratch.

crates/consensus/authority/src/bitcoin_checkpoint/chain.rs
shumkov: Yep, it also participates in overflow.

crates/consensus/authority/src/bitcoin_checkpoint/chain.rs
shumkov: Nope, but it moves the window closer to overflowing in other methods.

bin/reth/src/commands/poa/mod.rs
scottmillner: im fine with it hardcoded in the binary
i would not expect such a big difference so don't think 18 is necessary

scottmillner:
approach ack
will re-review once comments addressed

and thanks for taking this on!

crates/consensus/authority/src/bitcoin_checkpoint/chain.rs
shumkov: Shouldn't be possible due to check above:

```
        // Confirmation depth must belong to the configured window
        if !self.confirmation_window.contains(&depth) {
            return None;
        }
```

crates/consensus/authority/src/bitcoin_checkpoint/chain.rs
shumkov: You don't like using `matches!`? What do you recommend?

P.S. I prefer to use `asset_matches!` but it's still experimental. I could add a crate that implements it for the current version of Rust.

crates/consensus/authority/src/bitcoin_checkpoint/chain.rs
shumkov: I used `try_new` because I thought it's a part of the naming convention: `AuthorityConsensusBuilder::try_new`. In my opinion, it looks more idiomatic.

crates/consensus/authority/src/bitcoin_checkpoint/chain.rs
shumkov: It comes from the original implementation, where we have `chain.parent_confirmation_depth` as a depth we consider a strong confirmation depth. We have it in the chain as `strong_confirmation_depth`, so we can get a preferred checkpoint using the `strong` method to propose. On the other side, for checkpoint validation, we would like to have historical checkpoints in case a validating node is ahead in the sync process. There is another case when the validating node is behind, and we can define some number of checkpoints that aren't strong yet but still okayish for validation. To highlight this difference in the logic in code, I've introduced `historical_checkpoints_count` and `weak_checkpoints_count` params. The first is a number of checkpoints after strong depth, and the second is a number of checkpoints before the strong depth. Using the range is quite convenient because we can achieve full visibility on what confirmation depth of checkpoints in the chain are supposed to be, and it is heavily used in code to iterate through confirmations, check if depth is part of the chain, get the lowest and highest depth. You can't always rely on `checkpoints.len()` since it represents the current state, which is not always what we need, especially when the checkpoints chain is partially filled.

SatwikPrabhu (dismissed):

scottmillner:
i know `test_pegin_v1` is flaky but it should eventually pass
 it is failing repeatedly so either there is a bug or the test needs to be refactored
 
 most likely the test needs to be refactored:
 - i'll get to this asap but will hold off on merging until test passes
 - or ill just test everything locally and merge with a note that test will fail until i fix it

crates/consensus/authority/src/bitcoin_checkpoint/chain.rs
lamafab: It's fine, that `if` condition in there just looks a little weird to me. But that's just a nitpick.

bin/reth/src/commands/poa/mod.rs
```diff
@@ -438,62 +434,24 @@ impl<Ext: clap::Args + fmt::Debug> PoaNodeCommand<Ext> {
...
-                        last_tip = tip_hash;
-                    }
-                    tokio::time::sleep(SLEEP).await;
-                }
-            }),
+        let bitcoin_checkpoints = Arc::new(BitcoinCheckpointsChain::try_new(
+            chain.bitcoin_checkpoint_confirmation_depth as usize,
+            chain.historical_bitcoin_checkpoints_count,
+            chain.weak_bitcoin_checkpoints_count,
+        )?);
```
lamafab, line 498: I'm not quite sure how the `ChainSpec` update/migration must be handled here (cc @scottmillner)?

lamafab (approved):
Looks great! Thanks for this.

lamafab:
Looks like the CI test keeps timing out.

bin/reth/src/commands/poa/mod.rs
scottmillner: updating the chainspec is fine since it doesn't impact the genesis hash: nothing in the block header has changed

i recently updated the chainspec with the `lst_fee_receiver` and there were no impacts

it's when we change the minting contract or the pre-allocations to accounts (ie hot wallet for pegins) that the genesis hash changes

scottmillner:
when running locally,  i repeatedly hit:
`A proposed bitcoin checkpoint is not a part of local checkpoint chain.`

i think this is bc we have `historical_bitcoin_checkpoints_count` to 0 for testnet

I also occasionally hit: `MissingBitcoinCheckpoint` in `prepare_proposal`

this seems to be less stable then the current implementation (at least on testnet):

@shumkov can you test this all out locally and confirm behavior
i really appreciate the work and nice clean code but excluding from mainnet release for now
until can confirm expected behavior

rwlockbg (dismissed):
Very clean and nice code. Thanks for doing this @shumkov. Approved on my end, nice work!

lamafab (dismissed):

shumkov:
@coderabbitai review

crates/consensus/authority/src/bitcoin_checkpoint/syncer.rs
```diff
@@ -0,0 +1,522 @@
...
+            // Sync bitcoin checkpoints to the tip in a blocking task
+            // because we are using block IO operations in bitcoind RPC client
+            let syncer_lock_clone = Arc::clone(&syncer_lock);
+            let result = tokio::task::spawn_blocking(move || {
+                let mut syncer = syncer_lock_clone.blocking_lock();
+                syncer.sync_new_blocks()
+            })
+            .await
+            .expect("spawned blocking task failed to sync bitcoin checkpoints");
+
```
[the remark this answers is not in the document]
shumkov: This will be logged by panic hook

crates/consensus/authority/src/bitcoin_checkpoint/chain.rs
```diff
@@ -0,0 +1,764 @@
...
+        // Remove the oldest checkpoint if we exceed the size limit
+        if new_checkpoints.len() == self.confirmation_window_size {
+            new_checkpoints.pop_front();
+        }
+
+        new_checkpoints.push_back(checkpoint);
+
+        self.checkpoints.store(Arc::new(new_checkpoints));
+
+        Ok(())
```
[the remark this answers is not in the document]
shumkov: we do not call this in parallel

crates/chainspec/src/spec.rs
```diff
@@ -193,7 +201,9 @@ pub static BOTANIX_MAINNET: Lazy<Arc<ChainSpec>> = Lazy::new(|| {
         paris_block_and_final_difficulty: Some((0, U256::from(0))),
         hardforks: EthereumHardfork::botanix().into(),
         deposit_contract: None, // only relevant for PoS chains
-        parent_confirmation_depth: 18,
+        bitcoin_checkpoint_confirmation_depth: 18,
+        weak_bitcoin_checkpoints_count: 1,
+        historical_bitcoin_checkpoints_count: 1,
         leader_selection_window: Some(20),
```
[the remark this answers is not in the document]
shumkov: They are

crates/consensus/authority/src/bitcoin_checkpoint/stream.rs
```diff
@@ -0,0 +1,243 @@
...
+/// A stream for testing or simulation that emits dummy Bitcoin block hash messages at regular time intervals.
+///
+/// This stream mimics a ZMQ block notification feed by periodically yielding
+/// dummy `SocketMessage::Message(Message::HashBlock)` items with a placeholder (all zeros) block hash.
+/// It can be used in tests or in scenarios where a real ZMQ connection is not available.
+///
+/// - The first message is emitted immediately upon the first poll.
+/// - Each subsequent message is emitted after the configured `interval` elapses.
+/// - Can be rapidly polled (even with a zero interval) for high-frequency message generation.
+pub struct DummyHashBlockStream {
```
rwlockbg, line 20: Why dont you put this under test_utils folder inside bitcoin_checkpoint ? It is a dummy stream but it stays along with the other non-dummy files.

crates/consensus/authority/src/bitcoin_checkpoint/syncer.rs
```diff
@@ -0,0 +1,920 @@
...
+        // Sync at the start to ensure we have the latest checkpoints
+        tracing::debug!("Syncing bitcoin checkpoints at the start");
+        trigger_checkpoints_sync(tx.clone());
+
+        while rx.recv().await.is_some() {
+            let syncer_lock_clone = Arc::clone(&syncer_lock);
+
+            let result = tokio::task::spawn_blocking(move || {
+                let mut syncer = syncer_lock_clone.blocking_lock();
```
rwlockbg, line 261: blocking_lock is bad. Why cannot you just async await it inside (just spawn) and as you are doing now await the future to finish?

rwlockbg: The tokio mutex is async/await btw.

shumkov: Previously, we ran the blocking JSON RPC block sync process right in the tokio main thread. It blocked normal async tasks. Now we spawn a dedicated blocking thread and use a blocking lock inside. This lock doesn't affect anything.

crates/consensus/authority/src/bitcoin_checkpoint/stream.rs
shumkov: We use this stream in case if zmq socket is not provided. So it's not for tests. It's a workaround until we make zmq socket mandatory (since v2).

scottmillner (approved):
🔥 🔥 🔥

SatwikPrabhu (approved):
🔥

crates/consensus/authority/src/bitcoin_checkpoint/stream.rs
rwlockbg: ok, understood.

crates/consensus/authority/src/bitcoin_checkpoint/syncer.rs
rwlockbg: Fair enough, its not an issue for me. I understand the idea behind it. Was just double-checking. Thanks

rwlockbg:
Nice implementation! Thanks @shumkov

rwlockbg (approved):

PR #709, opened 2025-05-06
fix(frost_task): handle unwraps in `DkgRunnerTask` & set default timeout

lamafab:
Addressing some small (but somewhat critical) issues as addressed by @SatwikPrabhu: https://github.com/botanix-labs/Macbeth/pull/663#discussion_r2074096715

Note that I'll also have to add a `frost_id -> secp256k1::PublicKey` lookup to the `FrostTask`, in order to make Marks' life easier. Will submit that tomorrow.

### DIFF
--- a/crates/consensus/authority/src/frost_task.rs
+++ b/crates/consensus/authority/src/frost_task.rs
@@ -694,11 +694,13 @@ where
         tx
     }
     async fn run(mut self) {
+        const DEFAULT_TIMEOUT: Duration = Duration::from_secs(10);
+
         // On startup, we call the btc-server immediately to get the initial
         // payloads. Only the coordinator will have something to send at this
         // point, while non-coordinators wait for the coordinators first message
         // before any messages get sent.
-        let mut timeout = std::time::Duration::from_millis(0);
+        let mut timeout = Duration::from_millis(0);
 
         loop {
             match tokio::time::timeout(timeout, self.rx.recv()).await {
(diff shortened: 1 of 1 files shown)


crates/consensus/authority/src/frost_task.rs
```diff
@@ -708,14 +710,22 @@ where
                         payload: dkg.data,
                     };
 
-                    let resp = self.btc_server.new_dkg_payload(req).await.unwrap();
+                    let resp = match self.btc_server.new_dkg_payload(req).await {
+                        Ok(r) => r,
+                        Err(err) => {
+                            timeout = Duration::from_secs(10);
```
SatwikPrabhu: Maybe this can be a predefined constant?

crates/consensus/authority/src/frost_task.rs
```diff
@@ -734,7 +744,14 @@ where
                 Err(_) => {
                     warn!(target: "consensus::authority::frost_task::DkgRunnerTask", "DKG timeout triggered");
 
-                    let resp = self.btc_server.get_dkg_payloads(client::Empty {}).await.unwrap();
+                    let resp = match self.btc_server.get_dkg_payloads(client::Empty {}).await {
+                        Ok(r) => r,
+                        Err(err) => {
+                            timeout = Duration::from_secs(10);
```
SatwikPrabhu: Same here as well

scottmillner (approved):

PR #710, opened 2025-05-06
fix(abci): potential panic if txs are exceeding `max_tx_bytes`

shumkov:
> Requirement 2 [PrepareProposal, tx-size]: When p’s Application calls PrepareProposal, the total size in bytes of the transactions returned does not exceed PrepareProposalRequest.max_tx_bytes.

https://docs.cometbft.com/v1.0/spec/abci/abci++_app_requirements

> `[mempool]` Application can now set `ConsensusParams.Block.MaxBytes` to -1
>  to have visibility on all transactions in the
>  mempool at `PrepareProposal` time.
>  This means that the total size of transactions sent via `RequestPrepareProposal`
>  might exceed `RequestPrepareProposal.max_tx_bytes`.
>  If that is the case, the application MUST make sure that the total size of transactions
>  returned in `ResponsePrepareProposal.txs` does not exceed `RequestPrepareProposal.max_tx_bytes`,
>  otherwise CometBFT will panic.

https://github.com/cometbft/cometbft/blob/main/CHANGELOG.md?plain=1#L803


crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -1196,7 +1201,14 @@ where
     fn prepare_proposal(&self, request: RequestPrepareProposal) -> ResponsePrepareProposal {
         debug!("prepare_proposal request: {:?}", request);
         info!("prepare_proposal request for height: {:?}", request.height);
-        let _txs_bytes = request.txs;
+
+        if !request.txs.is_empty() {
+            panic!("Transactions are not expected from CometBFT mempool");
```
arminsabouri, line 1201: Do we want to panic or just log a warning? If someone POSTs a tx to the comet BFT would it end up here?

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -1196,7 +1201,14 @@ where
...
+        }
+
+        let Ok(max_tx_bytes) = request.max_tx_bytes.try_into() else {
+            panic!("Invalid request proposal max_tx_bytes value: {}", request.max_tx_bytes);
```
arminsabouri: Same idea here

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -1207,13 +1219,28 @@ where
...
+        // NDD goes to a block as the first transaction
+        // so we need to take into account its size
+
+        let non_deterministic_data_bytes_len = non_deterministic_data_bytes.len();
+        if non_deterministic_data_bytes_len > max_tx_bytes {
+            // We should panic bc there is a critical bug and there should be a chain halt.
+            panic!(
+                "Non-deterministic data size: {} exceeds the max tx bytes allowed size {}",
+                non_deterministic_data_bytes_len, max_tx_bytes
+            );
```
arminsabouri, line 1225: This node could be doing other useful things (FROST signing, RPC server, signing other blocks, etc...) but for some critical reason as you pointed out its not creating correct block templates. This definetly should be getting logged as an error. I am not sure if panicking is the right approach here.

crates/consensus/authority/src/comet_bft/abci.rs
shumkov: We've disabled mempool in CometBFT a few PR ago so it must be incorrect behaviour only: some bug or misconfiguration.

crates/consensus/authority/src/comet_bft/abci.rs
shumkov: Some, if everything is correct this shouldn't be reachable.

crates/consensus/authority/src/comet_bft/abci.rs
shumkov: I moved this condition without any changes. We panic here for the same reason, it could happen due to change in code or configuration file. The system won't work this way. In my opinion, this is exactly a case for panic, so developer sees it immediately. It shouldn't come through review and tests to production.

I see your point as well to make system more modular and fault-tolerant. We could return errors and in this case the poa will be living with screaming logs, and CometBFT restarting. This is probably better in case if this error leaks to production.

crates/consensus/authority/src/comet_bft/abci.rs
arminsabouri: But is it possible some other federation member may have not disabled their mempool

crates/ethereum/payload/src/lib.rs
```diff
@@ -291,6 +291,9 @@ where
         initialized_block_env.get_blob_gasprice().map(|gasprice| gasprice as u64),
     ));
 
+    // TODO: Shall we use no_updates() to freeze the view of the mempool
+    //  so new arrivals don’t change the order mid-build. This ensures stable iteration and block creation time.
```
scottmillner: makes sense to me be to use `no_updates()`

scottmillner:
utAck
will re-review when out of draft and ci jobs are green

as far as panics:
- the panics make sense to me sinse i'd want the node to shutdown and we see it immediately
- but also see the viewpoint of returning an error and being loud in the logs

so i'm fine either way

arminsabouri:
> utAck will re-review when out of draft and ci jobs are green
> 
> as far as panics:
> 
> * the panics make sense to me sinse i'd want the node to shutdown and we see it immediately
> * but also see the viewpoint of returning an error and being loud in the logs
> 
> so i'm fine either way

Regarding Panics: There are a two main considerations:
* Is this a terminal state? As in, is there no way to recover from this with out human intervention. regarding being notified, we can find out about this error state if we just log:error.
* Is the node useful for other tasks while in this state? Maybe its not proposing blocks but is it signing for FROST payloads or validating other block proposals?

crates/consensus/authority/src/comet_bft/abci.rs
shumkov: That's right. What would be the most visible signal to him/her? The service should not start or fail immediately, I guess?

crates/payload/basic/src/lib.rs
```diff
@@ -384,6 +384,7 @@ where
                 config: payload_config,
                 cancel,
                 best_payload,
+                max_tx_bytes: None,
```
SatwikPrabhu, line 387: Maybe I'm missing something, but doesnt passing None here lead to no limit being applied?

crates/payload/basic/src/lib.rs
```diff
@@ -505,6 +506,7 @@ where
                 config: self.config.clone(),
                 cancel: Cancelled::default(),
                 best_payload: None,
+                max_tx_bytes: None,
```
SatwikPrabhu, line 509: Same as https://github.com/botanix-labs/Macbeth/pull/710/files#r2078360455

scottmillner:
looks like a unit test is failing
ill take a look if i have time

fixed: just needed to set `max_tx_bytes` in the request in the test otherwise defaults to 0

crates/payload/basic/src/lib.rs
scottmillner: Yes but other parts of reth use `BuildArguments` and we don't want to change that behavior so need to pass None.

We only care about `max_tx_bytes` when getting the payload in `prepare_proposal`

crates/payload/basic/src/lib.rs
scottmillner: see comment above

scottmillner (dismissed):

scottmillner (approved):

PR #711, opened 2025-05-06
pref(abci): unnecessary read lock on the process proposal

shumkov:
We lock storage for read before we check the aggregate public key and open read lock twice in `self.aggregate_public_key()`

### DIFF
--- a/crates/consensus/authority/src/comet_bft/abci.rs
+++ b/crates/consensus/authority/src/comet_bft/abci.rs
@@ -757,7 +757,9 @@ where
                     Ok(snapshot_sync_state_lock) => snapshot_sync_state_lock,
                     Err(e) => {
                         error!("Error getting a snapshot state lock: {:?}", e);
-                        return ResponseOfferSnapshot { result: SnapshotOfferResult::Reject as i32 };
+                        return ResponseOfferSnapshot {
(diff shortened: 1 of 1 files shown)


crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -804,7 +806,9 @@ where
                     Ok(snapshot_sync_state_lock_height) => snapshot_sync_state_lock_height,
                     Err(e) => {
                         error!("Error getting a snapshot state lock: {:?}", e);
-                        return ResponseOfferSnapshot { result: SnapshotOfferResult::Reject as i32 };
+                        return ResponseOfferSnapshot {
```
arminsabouri, line 809: Is this just a result of running make fmt

arminsabouri (approved):
Lgtm. Please in the future try not to include formatting changes in the same commit. Thanks

crates/consensus/authority/src/comet_bft/abci.rs
shumkov: It seems my formatting in IDE thinks differently ;)

shumkov:
Absolutely agree. I will revert.

rwlockbg (approved):

scottmillner (approved):

PR #712, opened 2025-05-06
fix(coin_selection): use Segwit satisfaction weight instead of Taproot.

lamafab:
We rely on the `bdk_wallet` [coin selection](https://docs.rs/bdk/latest/bdk/wallet/coin_selection/struct.BranchAndBoundCoinSelection.html#method.coin_select) mechanism to pick to necessary amount of UTXOs (inputs) for our transaction. But before we pass on the UTXO, we prepare the necessary `WeightedUtxo` structure where we use **66wu** (Taproot) for the [satisfaction weight field](https://docs.rs/bdk/latest/bdk/struct.WeightedUtxo.html#structfield.satisfaction_weight). However, all our transactions are Segwit (P2WPKH), hence we should be using **107wu**. Our current implementation severely underestimates the actual transaction weight. 

I believe that this is what is causing that _"min relay fee not met"_ issue we occasionally encounter in `bitcoind`, but we probably need to do further monitoring on whether this is the main cause.

### DIFF
--- a/bin/btc-server/src/util.rs
+++ b/bin/btc-server/src/util.rs
@@ -1422,11 +1422,26 @@ mod tests {
 
     #[test]
     fn test_btc_per_kb_to_sat_per_vb() {
+        // 0.000_5 BTC = 50_000 sats
+        // 50_000 sats/kb = 50 sats/vb
+
         let btc_per_kb =
-            bitcoin::Amount::from_float_in(0.0005, bitcoin::Denomination::Bitcoin).unwrap();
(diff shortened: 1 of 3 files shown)


scottmillner (approved):

SatwikPrabhu (approved):

lamafab:
Btw that const should be called `SEGWIT_SATISFACTION_WEIGHT`, not `SEGWIT_KEYSPEND_SATISFACTION_WEIGHT`, since "keyspend" is a Taproot related concept. We can fix this later :)

PR #714, opened 2025-05-07
Revert "replace Taproot satisfaction weight with Segwit satisfaction …

arminsabouri:
…weight"

This reverts commit 694cf78f0c4e03f8321a54e5919d41798f0374fc.

There was a misunderstanding that our inputs are non-taproot. All of the inputs we create are taproot (segwitv0). And all the spends are taproot key spends which should be using a estimated 66 wu.


lamafab (approved):
Just a note that **pegouts** can be non-Taproot, and most of them probably are P2WPKH. My mistake 😬

scottmillner (approved):

scottmillner:
thanks for catching this!

PR #715, opened 2025-05-07
fix(frost): infinite await on error

shumkov:
Fixes https://github.com/botanix-labs/botanix/issues/1014

### DIFF
--- a/crates/consensus/authority/src/frost_task.rs
+++ b/crates/consensus/authority/src/frost_task.rs
@@ -38,9 +38,10 @@ use reth_provider::{
 };
 use reth_revm::primitives::FixedBytes;
 use tendermint_rpc::client::HttpClient;
-use tokio::sync::mpsc;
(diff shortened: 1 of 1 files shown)


crates/consensus/authority/src/frost_task.rs
```diff
@@ -38,9 +38,10 @@ use reth_provider::{
 };
 use reth_revm::primitives::FixedBytes;
 use tendermint_rpc::client::HttpClient;
-use tokio::sync::mpsc;
+use tokio::sync::mpsc::{self, error::SendError};
 use tracing::{error, info, warn};
 
+// TODO: @rwlock Combine with FrostTaskError?
```
rwlockbg, line 44: @shumkov Good idea, yes!

rwlockbg (approved):

scottmillner (approved):

PR #718, opened 2025-05-07
fix(coordinator): sync blocks before getting pending pegouts for make_tx

scottmillner:
This PR add some additional logging around creating psbts and adding txs once broadcast.
It also syncs the pegout scheduler before the coordinator gets the pending pegouts list. Syncing may add tracked pegouts back to the pending pegouts list so should be done first.


SatwikPrabhu:
Do we also need some logging in the `make_tx` method just in case?

scottmillner:
> Do we also need some logging in the `make_tx` method just in case?

we already have a bunch of debug logs in `make_tx`: we should have good insight across the whole flow now

SatwikPrabhu (approved):

PR #719, opened 2025-05-08
build: consistent rust code format

shumkov:
Now, we are using nightly builds for formatting. Since we all could have a different nightly version installed, we got different formatting, which is overriding from PR to PR.

This PR sets a specific version in `rust-toolchain.toml` and `cargo fmt` uses it.

### DIFF
--- a/.github/actions/setup-rust/action.yaml
+++ b/.github/actions/setup-rust/action.yaml
@@ -1,0 +1,48 @@
+---
+name: "Setup Rust"
+description: "Install Rust toolchain, dependencies, and setup cache"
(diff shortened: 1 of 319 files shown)


scottmillner:
circling back
do we still want this pr?

shumkov:
@scottmillner This PR needs to be fixed - it changes reth formatting which shouldn't be a case. It should be probably done as part of reth upgrade task. What are you thoughts @rwlockbg?

scottmillner:
closing for now but feel free to reopen @shumkov

PR #721, opened 2025-05-08
fixed retrieving chunk by id in load_snapshot_chunk

rwlockbg:


SatwikPrabhu (dismissed):
Code Ack

scottmillner:
just noticed `test_state_sync` is passing although state sync is actually failing for the syncing node:

i noticed when i kept hitting this when running locally:

```
2025-05-08T16:12:35.281262Z  INFO [CometBFT-4] >>>>>> E[2025-05-08|16:12:35.280] appHash verification failed                  module=statesync expected=8D17A74CA4A25F74A4B94E436DB67AEEFF43F549AA5EF9392AB5CC4695E9A175 actual=3797638175875C37CEFA72EF546DB685E43C81BA4AF8238B48A495F98D61588D
```

The issue is we're missing several steps that ABCI Driver does when appending a block so the block never gets committed. We need to add those steps in `apply_snapshot_chunk`. The steps after `db_rw.commit()` in the driver

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -1183,6 +1199,46 @@ where
...
         };
 
+        for sealed_block_with_senders in sealed_blocks_with_senders.into_iter() {
+            let senders = sealed_block_with_senders.senders().unwrap_or_default();
+            let hashed_state = exec_outcome.hash_state_slow();
+            let block_height = sealed_block_with_senders.block.number;
+            let new_header = sealed_block_with_senders.block.header.clone();
+            let sealed_block = sealed_block_with_senders.block.clone();
+
+            let executed_block = ExecutedBlock::new(
```
scottmillner, line 1288: do we need to determine the `exec_outcome` and `hashed_state` for each block, not use those as the result of batched block outcome?

bc `ExecutedBlock::new` is per block not batches

as is, we're saying the result of each individual block is the outcome of the batch of blocks

but since this is just for the in-memory canonical head it doesn't matter since we only care about the final outcome?

rwlockbg: @scottmillner  The exec_outcome is computed using `blocks_with_senders`  (Vec<BlockWithSenders>) which is basically the received serialized+compressed batch of blocks (the chunk). The `batch_execute` executes them all as a batch. In the above-mentioned loop we are just running over each block_with_senders and forming the payload:

```
            let executed_block = ExecutedBlock::new(
                Arc::new(sealed_block),
                Arc::new(senders),
                Arc::new(exec_outcome.clone()),
                Arc::new(hashed_state.clone()),
            );
```
            
which contains the currently iterated block with senders and the exec_outcome of the whole batch. Internally the canonical chain finds the current block exec results inside the exec_outcome and does the state update. This is how I understand it. Might be wrong though...

scottmillner:
the last commit seems to have broken a lot of int tests?

scottmillner:
circling back:
- looks like int tests still failing (timing out)

scottmillner (dismissed):

scottmillner (dismissed):

scottmillner (approved):

PR #722, opened 2025-05-08
improve(coin_selection): consider output fees in UTXO input selection

lamafab:
This is a conservative change to the coin selection process  - which is not very good - which takes output fees into account.

Notably:

* We iterate through all outputs and sum up their weights, then calculate the fee according to the passed-on `fee_rate`. That fee is then passed onto the coin selection library with the `target_amount` (sum of all pegouts). The goal is that enough UTXO's are selected as inputs.

* As of now, the final `absolute_fee` as returned by the `bitcoin` crate does not respect the passed on `fee_rate`. Technically we'd need to adjust the change output such that `absolute_fee = fee_rate * weight(tx)`. This is straight-forward to do, but considered a _steep_ change.

* Recently, the coin selection function has been adjusted to ignore pegouts that overflow - meaning the pegout amount is smaller than the split `fee_per_output`. However, whenever a pegout is dropped, the `absolute_fee` and hence the `fee_per_pegout` MUST be recalculated and each pegout value MUST be updated accordingly. **We do not do this**. This leads to even more imprecision.

* Added basic sanity checks.

I can do a second/better attempt at this later on, but there are other priorities right now.

### DIFF
--- a/bin/btc-server/src/wallet/coin_selection.rs
+++ b/bin/btc-server/src/wallet/coin_selection.rs
@@ -7,7 +7,7 @@ use bdk_wallet::coin_selection::{
 };
 use bitcoin::{
     psbt::{Error as PsbtError, ExtractTxError, Psbt},
-    Amount, FeeRate, OutPoint, ScriptBuf, TxOut,
+    Amount, FeeRate, OutPoint, ScriptBuf, TxOut, Weight,
 };
 use std::collections::HashMap;
 use thiserror::Error;
@@ -75,8 +75,27 @@ pub(crate) fn coin_selection(
         }
     };
 
+    // NOTE (lamafab): The coin selection algorithm that we use does not appear
+    // to consider output weights, nor base weights, at all. We hence compute
+    // the estimated fee for all outputs and make that part of the
+    // `target_amount` later on, with the goal that the coin selection algorithm
+    // includes the necessary amount of inputs to cover the final fee. We just
+    // ignore the base weights.
+
+    // Calculate the weight of all the outputs. Do note that we allow different
+    // transaction variants, be it P2SH, P2WPKH, etc.
+    let estimated_output_weight: Weight =
+        // Apply pegouts.
+        outputs.iter().fold(Weight::ZERO, |acc, (tx_out, _)| acc + tx_out.weight())
+        // Apply change output (P2TR), which _might_ not be set.
+        + Weight::from_wu(172);
+
+    let estimated_output_fee =
+        fee_rate.fee_wu(estimated_output_weight).ok_or(CoinSelectionError::FeeRateOverflow)?;
+
     let coin_select =
         bdk_wallet::coin_selection::BranchAndBoundCoinSelection::new(0, OldestFirstCoinSelection);
+
     let target_amount = outputs.iter().map(|o| o.0.value).sum::<Amount>();
 
     let mut rng = rand::thread_rng();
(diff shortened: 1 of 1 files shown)


arminsabouri:
Please squash [4dcdc48](https://github.com/botanix-labs/Macbeth/pull/722/commits/4dcdc48db26197903cc7ffad5f5aa951e301281a) into the commit that introduced the typo.

scottmillner:
utAck

ill review again in the morning but giving my ✅ but def would wait for other reviews as well

but it makes sense to me so far:
 - including `estimated_output_fee` in `target_amount`
 - and sanity checks are nice 
 - but if we have pegout that overflows we aren't recalculating `fee_per_pegout` and `fee_per_pegout` but we should
 - so there is some imprecision as you noted but this seems like a nice improvement considering the limitations:
   - ie the crate doesn't respect the fee_rate passed

scottmillner (approved):

rwlockbg (approved):

bin/btc-server/src/wallet/coin_selection.rs
```diff
@@ -75,8 +97,27 @@ pub(crate) fn coin_selection(
...
+    // includes the necessary amount of inputs to cover the final fee. We just
+    // ignore the base weights.
+
+    // Calculate the weight of all the outputs. Do note that we allow different
+    // transaction variants, be it P2SH, P2WPKH, etc.
+    let estimated_output_weight: Weight =
+        // Apply pegouts.
+        outputs.iter().fold(Weight::ZERO, |acc, (tx_out, _)| acc + tx_out.weight())
+        // Apply change output (P2TR), which _might_ not be set.
+        + Weight::from_wu(172);
```
arminsabouri, line 113: This seems rather large for change. How was this calculated?The only things getting serialized should be :
* Amount u64
* Var int spk 1 byte
* taproot SPK: 34 bytes

bin/btc-server/src/wallet/coin_selection.rs
```diff
@@ -158,14 +207,51 @@ pub(crate) fn coin_selection(
...
+            }
+            .into());
+        }
 
-    // TODO should check that min relay fee rate is met
-    // TODO should check that none of the outputs are now dust
+        // VALIDATE: Change output value must be equal the total input value
+        // minus the target amount. Notably, the pegouts cover all the fees.
+        if change_available {
+            let change_output = tx.output.last().expect("change output not included");
```
arminsabouri, line 242: Is change always the last output? Other code in this repo searches for the change output instead of hardcoding a index.

arminsabouri:
> The goal is that enough UTXO's are selected as inputs.

I think the changes in this PR (mainly target_amount + fee) would def result in a higher total fee rate and hopefully gets us over the min relay fee issue. But doubling the fee rate in the current code _seems_ like it would have the same affect? Im not sure. Ideally BDK would have a way for us to pass estimated tx weight into coin selection.

Another issue seems to be that bdk is working with incorrect information. i.e  we tell BDK we are creating X sats in outputs but then don't. Intead its X - fee_per_output. My intuition says that BDK would over estiamte in fees so we would be over the min relay fee rate but that doesnt solve out underlying issue. But still. Retrospectively seems odd to me.

bin/btc-server/src/wallet/coin_selection.rs
scottmillner: looking at `fn create_psbt()`: looks like change will always be last ouput:

```
let mut output: Vec<TxOut> = outputs.iter().map(|(out, _)| out).cloned().collect();
    if let Some(change) = change {
        output.push(change);
    }
```

PR #723, opened 2025-05-09
fixed unwraps and added safe dks request message decoding

rwlockbg:
Fixes issue: https://github.com/botanix-labs/botanix/issues/1003#issuecomment-2866606332

### DIFF
--- a/bin/btc-server/src/bin/main.rs
+++ b/bin/btc-server/src/bin/main.rs
@@ -1138,7 +1138,8 @@ where
 
         debug!("Cord Fee rate: {:?}", fee_rate);
 
-        // First sync the pegout scheduler as this may add tracked pegouts back to the pending pegouts list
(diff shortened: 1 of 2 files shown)


arminsabouri (commented):
Would it make sense to unit test this to prevent regression?

rwlockbg:
> Would it make sense to unit test this to prevent regression?

@0xBEEFCAF3 There is a test already: https://github.com/botanix-labs/Macbeth/pull/723/files#diff-1067a83fc721b63f0ad37bcf646babdbf6c89ac55f7537c532326ea6731c0c76L518 I checked it already.

lamafab (approved):
Missed those in the DKG changes, thanks!

SatwikPrabhu (approved):

scottmillner (approved):

PR #724, opened 2025-05-09
added PeerMessageResponse error for frost task in order to drop conne…

rwlockbg:
…ctions

Fixes last comment in : https://github.com/botanix-labs/botanix/issues/1005#issuecomment-2866617592


scottmillner (approved):
nice fix! 🔥

PR #725, opened 2025-05-09
Fix(psbt): insufficient psbt validation

SatwikPrabhu:
Fix for https://github.com/botanix-labs/botanix/issues/1008


crates/consensus/authority/src/utils.rs
```diff
@@ -514,6 +521,46 @@ pub async fn validate_psbt_by_ids(
         )));
     }
 
+    // Verify that there are no duplicate Pegout IDs.
```
scottmillner, line 524: These are the same checks already done in `fn validate_ouputs()` **so let's just call that function here.** That way  the logic is consolidated in a single function. 

The three checks done are:
- same number of elements check
- no duplicate PegoutIds
- only one change output and it's spk is correct

As @lamafab mentioned, I definitely think as a follow up to this pr, we consolidate all of our psbt validation logic in a single place/mod bc it's hard to follow since it's spread out everywhere. 

Putting here for reference, we have these methods:
- `fn validate_psbt_by_ids()`
- `fn validate_psbt_by_output()`
- `fn validate_outputs()`

@SatwikPrabhu this is outside the scope of this pr, just wanted to document in this pr.

scottmillner:
Just want to talk this through @SatwikPrabhu @lamafab 

The key comment on the SigmaPrime finding is:

`enforcing psbt.outputs vec to align perfectly with psbt.unsigned_tx.outputs`
which they say we're not doing but we are, right?

Now with @SatwikPrabhu's check: `psbt.outputs.len() != psbt.unsigned_tx.output.len()`
and your logic starting from comment `// get pegouts from db`
we've accomplished checking the exact alignment?

@lamafab's check gets the PegoutIds from the psbt along with their index, then using that index, checks the same position in `psbt.unsigned_tx.output` and does spk and amount check

SatwikPrabhu:
> The key comment on the SigmaPrime finding is:
> 
> enforcing psbt.outputs vec to align perfectly with psbt.unsigned_tx.outputs
> which they say we're not doing but we are, right?
> 
> Now with @SatwikPrabhu's check: psbt.outputs.len() != psbt.unsigned_tx.output.len()
> and your logic starting from comment // get pegouts from db
> we've accomplished checking the exact alignment?
> 
> @lamafab's check gets the PegoutIds from the psbt along with their index, then using that index, checks the same position in psbt.unsigned_tx.output and does spk and amount check

This checks out according to me. I'll add the change to just call `validate_outputs` instead of doing the validations explicitly in `validate_psbt_by_ids`. I missed that earlier.

bin/btc-server/src/util.rs
```diff
@@ -420,13 +420,28 @@ pub enum ValidateOutputsError {
...
 
 /// Check:
 /// - additional outputs are change outputs
 /// - pegouts have not already been finalized
 /// - there are no duplicate outputs
 pub(crate) fn validate_outputs(psbt: &Psbt, db: &database::Db) -> Result<(), ValidateOutputsError> {
+    // Ensure psbt.outputs and psbt.unsigned_tx.output have the same number of elements.
+    // This is critical to prevent a malicious coordinator from adding arbitrary outputs
+    // to psbt.unsigned_tx.output that are not declared in psbt.outputs.
+    if psbt.outputs.len() != psbt.unsigned_tx.output.len() {
```
shumkov, line 435: We still need to verify the output index order, such as we have it for the frost task.

crates/consensus/authority/src/utils.rs
```diff
@@ -505,6 +505,13 @@ pub async fn validate_psbt_by_ids(
...
     psbt: &Psbt,
 ) -> Result<(), PsbtValidationError> {
+    if psbt.outputs.len() != psbt.unsigned_tx.output.len() {
+        error!(target: "consensus::authority::frost_task::validate_psbt_by_ids", "psbt.outputs length ({}) does not match psbt.unsigned_tx.output length ({})", psbt.outputs.len(), psbt.unsigned_tx.output.len());
+        return Err(PsbtValidationError::FailedToValidatePsbtByIds(String::from(
+            "Mismatch between number of PSBT outputs and unsigned transaction outputs",
+        )));
+    }
+
     let pegout_ids = extract_pegout_ids(psbt);
```
shumkov, line 515: The `extract_pegout_ids` ignores invalid pegout ID serialization.

crates/consensus/authority/src/utils.rs
```diff
@@ -514,6 +521,46 @@ pub async fn validate_psbt_by_ids(
...
+    }
+    
+    // The preceding checks (output length equality, duplicate pegout ID, and change output count)
+    // ensure that every entry in `psbt.unsigned_tx.output` (due to the length check)
+    // is accounted for as either a pegout (validated in the main loop below)
+    // or the single allowed change output. Any other scenario (e.g., undeclared outputs,
+    // too many change outputs) would have triggered an earlier error.
+
     // check if a corresponding output exists in the psbt and is for the right amount
     let fee_per_output = psbt.fee_per_output(pegout_ids.len() as u64)
```
shumkov, line 565: We should pass only pegouts here?

crates/consensus/authority/src/utils.rs
shumkov: This requires quite a lot of refactoring since we need to provide data differently. @lamafab suggested introducing a new function and calling both old and new code.

crates/consensus/authority/src/utils.rs
```diff
@@ -514,6 +521,46 @@ pub async fn validate_psbt_by_ids(
...
+            change_output_count += 1;
+        }
+    }
+
+    if change_output_count > 1 {
+        error!(target: "consensus::authority::frost_task::validate_psbt_by_ids", "Multiple change outputs (non-pegout IDs) found: {}", change_output_count);
+        return Err(PsbtValidationError::FailedToValidatePsbtByIds(String::from(
+            "Multiple change outputs (non-pegout IDs) found in PSBT outputs",
+        )));
+    }
```
shumkov, line 556: we aren't checking against already finalized peg outs

bin/btc-server/src/util.rs
```diff
@@ -420,13 +420,28 @@ pub enum ValidateOutputsError {
...
+        error!(
+            target: "btc_server::util::validate_outputs",
+            "psbt.outputs length ({}) does not match psbt.unsigned_tx.output length ({})", 
+            psbt.outputs.len(), 
+            psbt.unsigned_tx.output.len()
+        );
+        return Err(ValidateOutputsError::OutputCountMismatch(psbt.outputs.len(), psbt.unsigned_tx.output.len()));
+    }
+
```
shumkov, line 444: Make sure we have one and only one change output

crates/consensus/authority/src/utils.rs
shumkov: Yes we should check `change_output_count == 1`

crates/consensus/authority/src/utils.rs
shumkov: Yes, that's correct.

crates/consensus/authority/src/utils.rs
shumkov: We decided to ignore for now

bin/test-suite/src/mint_attack_contract_abi.rs
lamafab, line 1: Probably revert this.

bin/test-suite/src/multi_mint_helper_abi.rs
lamafab, line 1: Probably revert this.

lamafab (dismissed):

crates/consensus/authority/src/utils.rs
lamafab: Every transaction having a change output is very likely, but not a requirement (or necessity).

shumkov: Yesterday we discussed that it must be present. @scottmillner guys please decide.

rwlockbg (dismissed):

SatwikPrabhu:
@shumkov @scottmillner @lamafab I think I have addressed the following comments and added unit tests for the same.
https://github.com/botanix-labs/Macbeth/pull/725#discussion_r2084989776
https://github.com/botanix-labs/Macbeth/pull/725#discussion_r2084487690
https://github.com/botanix-labs/Macbeth/pull/725#discussion_r2084485233 --> Currently I'm only checking the script pubkey. Checking the amount involved some fee calculations which I felt could be done in another PR.

As for https://github.com/botanix-labs/Macbeth/pull/725#discussion_r2084974019 --> I think this would be great to add but I wasnt able to figure out an easy way to do this here. The only way to get the finalized pegouts in the frost task seemed to be getting a stream of chunks of pegout ids via `btc_server.get_finalized_pegout_ids()` and this seemed a bit of an over-optimization for this PR as we do this check more efficiently in `validate_outputs` anyway. Let me know your thoughts..

Apologies for the delay and if some of the code here seems a bit messy and any badly written code comments/debug prints and some failed int tests. I will clean this up tomorrow morning. But in the meanwhile would love to get your reviews. In any case, I don't think the things not addressed in this PR (like the amount check and the finalized pegouts check) should block merging this PR before the code freeze as they can be addressed later.

SatwikPrabhu:
The int test failures seem to indicate that my new logic of ensuring the order of the pegout ids to match those seen in the db is problematic as the coordinator clears the pending pegouts from the db `get_round2_signing_package`. Not sure how to go about this issue here..

crates/consensus/authority/src/utils.rs
lamafab: In that case, our `fn coin_selection` mechanism in `bin/btc-server/src/wallet/coin_selection.rs` must be adjusted:

```rust
// Try once with finalized, then add pending and try again.
let selection = coin_select
    .coin_select(
        required_utxos.values().map(to_bdk).collect::<Vec<_>>(),
        available_utxos.values().map(to_bdk).collect::<Vec<_>>(),
        fee_rate,
        target_amount,
        &change_script, // drain_script
        &mut rng,
    )
    .map_err(CoinSelectionError::CoinSelectionBdk)?;

// ...

let change: Option<TxOut> = match selection.excess {
    bdk_wallet::coin_selection::Excess::Change { amount, .. } => {
        Some(TxOut { script_pubkey: change_script.clone(), value: amount })
    }
    bdk_wallet::coin_selection::Excess::NoChange { .. } => None,
};
```

shumkov: @SatwikPrabhu you can use `db.finalized_pegout_ids.get(pegout_id)`. I added a method to `Db` for that in my refactoring PR. I believe you can do it as well

crates/consensus/authority/src/utils.rs
```diff
@@ -434,18 +434,39 @@ pub enum PsbtValidationError {
...
+    /// Invalid pegout ID serialization
+    InvalidPegoutIdSerialization(usize),
 }
 
 /// Extract pegouts ids from psbt
-pub fn extract_pegout_ids(psbt: &Psbt) -> Vec<(usize, PegoutId)> {
+pub fn extract_pegout_ids(psbt: &Psbt) -> Result<Vec<(usize, PegoutId)>, PsbtValidationError> {
+    let pegout_id_key_prefix = vec![0xFC]; // Proprietary prefix used by btcserverlib
+    let pegout_id_key_subtype = 0x07; // Subtype for pegout_id
+    let pegout_id_key_data = vec![b'p', b'e', b'g', b'o', b'u', b't']; // Key data itself
```
lamafab: Those should probably be constants? Statically initialized, at least.

crates/consensus/authority/src/utils.rs
```diff
@@ -434,18 +434,39 @@ pub enum PsbtValidationError {
...
+            // Iterate through proprietary key-value pairs
+            output.proprietary.iter().find_map(|(key, value_bytes)| {
+                // Check if the key matches the expected pegout_id key structure
+                if key.prefix == pegout_id_key_prefix &&
+                   key.subtype == pegout_id_key_subtype &&
+                   key.key == pegout_id_key_data
+                {
+                    // If key matches, attempt to parse the value bytes
+                    match PegoutId::from_bytes(value_bytes.as_slice()) {
+                        Ok(id) => Some(Ok((pos, id))), // Success: Wrap in Ok(Ok(...))
```
lamafab: > // Success: Wrap in Ok(Ok(...))

Nitpick: `Some(Ok(...))`

crates/consensus/authority/src/utils.rs
```diff
@@ -505,24 +530,80 @@ pub async fn validate_psbt_by_ids(
...
+
+    for i in 0..psbt.outputs.len() {
+        if !validated_pegout_indices.contains(&i) {
+            change_output_count += 1;
+        }
+    }
+    eprintln!("[validate_psbt_by_ids] Validated pegout indices: {:?}, Change output count: {}", validated_pegout_indices, change_output_count);
+
+    // Check if the number of change outputs is exactly 1
+    if change_output_count != 1 {
```
lamafab: As clarified in the comments, right now we do construct transactions where a change-output might not be present.

https://github.com/botanix-labs/Macbeth/pull/725/files#r2086709769

bin/btc-server/src/util.rs
```diff
@@ -418,67 +418,128 @@ pub enum ValidateOutputsError {
...
+            }
         }
     }
 
-    // check for duplicate outputs by pegout ids
-    let unique_pegout_ids: HashSet<PegoutId> = psbt_pegout_ids.iter().cloned().collect();
-    if unique_pegout_ids.len() != psbt_pegout_ids.len() {
-        return Err(ValidateOutputsError::DuplicateOutputs);
+    // Check if exactly one change output was found
+    if change_output_identified_idx.is_none() {
```
lamafab: Same here: https://github.com/botanix-labs/Macbeth/pull/725/files#r2086709769

bin/btc-server/src/util.rs
```diff
@@ -418,67 +418,128 @@ pub enum ValidateOutputsError {
...
-
-        let has_correct_change = change_output.script_pubkey == expected_script_pubkey;
-        if !has_correct_change {
-            return Err(ValidateOutputsError::InvalidChangeOutput);
-        }
+    // Change Output Script Validation (Now uses unwrap, assuming the check above passes)
+    let change_idx = change_output_identified_idx.unwrap(); // Use unwrap safely after the check
+    let agg_pk = public_key_package.verifying_key().to_secp_pk()
+        .expect("Failed to convert verifying key to secp pk for change output validation");
+    let expected_change_script = generate_taproot_change_scriptpubkey(&agg_pk);
```
lamafab: Generally speaking; considering that the aggregated public key is fixed long-term, we should consider computing that `expected_change_script` once on startup, instead of doing it every single time this function is called. But this does not need to be in this PR, of course.

bin/btc-server/src/util.rs
```diff
@@ -418,67 +418,128 @@ pub enum ValidateOutputsError {
...
+                    })?;
+                
+                let expected_script_pubkey = stored_pegout_request.spk; 
+
+                // Get Actual Transaction Output
+                let actual_tx_out = psbt.unsigned_tx.output.get(idx)
+                    .expect("Index should be valid due to earlier length check"); // .expect is safe here
+
+                // Order & Script Match
+                if actual_tx_out.script_pubkey != expected_script_pubkey {
```
lamafab: Additionally to checking the destination, we should also check that `actual_tx_out.value == stored_pegout_request.value`?

lamafab (changes requested):
Looks (mostly) good! 🚀

bin/btc-server/src/util.rs
lamafab: The value of the change-output should also be validated.

SatwikPrabhu:
As per discussion with @scottmillner and @shumkov , we want to put the changes of the second commit aside for after mainnet due to complexities involved in sharing L2 data to the btc_server and the btc_server db access to the frost task. So reverted that commit.

crates/consensus/authority/src/utils.rs
scottmillner: this will be done in a separate pr
resolving comment

crates/consensus/authority/src/utils.rs
scottmillner: per our offline discussion:
- btc-server does not have access to the the L2 db
- a grpc endpoint would need to be added to communicate with the reth node

this will be done in a separate pr

scottmillner (approved):

PR #726, opened 2025-05-09
set limits for finalized pegout ids to 3 months age

rwlockbg:
The idea behind this PR is to introduce checks both on the signing side but also inside the btc-server to prevent signing finalized pegout ids which are older than 3 months - the age cutoff has already been agreed by Willem and Scott. I have introduced botanix eth client inside the btc-server which checks if the pegout id is inside a block older than 3 months and if so fails to sign it. Same is implemented inside the wallet state sync whereby received pegout ids are being tested for maturity and if older than 3 months are not being persisted neither sent over the wire.

## Summary by CodeRabbit

- **New Features**
  - Added timestamp fields to pegout-related data, enabling tracking of pegout request and finalization times.
  - Pegout messages now include timestamp metadata for improved transparency.
  - Introduced block age validation to ensure pegouts are within an acceptable time range.

- **Bug Fixes**
  - Enhanced backward compatibility by making new timestamp fields optional and supporting deserialization of older data formats.

- **Refactor**
  - Converted several asynchronous functions to synchronous for streamlined processing.
  - Renamed internal fields for clarity and consistency (e.g., `port` to `btc_server_port`).
  - Improved code formatting, style, and logging across modules.

- **Chores**
  - Updated dependencies and cleaned build configurations.
  - Expanded tests to cover timestamp handling, pegout validation errors, and serialization compatibility.


crates/consensus/authority/src/utils.rs
```diff
@@ -286,10 +309,18 @@ pub(crate) async fn get_block_pegouts(
     block: u64,
     client: &impl BlockReaderIdExt,
     btc_network: bitcoin::Network,
+    max_cutoff_age: Option<Duration>,
```
scottmillner, line 380: nit > update method docs with `max_cutoff_age` arg

scottmillner:
Thanks for adding this!

We need to add the check when signing psbts as well:
- Signers will reject the psbt if it contains any pegout ids that are older than 3 months
- This logic will be inside `fn validate_outputs()`

If we don't have this check, then we allow pegouts greater than 3 months without checking them

And we need to prune the sled db of finalized pegouts older than 3 months since this is a sliding window.
Otherwise, we haven't solved the sled memory issue with storing an unbounded list.

scottmillner:
@PhoenixCodeCraft This pr is related to an offline discussion we had around the btc-server db limitiations
- The finalized pegouts list is unbounded and would eventually cause a memory issue with the btc-server sled db
- This pr enforces:
  - btc-server only stores finalized ouputs for the last 3 months to prevent db memory issue
  - Therefore, signers will only sign psbt less than 3 months old so they can cross-check against finalized outputs list

So any pegout older than 3 months will not be honored but not sure how we'd even get into that scenario.  That would mean somehow our system wasn't able to honor that pegout within 3 months after it was burned on the L2.

This pr isn't a blocker for mainnet release but needs to be released within a reasonable time (before we hit a btc-server memory issue which should take some time).

PhoenixCodeCraft:
Ack, do we have an estimate of the size of the btc server sled db?

crates/consensus/authority/Cargo.toml
```diff
@@ -28,8 +28,9 @@ frost-secp256k1-tr = { workspace = true }
 futures = { workspace = true }
 # async
 futures-util = { workspace = true }
-
 hex = "0.4.3"
+
+lazy_static = "1.4.0"
```
shumkov: You can use `LazyCell` from std.

crates/consensus/authority/src/utils.rs
```diff
@@ -271,6 +274,26 @@ pub(crate) async fn epoch_pegouts(
...
+/// # Arguments
+///
+/// * `timestamp` - The timestamp of the block in seconds since the UNIX epoch.
+/// * `max_age_cutoff` - The maximum acceptable age of the block as a `Duration`.
+///
+/// # Returns
+///
+/// Returns `true` if the block's age is less than the specified max cutoff age, otherwise `false`.
+pub fn is_block_age_acceptable(timestamp: u64, max_age_cutoff: Duration) -> bool {
+    let now = match SystemTime::now().duration_since(UNIX_EPOCH) {
```
shumkov, line 355: Is it fine that this is non-deterministic?

crates/consensus/authority/src/wallet_state_sync.rs
```diff
@@ -31,11 +38,9 @@ use tokio::sync::{mpsc::error::SendError, RwLock};
 use tracing::{debug, error, info, trace, warn};
 use uuid::Uuid;
 
-use crate::{
-    prost_parser::ProstMessageSerdelizer,
-    utils::{get_block_pegouts, EpochPegoutsError},
-    Storage,
-};
+lazy_static! {
+    static ref MAX_BLOCK_TS_CUTOFF_DURATION: Duration = Duration::from_secs(30 * 24 * 60 * 60 * 3); // 3 months
```
shumkov: We could use `const` as well

crates/consensus/authority/src/utils.rs
rwlockbg: yes, it is only at the point of evaluation.

crates/consensus/authority/Cargo.toml
rwlockbg: True, but reth is consistenly using laz_static so I have just followed the convention.

crates/consensus/authority/src/wallet_state_sync.rs
rwlockbg: I have added a const for the time, but a const for the Duration wont work - it is a runtime init issue, hence the lazy_static

crates/consensus/authority/src/utils.rs
rwlockbg: added

scottmillner:
@rwlockbg looks like the wrong commit was pushed to this branch?
it's the commit for the snapshot manager

crates/consensus/authority/Cargo.toml
shumkov: It would be nice to separate our code from reth so we could use modern approaches.

crates/consensus/authority/src/utils.rs
```diff
@@ -286,10 +310,18 @@ pub(crate) async fn get_block_pegouts(
...
     client: &impl BlockReaderIdExt,
     btc_network: bitcoin::Network,
+    max_cutoff_age: Option<Duration>,
 ) -> Result<Vec<PegoutId>, EpochPegoutsError> {
     let mut pegouts = Vec::new();
     match client.block_by_number(block) {
         Ok(Some(block)) if bloom_contains_pegout(block.header.logs_bloom) => {
+            if !is_block_age_acceptable(
+                block.header.timestamp,
+                max_cutoff_age.unwrap_or(Duration::ZERO),
```
lamafab: Is `Duration::ZERO` the right call? This will invalidate every block, since `now - 0 = now`, and `timestamp > now == false`.

crates/consensus/authority/src/utils.rs
```diff
@@ -286,10 +310,18 @@ pub(crate) async fn get_block_pegouts(
     block: u64,
     client: &impl BlockReaderIdExt,
     btc_network: bitcoin::Network,
+    max_cutoff_age: Option<Duration>,
 ) -> Result<Vec<PegoutId>, EpochPegoutsError> {
     let mut pegouts = Vec::new();
     match client.block_by_number(block) {
         Ok(Some(block)) if bloom_contains_pegout(block.header.logs_bloom) => {
+            if !is_block_age_acceptable(
+                block.header.timestamp,
```
lamafab: Do note that this `block.header.timestamp` is originally constructed in `abci.rs`, and is simply retrieved from the payload (`time`) as submitted by the ABCI-client to `fn process_proposal`.

https://docs.cometbft.com/v0.38/spec/abci/abci++_methods#processproposal

I do not know how this timestamp is selected or verified, and what type of manipulation we open ourselves up here.

rwlockbg: Acc. to the docs it is - A scalar value equal to the reasonable output of Unix’s time() at this block’s inception; formally Hs.
Since we are talking about a 3 months span, the exact precision of the ts does not really matter that much here - imo. Acc. to the def above the blocks inception ts makes total sense for us to measure the time elapsed.

crates/consensus/authority/src/utils.rs
rwlockbg: I have reworked it now so we dont enter edge cases like this - but you are right, I did not try the None variant as we do want to limit to Some(age). Thanks, pls re-review the fix.

rwlockbg:
> @rwlockbg looks like the wrong commit was pushed to this branch? it's the commit for the snapshot manager

Sorry, yes, removed it long ago.

crates/consensus/authority/src/wallet_state_sync.rs
rwlockbg: @shumkov Changed it as you requested using once_cell.

crates/consensus/authority/Cargo.toml
rwlockbg: @shumkov Changed it as you requested using once_cell.

crates/consensus/authority/src/utils.rs
lamafab: I'm not quite sure how reproducible `fn get_block_pegouts` has to be; note that because `now` is purely subjective, there could be edge cases where some validators accept the block and others do not.

lamafab:
So is it expected that this is only enforced in the context of wallet state sync?

scottmillner:
> So is it expected that this is only enforced in the context of wallet state sync?

enforced during wallet state sync and when signing psbts:
- wallet state sync only gossips most recent 3 months finalized pegouts
- and signers will only sign a psbt whose pegouts are less than 3 months olds

scottmillner:
circling back to this comment of mine:
```
And we need to prune the sled db of finalized pegouts older than 3 months since this is a sliding window.
Otherwise, we haven't solved the sled memory issue with storing an unbounded list.
```

this is an important pr but not a blocker for private mainnet imo:
- this could be in the first release after launch

so you don't need to finish this pr today since we still have the cometbft issues

crates/consensus/authority/src/utils.rs
rwlockbg: @lamafab I am afraid that is what we have in terms of timestamp. You ahve other suggestions ?

lamafab: Let's just have the CometBFT super-majority handle this edge case; worst case some proposed blocks do not get backed, and replaced with others.

lamafab:
> and signers will only sign a psbt whose pegouts are less than 3 months olds

NOW it has been added :)

crates/consensus/authority/src/wallet_state_sync.rs
shumkov: Interesting, `Duration::new` marked as `const`

scottmillner:
circling back:
- will re-review when ci/cd is green

bin/btc-server/src/botanix_client.rs
```diff
@@ -0,0 +1,434 @@
...
+    }
+}
+
+#[async_trait::async_trait]
+impl BotanixEthClientTrait for BotanixEthClient {
+    fn http_provider(&self) -> &Provider<Http> {
+        self.http_client.provider()
+    }
+
+    async fn nonce(&self, address: &str) -> U256 {
```
arminsabouri: Can we use a stronger type that str? Fixed 20 Bytes maybe?

bin/btc-server/src/botanix_client.rs
```diff
@@ -0,0 +1,434 @@
...
+        let account =
+            NameOrAddress::Address(ethers::types::Address::from_slice(address.as_bytes()));
+        self.http_client
+            .provider()
+            .get_transaction_count(account, Some(BlockId::Number(BlockNumber::Latest)))
+            .await
+            .expect("nonce to be returned")
+    }
```
arminsabouri: Shouldnt this return a result? For any given string I dont think we can expect a nonce to be returned, right?

bin/btc-server/src/botanix_client.rs
```diff
@@ -0,0 +1,434 @@
...
+pub enum Error {
+    /// Provider error: `{0}`
+    Provider(#[from] ProviderError),
+    /// Url parse error: `{0}`
+    UrlParse(#[from] url::ParseError),
+}
+
+/// Trait defining the interface for Ethereum client operations
+#[async_trait::async_trait]
+pub trait BotanixEthClientTrait: Send + Sync {
```
arminsabouri: It seems like only get_tx_by_hash and get_block_by_id are used. I would trim this to only the methods we need.

arminsabouri (changes requested):

arminsabouri:
This is a large diff to critical code with no PR description. Can we please tag a ticket or write a couple sentences describing what problem we are solving. Thanks

bin/btc-server/src/botanix_client.rs
rwlockbg: This is a replica of the client we have in reth. I will refactor both of them in a separate PR

bin/btc-server/src/botanix_client.rs
rwlockbg: This is a replica of the client we have in reth. I will refactor both of them in a separate PR

bin/btc-server/src/botanix_client.rs
rwlockbg: This is a replica of the client we have in reth. I will refactor both of them in a separate PR

bin/btc-server/src/util.rs
```diff
@@ -451,10 +489,28 @@ pub(crate) fn validate_outputs(psbt: &Psbt, db: &database::Db) -> Result<(), Val
         };
     }
 
-    // check outputs are not in finalized pegouts list
+    // check outputs are not in finalized pegouts list and not too old
     let finalized_pegouts_id =
         db.get_finalized_pegout_ids()?.into_iter().map(|id| id.id).collect::<Vec<_>>();
+
     for id in &psbt_pegout_ids {
```
scottmillner, line 498: lets add a test for `validate_outputs()` that covers the failure path:
- specifically a test that returns `ValidateOutputsError::PegoutIdTooOld`

scottmillner:
this might've gotten lost in all the comments:

```
And we need to prune the sled db of finalized pegouts older than 3 months since this is a sliding window.
Otherwise, we haven't solved the sled memory issue with storing an unbounded list.
```

We still need to do the above since this was the original motivation for this pr.

bin/btc-server/src/util.rs
```diff
@@ -274,11 +298,12 @@ impl PartialEq for ValidatePSBTError {
 /// required signers. `ROUND2`: Checks if there are enough round 2 partial signatures. Ensuring we
 /// never add more than a quorum of signers `ROUND2_TRANSITION`: Validates partial signatures during
 /// the transition to round 2, ensuring signers match and Frost IDs align.
-pub fn validate_psbt(
+pub async fn validate_psbt(
```
scottmillner: nit: update method docs to include arg `botanix_eth_client`

bin/btc-server/src/util.rs
rwlockbg: yes, added

bin/btc-server/src/util.rs
rwlockbg: added

bin/btc-server/src/util.rs
```diff
@@ -466,10 +507,27 @@ pub(crate) fn validate_outputs(psbt: &Psbt, db: &database::Db) -> Result<(), Val
         };
     }
 
-    // check outputs are not in finalized pegouts list
+    // check outputs are not in finalized pegouts list and not too old
     let finalized_pegouts_id =
         db.get_finalized_pegout_ids()?.into_iter().map(|id| id.id).collect::<Vec<_>>();
+
     for id in &psbt_pegout_ids {
+        if botanix_eth_client.is_connected().await {
```
lamafab: Afaik this will just skip the `is_block_age_acceptable` check entirely if the client is not connected? It should return an error instead, or similar.

bin/btc-server/src/util.rs
```diff
@@ -1237,20 +1363,24 @@ mod tests {
 
         // Test the other way around: psbt.outputs is longer
         let mut psbt2 = create_psbt(1, 0, None); // 0 outputs in psbt.outputs
-        // psbt2.unsigned_tx.output is initially empty from create_psbt with 0 outputs
-        // Add one to psbt.outputs to create a mismatch where psbt.outputs is longer
-        psbt2.outputs.push(Default::default()); 
+                                                 // psbt2.unsigned_tx.output is initially empty from create_psbt with 0 outputs
+                                                 // Add one to psbt.outputs to create a mismatch where psbt.outputs is longer
```
lamafab: Something weird going on here?

bin/test-suite/src/mint_attack_contract_abi.rs
lamafab, line 1: Remove this.

bin/test-suite/src/multi_mint_helper_abi.rs
lamafab, line 1: Remove this.

lamafab (commented):
To be totally honest, I'm not sure if it's a good idea to rely on the `BotanixEthClient` - yet another `btc-server` config parameter - to validate pegouts in such a way. If we really want to introduce a cut-off date for pegouts, we should make this information available in the pegout structure itself.

Respectively, the `fn botanix_mint_contract_checks` in the EVM code has easy access to the L2 timestamp and can just append this information to each pegout entry. And instead of relying on a `timestamp` we should probably just use the L2 `height`, and estimate the three month period based on that. This way we also get determinism for free.

EDIT: Some of the works here is also dependent on this: https://github.com/botanix-labs/Macbeth/pull/735 (or vice-versa)

scottmillner:
> To be totally honest, I'm not sure if it's a good idea to rely on the `BotanixEthClient` - yet another `btc-server` config parameter - to validate pegouts in such a way. If we really want to introduce a cut-off date for pegouts, we should make this information available in the pegout structure itself.
> 
> Respectively, the `fn botanix_mint_contract_checks` in the EVM code has easy access to the L2 timestamp and can just append this information to each pegout entry. And instead of relying on a `timestamp` we should probably just use the L2 `height`, and estimate the three month period based on that. This way we also get determinism for free.
> 
> EDIT: Some of the works here is also dependent on this: #735 (or vice-versa)

@rwlockbg i like @lamafab suggestion here:
- we can just add the needed info to the pegout structure and i like using height instead of timestamps
- then we don't need the BotanixEthClient

bin/btc-server/src/util.rs
rwlockbg: @lamafab There is an issue here of a cyclic dependency - you need your btc-server to run first before the core node can start. If I remove the connected part then we would panic / fail here and the btc-server wont start leading to the reth node not being to start. With this "Semi-conservative" check we let the btc-server start first - in this process obviously the above returning not connected and no checks being performed. Once it is however up, then the reth node is able to start and a few secs down the line the connection above is already up and the checks gets triggered. With the current overall architecture there is no other way to get around this issue. I think this is an elegant solution which works pretty  well considering what we have. cc @scottmillner

bin/test-suite/src/mint_attack_contract_abi.rs
rwlockbg: sure

bin/test-suite/src/multi_mint_helper_abi.rs
rwlockbg: sure

rwlockbg:
> > To be totally honest, I'm not sure if it's a good idea to rely on the `BotanixEthClient` - yet another `btc-server` config parameter - to validate pegouts in such a way. If we really want to introduce a cut-off date for pegouts, we should make this information available in the pegout structure itself.
> > Respectively, the `fn botanix_mint_contract_checks` in the EVM code has easy access to the L2 timestamp and can just append this information to each pegout entry. And instead of relying on a `timestamp` we should probably just use the L2 `height`, and estimate the three month period based on that. This way we also get determinism for free.
> > EDIT: Some of the works here is also dependent on this: #735 (or vice-versa)
> 
> @rwlockbg i like @lamafab suggestion here:
> 
> * we can just add the needed info to the pegout structure and i like using height instead of timestamps
> * then we don't need the BotanixEthClient

No, you always need the botanix eth client as this is one of the things that the auditors specifically required from us - L2 checks to be performed live. This is also what we don on the reth side. Anyone can inject any data (block, ts). The whole point of what we are doing is to verify the pegout id is there on the L2 at the moment of it being queried. Plus, I would not touch right now any internal structures or db tables.

bin/btc-server/Cargo.toml
```diff
@@ -50,6 +51,7 @@ log = "0.4.18"
 
 merlin = "3.0.0"
 miniscript = "12.3.0"
+once_cell = "1.21"
```
shumkov: It's a part of the standard library now

bin/test-suite/src/mint_attack_contract_abi.rs
rwlockbg: Cannot remove it as my linter formats it back acc. to the standard. The main has to be changed.

bin/test-suite/src/multi_mint_helper_abi.rs
rwlockbg: Cannot remove it as my linter formats it back acc. to the standard. The main has to be changed.

shumkov:
>  And instead of relying on a timestamp we should probably just use the L2 height, and estimate the three month period based on that. This way we also get determinism for free.

I like the idea of making it deterministic. I think both block time and height will work. The block time will be more precise because there are things are afficting numer of blocks per time range:
- I saw in the logs blocks with transactions processing 7 and 0.3 seconds without them. There might be very different time ranges if we take 100 blocks from different parts of a chain.
- In case of a chain halt, we could have a pause in block heights for 4-12 hours while we fix it. 

But if we are talking about months, it doesn't matter much.

bin/btc-server/src/util.rs
rwlockbg: I dont have this locally. No idea.

rwlockbg:
> > And instead of relying on a timestamp we should probably just use the L2 height, and estimate the three month period based on that. This way we also get determinism for free.
> 
> I like the idea of making it deterministic. I think both block time and height will work. The block time will be more precise because there are things are afficting numer of blocks per time range:
> 
> * I saw in the logs blocks with transactions processing 7 and 0.3 seconds without them. There might be very different time ranges if we take 100 blocks from different parts of a chain.
> * In case of a chain halt, we could have a pause in block heights for 4-12 hours while we fix it.
> 
> But if we are talking about months, it doesn't matter much.

We are talking about 3 months, it does not matter. Even if it deviates by 5 minutes, this should not be an issue.

shumkov:
@coderabbitai review

bin/btc-server/Cargo.toml
```diff
@@ -38,6 +38,7 @@ derive_more = { version = "1.0", features = ["full"] }
 displaydoc = "0.2"
 dotenvy = { version = "0.15.7" }
 env_logger = "0.11.3"
+ethers = "2.0.11"
```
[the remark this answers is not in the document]
rwlockbg: updated the lib

bin/btc-server/src/util.rs
```diff
@@ -466,13 +494,28 @@ pub(crate) fn validate_outputs(psbt: &Psbt, db: &database::Db) -> Result<(), Val
...
+        let finalized_pegouts_id = finalized_pegouts_ids
+            .iter()
+            .find(|(finalized_pegouts_id, _)| finalized_pegouts_id == id)
+            .cloned();
+
+        if finalized_pegouts_id.is_some() {
             return Err(ValidateOutputsError::AlreadyFinalizedPegouts(vec![*id]));
         }
+
+        // if let Some((id, Some(timestamp))) = finalized_pegouts_id {
```
rwlockbg: @scottmillner It makes no sense to have this check here anymore since above we are already returning on too already finalized pegout ids which basically means that the age of those is irrelevant. My problem here is that the normal pegout id we get above from the Psbt do not have a timestamp (these are btclib structures) hence we cannot run this check here really.

scottmillner: yep, we can't do this check here
but we still need to enforce this on the reth side

we could do this in `validate_psbt_by_ids` and make sure the pegout id is within 3 months since we can crosscheck the block timestamp

bin/btc-server/Cargo.toml
rwlockbg: udpated

bin/btc-server/src/util.rs
scottmillner: ive created a feature branch (`feat-branch/truncate-finalized-outputs-latest-3-months`) to do the following:
- add 3 month check in `validate_psbt_by_ids`
- only store latest 3 months in btc-server db 

im merging this pr into the feature branch and I'll do the rest of the work

scottmillner (approved):
merging this into feature branch

remaining items for feature branch:
- reth side should check that psbt pegouts are within last 3 months before signing
- btc-server db should only store last 3 months of finalized pegouts and truncate anything older

PR #731, opened 2025-05-13
chore(abci): comprehensive logging

shumkov:
In order to identify non-deterministic issues we need more comprehensive logging.

## What was done
- Instrumented ABCI methods
- Enabled trace logs for `reth_authority_consensus` by default, while we are in a heavy testing-development stage ()
- Print an executed eth block and context to the log files
- The verbosity level still adjusts all other crates
- Added more panics to ABCI 😃

New defaults can be overridden:
```
-- log.stdout.filter "reth_authority_consensus=trace,block_with_context=off"
-- log.file.filter "reth_authority_consensus=trace,block_with_context=trace"
```

### DIFF
--- a/crates/consensus/authority/src/comet_bft/abci.rs
+++ b/crates/consensus/authority/src/comet_bft/abci.rs
@@ -1038,7 +1038,7 @@ where
             }
         };
 
-        // TODO: Implement tracing::Valueπ
+        // TODO: Implement tracing::Value
         if tracing::enabled!(tracing::Level::TRACE) {
             let debug_response = format!(
                 "ResponseLoadSnapshotChunk {{ chunk: [{}, {}, ...({} more)] }}",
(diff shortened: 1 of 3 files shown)


rwlockbg (dismissed):
I am ok with this changes, but once the problem is fixed this should all be removed as we will never use such tracing with spans - reth does not use it, and in general what do we care about build_and_execute tracing provided this is not a critical section in a low latency server e.g.

rwlockbg:
@shumkov Pls fix the linter

rwlockbg (dismissed):

scottmillner:
looks like unit tests are failing
the reth binary doesn't compile

rwlockbg (approved):

scottmillner (approved):
🔥

scottmillner:
smoke tested locally and all good
Screenshot 2025-05-13 at 3 33 13 PM

PR #735, opened 2025-05-14
refactor: psbt validation

shumkov:


scottmillner:
circling back
what's the current status for this pr? should we start reviewing it?

shumkov:
@scottmillner This PR needs more work since we don't have a finalized and pending pegouts aren't available on both sides. @rwlockbg is proposing to merge these two processes so we have everything at one place. @lamafab also had some thoughts. We need to discuss and investigate possible solutions.

I see three options on the table:
1. Introduce RPC methods so we can share pending and finalized pegouts 
2. huge refactoring to merge btc server into reth
3. don't do anything at this point 

@rwlockbg shall we start with Pros/Cons of your solution? We probably should create an issue for this topic.

PR #736, opened 2025-05-15
fix(frost-task): error if frost task can't send pegouts to btc-server

scottmillner:
https://github.com/botanix-labs/botanix/issues/1026

### DIFF
--- a/crates/consensus/authority/src/frost_task.rs
+++ b/crates/consensus/authority/src/frost_task.rs
@@ -373,7 +373,9 @@ where
                                         info!(target: "consensus::authority::frost_task::start_task", "Sent checkpoint to btc server");
                                     }
                                     Err(err) => {
-                                        error!(target: "consensus::authority::frost_task::start_task", "Error sending checkpoint to btc server: {}", err);
+                                        // We need to panic bc we have no way to recover any missed pegouts and they won't be sent to the btc server.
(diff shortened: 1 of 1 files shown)


rwlockbg (dismissed):
makes sense, yes!

rwlockbg (approved):

crates/consensus/authority/src/frost_task.rs
```diff
@@ -373,7 +373,9 @@ where
                                         info!(target: "consensus::authority::frost_task::start_task", "Sent checkpoint to btc server");
                                     }
                                     Err(err) => {
-                                        error!(target: "consensus::authority::frost_task::start_task", "Error sending checkpoint to btc server: {}", err);
+                                        // We need to panic bc we have no way to recover any missed pegouts and they won't be sent to the btc server.
+                                        // We could rollback and sync to reprocess the pegouts but we would intentionally be replaying old pegouts.
+                                        panic!("Error sending checkpoint to btc server: {}", err);
```
arminsabouri, line 378: Two questions: 
* Will this panic cause the entire program to shutdown? Some tasks explicity will not do that.
* Receiving this canon state notification implies this block is already saved, will this notification for this exact block ever get retried? Im not sure it will

arminsabouri (commented):
Had two questions about this PR

PR #739, opened 2025-05-15
chore(abci): add better logs

scottmillner:


lamafab (approved):
THANKS!

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -1530,14 +1531,18 @@ where
                         if tracing::enabled!(tracing::Level::DEBUG) {
                             let eth_block_hash = block.hash_slow();
 
-                            debug!(
+                            info!(
```
lamafab, line 1534: Honestly just leave this as `debug!`, the condition above also expclitliy checks for `Level::DEBUG`.

lamafab (dismissed):

SatwikPrabhu:
Added a few more prints around pegin validation just to be sure about that.

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -1404,6 +1404,7 @@ where
     #[instrument(level = "trace", ret, skip(self, request), fields(cfbt_block.height = request.height, cfbt_block.hash = hex::encode(&request.hash)))]
     fn process_proposal(&self, request: RequestProcessProposal) -> ResponseProcessProposal {
         trace!(request = ?RequestProcessProposalTruncatedDebug(&request), "process_proposal request");
+        info!("process_proposal request.txs.len()={:?}", request.txs.len());
```
shumkov, line 1407: We are already in `process_proposal` span + we have block execution info logs.

I think eventually it will make a lot of noise when we will be in stable production state. I believe we just need to increase verbosity level in settings by default and lower them when we feel more comfortable. In this case, we won't need to go to the code and change the verbosity level for each event.

If we want to have an info event for ABCI handlers, it's also fine, but let's have it the same for init_chain, prepare, process, and finalize, then.

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -1749,6 +1755,8 @@ where
 
         info!("Finalized block for height {}", request.height);
 
+        info!("app_hash: {:?}", hex::encode(block_hash.0));
```
shumkov, line 1758: Shall we just add the app hash to the info event above?

rwlockbg (approved):

crates/ethereum/evm/src/execute.rs
```diff
@@ -455,6 +455,8 @@ where
         let consensus_pkg = botanix_consensus_pkg;
         let btc_network = consensus_pkg.btc_network;
 
+        info!(tx_hash = %tx_hash, initial_btc_checkpoint_hash = %consensus_pkg.bitcoin_checkpoint.0.block_hash(), initial_btc_checkpoint_height = %consensus_pkg.bitcoin_checkpoint.1, "PeginCheck: Initial Bitcoin Checkpoint for tx");
```
shumkov, line 458: Let's instrument the function, so you don't need to repeat the same fields and use the debug level. You don't need to know this when you just check that the node is progressing well (this is how you want to use the info level) and on which block it's currently.

crates/primitives/src/botanix/peg_contract.rs
```diff
@@ -55,6 +56,10 @@ impl PeginData {
         bitcoin_commitment: &(bitcoin::block::Header, u32),
         aggregate_pk: &secp256k1::PublicKey,
     ) -> Result<U256, PeginDataError> {
+        info!(self_btc_height = self.bitcoin_block_height, self_amount = %self.amount, num_meta_proofs = self.meta.len(),
```
shumkov, line 59: Instrument function as well.

crates/primitives/src/botanix/peg_contract.rs
```diff
@@ -70,6 +75,13 @@ impl PeginData {
                 PeginMeta::V1(meta) => &meta.inner,
             };
 
+            let pegin_version_for_log = pegin.version;
+            let pegin_block_headers_len_for_log = pegin.block_headers.len();
+
+            info!(self_btc_height = self.bitcoin_block_height, pegin_version = pegin_version_for_log,
```
shumkov, line 81: trace

crates/primitives/src/botanix/peg_contract.rs
```diff
@@ -135,6 +147,11 @@ impl PeginData {
                 return Err(PeginDataError::InvalidBitcoinBlockHeight);
             }
 
+            info!(self_btc_height = self.bitcoin_block_height, calculated_diff = diff,
```
shumkov, line 150: trace

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -1530,14 +1531,18 @@ where
...
-                            debug!(
+                            info!(
                                 %cbft_block_hash,
                                 eth_block_hash = hex::encode(eth_block_hash),
                                 "update block cache for key {}",
                                 cbft_block_hash,
                             );
                         }
 
+                        info!(
```
shumkov, line 1542: we have already the app hash on the info level in build and execute.

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -1559,6 +1564,7 @@ where
     #[instrument(level = "trace", skip(self, request), fields(cfbt_block.height = request.height, cfbt_block.hash = hex::encode(&request.hash)))]
     fn finalize_block(&self, request: RequestFinalizeBlock) -> ResponseFinalizeBlock {
         trace!("request={:?}", request);
+        info!("txs len={:?}", request.txs.len());
```
shumkov, line 1567: Doesn't seem like a production log output https://github.com/botanix-labs/Macbeth/pull/739/files#r2092497212

shumkov (changes requested):

PR #747, opened 2025-05-17
improve(NDD): deprecate version 0 and make deserialization forward compatible

lamafab:
Preface: historic block sync for testnet is broken, so we have no reason to retain version 0. In this PR we remove version 0 and hence demand that the `block_fee_recipient_address` is always present, which is the case for new blocks - those changes here are backwards compatible.

And most importantly; with this PR the NDD deserialization process **becomes forwards-compatible**. As noted in the disclaimer:

```rust
// IMPORTANT: This is a versioning mechanism designed for
// forward compatibility. We want to support new,
// backwards-compatible versions with enhanced functionality
// without requiring coordinated upgrades or hard-forks.
//
// When an older node encounters a newer, unknown version:
// 1. It extracts the data it can understand based on the known structure
// 2. It ignores any trailing data that may be present in newer versions
// 3. It continues to function normally despite version differences
//
// By design, encountering an unknown version alone MUST NEVER
// cause an error. The caller is responsible for validating
// whether the returned NDD is appropriate for their specific
// business logic requirements.
```

Have a look at the unit tests (the diff makes it look confusing, look at them directly) and let me know what you think, **I'd appreciate some scrutiny** here (cc @0xBEEFCAF3) since we don't want to accidentally shoot ourselves in the foot.

# Changelog

### Breaking Changes
- Removed support for legacy VERSION_0 format in NonDeterministicData
- Changed `block_fee_recipient_address` from `Option<Address>` to required `Address`

### Additions
- Added forward compatibility support for future NDD versions
- Added versioning mechanism that allows older nodes to process newer NDD versions
- Added new test cases for unknown version handling and deprecated implementation compatibility

### DIFF
--- a/crates/consensus/authority/src/comet_bft/non_deterministic_data.rs
+++ b/crates/consensus/authority/src/comet_bft/non_deterministic_data.rs
@@ -151,14 +151,19 @@ mod tests {
                 .as_slice(),
         )
         .unwrap();
-        let ndd = NonDeterministicData::new(bitcoin_block_hash, pk);
-        assert_eq!(ndd.version, VERSION_0);
+        let block_fee_recipient_address =
+            Address::parse_checksummed("0x43C8bDCb9AFeBB1D834A7de18CC214a6FD1632d9", None)
+                .expect("valid address");
+
+        let ndd = NonDeterministicData::new(bitcoin_block_hash, pk, block_fee_recipient_address);
+        assert_eq!(ndd.version, VERSION_1);
         assert_eq!(ndd.bitcoin_block_hash, bitcoin_block_hash);
         assert_eq!(ndd.aggregated_public_key, pk);
+        assert_eq!(ndd.block_fee_recipient_address, block_fee_recipient_address);
     }
 
     #[test]
-    fn test_non_deterministic_data_new_v1() {
+    fn test_non_deterministic_data_serde() {
         let bitcoin_block_hash = BlockHash::all_zeros();
         let pk = secp256k1::PublicKey::from_slice(
             hex::decode("039bef292b80427d355cecb89eda8a50a7d2196a93d73dade5a0c4a07cd334815d")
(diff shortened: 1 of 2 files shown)


crates/consensus/authority/src/comet_bft/non_deterministic_data.rs
```diff
@@ -13,64 +13,37 @@ pub(crate) enum NonDeterministicDataDeserializeError {
...
-// Supported on testnet and mainnet
-const VERSION_1: u16 = 1;
+// The default NDD version.
+pub(crate) const VERSION_1: u16 = 1;
 
 /// Type that encapsulates non-deterministic data needed for consensus.
-/// When `block_fee_recipient_address` is `None`, the instance corresponds to VERSION_0.
-/// Otherwise VERSION_1.
 #[derive(Debug, Clone, PartialEq)]
 pub(crate) struct NonDeterministicData {
```
shumkov, line 23: To handle versions and serialization/deserialization, I would recommend using `enum` at the top level + serde + bincode (this is what reth is using for serialization?)

```
pub enum NonDeterministicData {
  V0(NonDeterministicDataV0),
  V1(NonDeterministicDataV0),
}
```

that's how we could introduce breaking changes in future

lamafab: The reason I avoided this pattern is precisely that encountering an unknown version must still return the known structure. I guess we could do something like this:

```rust
enum NonDeterministicData {
    V1(NddBody),
    Unknown(NddBody),
}
```

And a future upgrade would look like this:

```rust
enum NonDeterministicData {
    V1(NddBody),
    V2(NddBodyV2),
    Unknown(NddBodyV2),
}
```

EDIT: I misread your comment - this is exactly what you recommended :)

scottmillner:
utAck

crates/consensus/authority/src/comet_bft/non_deterministic_data.rs
lamafab: > To handle versions and serialization/deserialization, I would recommend using enum at the top level + serde + bincode (this is what reth is using for serialization?)

So I've decided to keep the existing version since your proposal is a breaking change. That being said, I do think that if/when there's a new version available, we should definitely follow that `enum Ndd { V1(...), V2(...), Unknown(...) }` pattern for **after** deserialization.

scottmillner (approved):
i know it's still in draft but LGM!

SatwikPrabhu (approved):

rwlockbg (approved):

PR #748, opened 2025-05-17
improve(ABCI): track hash of block to be finalized before committing

lamafab:
Currently, we simply call [`peek_newest`](https://docs.rs/schnellru/0.2.4/schnellru/struct.LruMap.html#method.peek_newest) on the block cache before committing that returned block, which sounds kind of shaky.

> Returns the most recently used **(inserted or accessed)** element of the map. Does not change the order of the elements.

While this is _probably_ fine - considering that the Comet client will call `commit` immediately after a successful `finalize_block` call - this change explicitly tracks the finalized block hash, which is then used to retrieve the sealed block from cache. This would handle an edge-case where (for some reason) `process_proposal` is called between the expected `finalize_block` and `commit` calls, which would result in us committing a non-finalized block candidate. This is unlikely to happen, but then again, this is external behavior.

Additionally, this reverts the cache-insert skip [as introduced yesterday](https://github.com/botanix-labs/Macbeth/releases/tag/v.1.1.1) for our debugging session.

### DIFF
--- a/crates/consensus/authority/src/comet_bft/abci.rs
+++ b/crates/consensus/authority/src/comet_bft/abci.rs
@@ -331,12 +331,21 @@ enum PayloadBuilderError {
     ParentBlockDoesNotExist,
 }
 
+struct BlockCache {
+    /// Cached blocks, inserted after execution in either `process_proposal` or
+    /// `finalize_block`.
+    cache: LruMap<BlockHash, BlockWithContext>,
+    /// The finalized block hash tracked by `finalize_block`, and then consumed
+    /// by `commit`.
+    tracked_final: Option<BlockHash>,
+}
+
 #[derive(Clone)]
 pub(crate) struct ABCIClient<EF, BF, DB, Pool> {
     storage: Storage<EF, BF, DB>,
     pool: Pool,
     bitcoin_checkpoint: BitcoinCheckpoint,
-    block_cache: Arc<RwLock<LruMap<BlockHash, BlockWithContext>>>,
+    block_cache: Arc<RwLock<BlockCache>>,
     driver_tx: tokio::sync::mpsc::Sender<ABCIDriverMessage>,
     #[allow(dead_code)]
     cbft_rpc_provider: HttpCometBFTRpcClientFactory,
@@ -384,12 +393,18 @@ where
         snapshot_format: u32,
         block_fee_recipient_address: Option<reth_primitives::Address>,
     ) -> Self {
+        // Saving the last 5 blocks that were proposed
+        let block_cache = Arc::new(RwLock::new(BlockCache {
+            cache: LruMap::new(ByLength::new(5)),
+            tracked_final: None,
+        }));
+
         Self {
             storage: storage.clone(),
             pool,
             bitcoin_checkpoint,
             // Saving the last 5 blocks that were proposed
-            block_cache: Arc::new(RwLock::new(LruMap::new(ByLength::new(5)))),
+            block_cache,
             driver_tx,
             cbft_rpc_provider,
             authority_consensus,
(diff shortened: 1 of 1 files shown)


shumkov (dismissed):

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -1765,27 +1784,23 @@ where
...
-            }
-            None => {
-                panic!("Error getting block from cache");
-            }
+
+        // Retrieve the finalized block.
+        let cbft_block_hash =
+            block_cache.tracked_final.take().expect("No tracked final block hash");
+
+        let Some(sealed_block_with_context) = block_cache.cache.remove(&cbft_block_hash) else {
```
shumkov: Shall we cleanup block hash after commit?

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -330,12 +330,21 @@ enum PayloadBuilderError {
     ParentBlockDoesNotExist,
 }
 
+struct BlockCache {
+    /// Cached blocks, inserted after execution in either `process_proposal` or
+    /// `finalize_block`.
+    cache: LruMap<BlockHash, BlockWithContext>,
```
shumkov, line 337: Do we still need LRU? HashMap should be fine now.

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -1538,7 +1552,8 @@ where
                             );
                         }
 
-                        // cache.insert(cbft_block_hash, block_with_context);
+                        // NOTE: We DO NOT set `tracked_final` here.
```
scottmillner: we have to set it here:
- during live consensus, we are never setting `tracked_final` bc we set that in `finalize_block` if we don't find the block in our cache(during historical sync)
- so we will panic everytime at `block_cache.tracked_final.take().expect("No tracked final block hash");` during live consensus

im guessing that's why the int tests are failing

ran this locally and confirm the panic will happen during live consensus

lamafab: Yep, missed that! Thanks will fix asap.

crates/consensus/authority/src/comet_bft/abci.rs
lamafab: Yes, `commit` must not be called unless there's a block to finalize.

crates/consensus/authority/src/comet_bft/abci.rs
lamafab: `LruMap` allows us to set a cache limit in a convenient way, additionally it prevents `process_proposal` from filling up the cache (unlikely to happen in practice, though).

crates/consensus/authority/src/comet_bft/abci.rs
lamafab: All green now!

scottmillner (dismissed):

rwlockbg (dismissed):

crates/consensus/authority/src/comet_bft/abci.rs
shumkov: Sorry, I mean to clean up other block caches when we commit a winner for this height. We won't need them anymore.

shumkov:
@lamafab switch to `dev` branch?

crates/consensus/authority/src/comet_bft/abci.rs
scottmillner: yeah, can't think of a reason we'd need to keep the other blocks

lamafab: The LRU hashmap handles this all for us; it only retains the most recent blocks (we specified five) and discards everything else automatically.

scottmillner (approved):

SatwikPrabhu (approved):

PR #756, opened 2025-05-28
feat(ABCI/storage): Insert pegins & pegouts into database on finalizing commit

lamafab:
Fixes: https://github.com/botanix-labs/botanix/issues/1026

This PR implements simple - but somewhat steep - changes such that extracted pegins and pegouts during block finalization are never lost, without coupling block production and pegout handling.

### Significant Changes

* `ABCIDriver`: we **now always store** (_"stage"_) the pegins & pegouts into the database, which must complete successfully, or `fn commit` - called by the ABCI client after calling `fn finalize_block` - never returns.
* `FrostTask`:
  	* As a subscriber to canonical commits, the Frost task submits the pegins & pegouts to the `btc-server`, initiating a new checkpoint. On successful completion, **it now removes** the staged pegins & pegouts from the database, cleaning up after itself. Additionally, this logic has been moved to an individual method.
	* On startup, **it now checks** the local database for any missed pegins & pegouts, which can happen during an unexpected `btc-server` interruption (as explained in the above issue). If one or more are found, it first submits those to the `btc-server` and initiates a new checkpoint before proceeding with the normal behavior.

### Improvements

* How should this be tested?
	* I discussed this with @scottmillner; writing a full E2E test for this is tricky since right now we don't have the necessary mechanism to shutdown or restart individual services.
	* ~There should at least be a unit test for the storage interfaces, making sure the serialzation & deserialization process works accordingly. Not having the `bitcoin` crate available in the storage crate makes this a little tedious.~
		* UPDATE: This has been added: https://github.com/botanix-labs/Macbeth/commit/e8d66788a6c3a638000f2f1bba45c7c56de166fe (impl), https://github.com/botanix-labs/Macbeth/commit/3c35626c52c9c732f607b4f628f048186eac7f1a (unit tests)
* What happens if the `new_consensus_checkpoint` interface for `btc-server` gets called multiple times with the same data? Is it idempotent?

## Summary by CodeRabbit

- **New Features**
  - Added support for persisting and retrieving staged blockchain headers with peg-in and peg-out data to enhance cross-chain reliability.
  - Introduced new database table, data models, and provider traits for managing staged headers.
  - Improved checkpoint submission logic to handle staged headers on startup and after connection interruptions, ensuring data durability.

- **Bug Fixes**
  - Enhanced peg-in and peg-out data handling to prevent loss during block commits and network disruptions.

- **Tests**
  - Added tests verifying accurate conversion between domain and database models for peg-in and peg-out data.


crates/consensus/authority/src/frost_task.rs
```diff
@@ -290,19 +302,24 @@ where
...
             Err(err) => {
                 // We need to panic bc we have no way to recover any missed
-                // pegouts and they won't be sent to the btc server.
-                // We could rollback and sync to reprocess the pegouts but
-                // we would intentionally be replaying old pegouts.
+                // pegouts and they won't be sent to the btc server. We could
+                // rollback and sync to reprocess the pegouts but we would
+                // intentionally be replaying old pegouts.
+                //
+                // TODO (lamafab): Is panicking still the right approach here?
```
scottmillner: we could just return since we have logic now to update btc-server with pegins/pegouts when it comes back online

scottmillner:
🔥 🔥 🔥 
nice job! i think this covers everything we need:
- decoupled block production from btc-server communication
- recover pegins/pegouts when btc-server is offline
- user pegouts are never missed and are honored on the L1

once we refactor the test-suite to handle shutting down specific processes, we can add an int test to cover this

scottmillner:
```
What happens if the new_consensus_checkpoint interface for btc-server gets called multiple times with the same data? Is it idempotent?
```

My first take on this:
- pegins should be idempotent which they already are I believe
- pegouts should not be:
  - before btc-server stores pending pegouts, we can check if they are in the finalized outputs list 
  - we do have logic where signers check pegouts aren't being replayed but we shouldn't store already finalized pegouts

lamafab:
> once we refactor the test-suite to handle shutting down specific processes, we can add an int test to cover this

Regarding testing; I just migrated the type conversion logic into dedicated functions in `authority/utils`, with basic unit tests for verification. See:

* https://github.com/botanix-labs/Macbeth/pull/756/commits/fdd77328809e08de10ae26a54940e0c9b35f0556 (impl)
* https://github.com/botanix-labs/Macbeth/pull/756/commits/a762b96914265aa911e391998557bc12a41bdfda (unit-tests)

But this does not test the desired behavior of the Frost task, of course. Also, I'll double-check the `new_consensus_checkpoint` (`btc-server`) implementation, just to be sure.

crates/consensus/authority/src/utils.rs
```diff
@@ -154,6 +155,67 @@ fn utxo_from_pegin_meta(pegin_meta: &PeginMeta) -> Utxo {
     }
 }
 
+pub(crate) fn get_staged_pegins_from_pegin_meta(pegins: &[PeginMeta]) -> Vec<models::PeginData> {
+    pegins
+        .iter()
+        .map(|pegin| {
+            let tx_out = pegin.tx().output.get(pegin.outpoint().vout as usize).expect("valid vout");
```
shumkov: Not sure it's a good idea to panic at this level. You could panic in so high level functions where it's sutable but utility and low level logic should return errors. We don't know in which circumstances it will be used. An exepction could be when we check this element a line above and then we get it below.

crates/consensus/authority/src/comet_bft/abci.rs
```diff
@@ -2233,6 +2235,34 @@ where
                             Arc::new(trie_updates.clone()),
                         );
 
+                        let pegins = sealed_block_with_peg
+                            .pegins()
+                            .iter()
+                            .flat_map(|p| p.meta.clone())
```
shumkov, line 2241: We don't need owned value down below

crates/consensus/authority/src/utils.rs
```diff
@@ -154,6 +155,67 @@ fn utxo_from_pegin_meta(pegin_meta: &PeginMeta) -> Utxo {
...
+pub(crate) fn get_utxos_from_staged_pegins(pegins: &[models::PeginData]) -> Vec<Utxo> {
+    pegins
+        .iter()
+        .map(|pegin| Utxo {
+            outpoint: Some(client::OutPoint { txid: pegin.txid.to_vec(), vout: pegin.vout as u32 }),
+            output: Some(TxOut {
+                value: pegin.value,
+                script_pubkey: Some(ScriptBuf { script: pegin.script_pubkey.to_vec() }),
+            }),
+            eth_address: String::from_utf8(pegin.eth_address.to_vec()).expect("to be valid utf-8"),
```
shumkov: same

crates/consensus/authority/src/utils.rs
lamafab: I don't think it's possible to construct a Pegin where this is invalid. But I just saw that this method is available for `PeginMeta` - should probably use that instead:

```rust
/// Get the transaction output from the pegin metadata
pub fn txout(&self) -> &TxOut {
    match self {
        Self::V0(meta) => meta.txout(),
        Self::V1(meta) => meta.inner.txout(),
    }
}

// ...

/// Get the txout for the pegin
pub fn txout(&self) -> &TxOut {
    self.tx
        .output
        .get(self.outpoint.vout as usize)
        .expect("we check on creation that vout exists")
}
```

But yes, generally you're right about this statement:

>  You could panic in so high level functions where it's sutable but utility and low level logic should return errors. We don't know in which circumstances it will be used.

crates/consensus/authority/src/comet_bft/abci.rs
lamafab: There are some unnecessary clones - I just mimicked the interfaces of the existing utility functions. But I should probably improve this.

scottmillner:
tested locally and everything behaved as expected:
- i did a pegout but crashed the btc-server before it could store it
- brought back online and it was processed and stored as expected
🔥

scottmillner (dismissed):
thanks for doing this!

shumkov:
@coderabbitai review

crates/consensus/authority/src/utils.rs
```diff
@@ -112,48 +113,115 @@ pub(crate) async fn get_psbt<BtcServerClient: BtcServerExtendedApi + Clone>(
...
+                outpoint: Some(client::OutPoint { txid: pegin.txid, vout: pegin.vout as u32 }),
+                output: Some(TxOut {
+                    value: pegin.value,
+                    script_pubkey: Some(ScriptBuf { script: pegin.script_pubkey }),
+                }),
+                eth_address: String::from_utf8(pegin.eth_address)?,
+            })
+        })
+        .collect()
+}
```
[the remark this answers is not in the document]
lamafab: Yeah definitely disagree with this approach.

crates/consensus/authority/src/frost_task.rs
```diff
@@ -227,6 +240,183 @@ where
...
+                // Remove staged entries for this block hash; it's now the
+                // responsibility of the btc-server to keep track of the pegins
+                // and pegouts.
+                let existed = self
+                    .storage
+                    .client
+                    .remove_staged_header(header_hash)
+                    .expect("to remove staged header");
+
+                debug_assert!(existed, "Staged header should exist for the given header hash");
```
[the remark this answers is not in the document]
lamafab: The database is local and embedded, and never expected to fail - unlike when you're dealing with web services. In my opinion, panicking is the correct behavior since there's no path to recovery besides dealing with environmental factors, such as filesystem permissions, disk space, etc.

lamafab:
@coderabbitai review

scottmillner (approved):

SatwikPrabhu (approved):
Looks good to me. Great work @lamafab 🔥

PR #759, opened 2025-05-30
Fix/track tx/confirm tracked tx is deeply confirmed

scottmillner:
fixes: https://github.com/botanix-labs/botanix/issues/1050

**Context**
Currently we add a created timestamp when storing tracked txs (which represent a pegout in a psbt).  However, we don't know when the tracked tx will actually be included in a block.  The pegout scheduler sometimes untracks tracked txs in `track_mempool()` even though they are not deeply confirmed bc it only uses the created timestamp.

**What Changed?**
The pegout scheduler confirms tracked txs are deeply confirmed during `track_mempool` before untracking them.

**Testing**
Unit tests refactored and additional unit test added that covers this path.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error handling for tracked transactions in the Pegout Scheduler, providing clearer messages when transactions are missing or not deeply confirmed.

- **New Features**
  - Added new error messages for specific transaction states in synchronization processes.
  - Enhanced test coverage for scenarios involving tracked transactions in the mempool or not deeply confirmed.

- **Style**
  - Reformatted and improved code readability for contract ABI bindings and helper modules; no functional changes.

### DIFF
--- a/bin/btc-server/src/pegout_scheduler/mod.rs
+++ b/bin/btc-server/src/pegout_scheduler/mod.rs
@@ -637,9 +637,14 @@ impl PegoutScheduler {
                     continue;
                 }
 
-                // TODO(scott): finalize this tracked tx
                 // This should not happen if sync_until functions correctly
-                return Err(SyncError::DeeplyConfirmedTxNotFinalized(txid));
+                error!(
+                    "PegoutScheduler::track_mempool: {:?}.",
+                    SyncError::DeeplyConfirmedTxNotFinalized(txid)
+                );
+                // Intentionally not untracking the tx here since this should not happen.
+                // If it does, the underlying issue should be fixed and the tracked tx handled.
+                continue;
             }
 
             // a tx that is in a deeply confirmed block should have been handled already
@@ -862,7 +867,7 @@ pub enum SyncError {
     Db(#[from] database::Error),
     #[error("tracked tx not included in a block: {0}")]
     TrackedTxNotInBlock(Txid),
-    #[error("deeply confirumed tx not finalized: {0}")]
+    #[error("deeply confirmed tx not finalized: {0}")]
     DeeplyConfirmedTxNotFinalized(Txid),
 }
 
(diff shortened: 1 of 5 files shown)


bin/btc-server/src/pegout_scheduler/mod.rs
```diff
@@ -619,8 +620,32 @@ impl PegoutScheduler {
...
+                    continue;
+                }
+
+                // This should not happen if sync_until functions correctly
+                warn!(
+                    "Tx {} is on-chain but not handled by sync_until: {:?}",
+                    &onchain_tx.txid, onchain_tx
+                );
+                // TODO: should we finalize the block here?
+                // For now we continue since hitting this indicates a bug in the sync logic.
```
lamafab: Shouldn't this be signaled (loudly) somehow? Currently, if the `txid` exists on-chain, this will always result in `continue;`, meaning the transaction always gets dropped, regardless of confirmation depth.

scottmillner: yes, we should be loud, thanks for pointing that out
~ive decided to return an error but that means we'll need to finalize that tracked tx bc:~
- ~the error will cause the call to `new_consensus_checkpoint` to fail until that tracked tx is removed~
- ~putting pr in draft till i add that logic~

open to other ideas/suggestions as well

rwlockbg (dismissed):

bin/btc-server/src/pegout_scheduler/mod.rs
scottmillner: so decided to log the error and continue

with the fixes in this pr, we should not hit this error and if we do, there's an underlying bug that needs to be addressed

bin/test-suite/src/mint_attack_contract_abi.rs
lamafab, line 1: Should this be here?

bin/test-suite/src/multi_mint_helper_abi.rs
lamafab, line 1: Should this be here?

lamafab (dismissed):
Great!

bin/test-suite/src/mint_attack_contract_abi.rs
scottmillner: no ser! removed, thanks!

bin/test-suite/src/multi_mint_helper_abi.rs
scottmillner: no ser! removed, thanks!

shumkov (approved):

rwlockbg (approved):

lamafab (approved):

PR #763, opened 2025-06-04
fixed round2 signing package validation

rwlockbg:
## Summary by CodeRabbit

- **Bug Fixes**
  - Adjusted signing package validation to use the correct signing round for improved accuracy.

- **Style**
  - Reformatted import statements and attribute macros for enhanced code readability and consistency.

No changes to user-facing functionality or configuration options.

### DIFF
--- a/bin/btc-server/src/signer/mod.rs
+++ b/bin/btc-server/src/signer/mod.rs
@@ -1,4 +1,7 @@
-use crate::wallet::psbt::{PsbtExt, PsbtInputExt};
+use crate::{
+    util::ROUND2,
+    wallet::psbt::{PsbtExt, PsbtInputExt},
+};
 use bitcoin::psbt::Psbt;
 use error::{SigningRound1Error, SigningRound2Error};
 use frost_secp256k1_tr::{
@@ -89,7 +92,7 @@ pub async fn get_round2_signing_package(
     // Each nonce pair is commitment to a input of the tx
     signing_nonces: &[(SigningNonces, SigningCommitments)],
 ) -> Result<(), SigningRound2Error> {
-    validate_psbt(psbt, ROUND1, min_signers, db)?;
+    validate_psbt(psbt, ROUND2, min_signers, db)?;
 
     let tx = psbt.clone().extract_tx()?;
     let num_inputs = tx.input.len();
--- a/bin/btc-server/src/util.rs
+++ b/bin/btc-server/src/util.rs
@@ -333,7 +333,7 @@ pub fn validate_psbt(
     }
 
     // Check if we have enough round 2 partial sigs
-    // TODO is this necessary? Will signing fail?
+    // TODO: Is this check necessary?
     let sigs = psbt.inputs.iter().map(|i| i.all_partial_signatures()).collect::<Vec<_>>();
     if flags & ROUND2 == ROUND2 {
         // if any of the maps have min signers we should fail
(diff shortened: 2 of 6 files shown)


bin/btc-server/src/signer/mod.rs
```diff
@@ -89,7 +92,7 @@ pub async fn get_round2_signing_package(
     // Each nonce pair is commitment to a input of the tx
     signing_nonces: &[(SigningNonces, SigningCommitments)],
 ) -> Result<(), SigningRound2Error> {
-    validate_psbt(psbt, ROUND1, min_signers, db)?;
+    validate_psbt(psbt, ROUND2, min_signers, db)?;
```
scottmillner, line 95: im gonna look at the `ROUND2` checks and see if it's necessary at all

there's a comment there from armin about the `ROUND2` check:
```
// Check if we have enough round 2 partial sigs
    // TODO is this necessary? Will signing fail?
```

scottmillner: keep this check for now:
- it probably isn't necessary but it does prevent redundant signing

scottmillner:
@DariusParvin this is a good pr for you to look at
Note: I don't think we need the `ROUND2_TRANSITION` and `ROUND2` flags since we don't actually use them currently and signing works fine.  But they could help prevent redundant or unnecessary signing so I don't think they're harmful.

scottmillner (approved):

bin/btc-server/src/signer/mod.rs
lamafab: Can signing requests timeout or be aborted?

bin/btc-server/src/util.rs
```diff
@@ -344,7 +344,9 @@ pub fn validate_psbt(
         }
     }
 
-    // validate partial sigs in round 2
+    // Validate partial sigs in round 2
+    // The ROUND2_TRANSITION flag is currently not being used.
+    // TODO: Is this check necessary?
```
lamafab, line 349: Is there a specific reason we use `u8` flags instead of enums, anyway? This seems unnecessarily error prone.

bin/btc-server/src/util.rs
```diff
@@ -333,7 +333,7 @@ pub fn validate_psbt(
     }
 
     // Check if we have enough round 2 partial sigs
-    // TODO is this necessary? Will signing fail?
+    // TODO: Is this check necessary?
```
lamafab, line 336: Btw;

```rust
if sig.len() > min_signers as usize {
    return Err(ValidatePSBTError::InvalidNumberOfPartialSignatures);
}
```

Shouldn't this do the opposite check? 🤔

scottmillner: I read this check as:
- If the psbt already has more than than the `min_signers`, there's no reason to add any more signatures so we stop signing. But it's normal for a signer to get a round 2 signing request and there not be `min_signers` since everyone hasn't signed yet.

This check nor the `ROUND2_TRANSITION` flags are currently being used so maybe we could remove both. Or maybe they're useful to short-circuit signing

scottmillner: But ultimately it's the coordinator who finalizes the psbt when the quorum of signers has been reached so the checks aren't necessary nor do they add additional security.  The checks just seem to prevent unnecessary work for signers.

bin/btc-server/src/util.rs
rwlockbg: It was a design decision that Armin took some time ago when reworking this btc-server part. I am not against it to be honest, but as you are saying having enums these days and other means bit composition seems to be outdated. We can surely re-work this in another PR but since its core logic I suggest we leave it here and just focus on the fix at hand.

bin/btc-server/src/util.rs
rwlockbg: The check is correct, agree with @scottmillner . It is just to catch an edge case.

shumkov (approved):
👍

PR #775, opened 2025-06-12
build: adjust reth log filter in `.env`

shumkov:
There is no way to change reth log level without editing docker compose file

## What was done?
- Added the `POA_LOG_STDOUT_FILTER` env to the `.env` file to configure the `--log.stdout.filter` option.

For example: 
```bash
POA_LOG_STDOUT_FILTER=reth_authority_consensus=trace,reth=trace,block_with_context=off
```

## How Has This Been Tested?
- Running a local network and seeing trace logs from the poa.

## Breaking Changes
None

## Checklist:

## Summary by CodeRabbit

- **Documentation**
  - Updated the example environment file to include a new log filter configuration option for POA stdout logs.

- **Chores**
  - Modified Docker Compose and Makefile configurations to support loading environment variables from both a global `.env` file and node-specific files.
  - Enhanced the POA Docker service to allow log filtering via an environment variable, with a default value provided if not set.

### DIFF
--- a/docker-local/docker-compose.yml
+++ b/docker-local/docker-compose.yml
@@ -65,6 +65,7 @@ services:
       - --ws.addr=0.0.0.0
       - --ws.api=eth,net,trace,txpool,web3,rpc,admin
       - -vvv
+      - --log.stdout.filter=${POA_LOG_STDOUT_FILTER:-block_with_context=off}
       - --btc-server=${COMPOSE_PROJECT_NAME?:err}-btc_server-1:8080
       - --btc-network=${BITCOIND_NETWORK:?err}
       - --btc-signing-server-jwt-secret=/app/config/bjwt.hex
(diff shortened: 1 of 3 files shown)


scottmillner (dismissed):

docker-local/docker-compose.yml
```diff
@@ -65,6 +65,7 @@ services:
       - --ws.addr=0.0.0.0
       - --ws.api=eth,net,trace,txpool,web3,rpc,admin
       - -vvv
+      - --log.stdout.filter=${POA_LOG_STDOUT_FILTER:-block_with_context=off}
```
lamafab, line 68: Shouldn't the default filter be `block_with_context=off` (without the `-` at the start)?

lamafab (approved):

docker-local/docker-compose.yml
shumkov: This is just variable interpolation syntax. `${var:-value}` - `value` will be used if `var` is not set or empty.

scottmillner (approved):

PR #776, opened 2025-06-13
feat (frost signing): enforce signing for pegouts within last 3 months

scottmillner:
Signers should only sign a psbt that contains pegouts within the last 3 months. The btc-server db is only storing the last 3 months of finalized pegouts so it doesn't store an unbounded list. Signers cross-check the finalized pegouts list when signing to prevent replaying pegouts that have already been honored on the L1. Signers must not sign for any pending pegouts outside of the 3 months window or an older pegout could be replayed.

The btc-server does not have access to the L2 db so cannot do this check.  

## What was done?
- Added `validate_psbt_id_by_maximum_cutoff_age` which is called in `validate_psbt_by_ids` which is called during psbt creation and every round of signing.

## How Has This Been Tested?
- Added unit tests

## Breaking Changes
None

## Checklist:

### DIFF
--- a/crates/consensus/authority/src/frost_task.rs
+++ b/crates/consensus/authority/src/frost_task.rs
@@ -384,7 +384,8 @@ where
         };
 
         if let Err(e) =
-            validate_psbt_by_ids(self.storage.client.clone(), self.storage.btc_network, &psbt).await
+            validate_psbt_by_ids(&self.storage.client.clone(), self.storage.btc_network, &psbt)
+                .await
         {
             error!(
                 target: "consensus::authority::frost_task::handle_canon_state_commit",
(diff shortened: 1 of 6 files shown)


lamafab (approved):

rwlockbg (approved):

crates/consensus/authority/src/frost_task.rs
```diff
@@ -384,7 +384,8 @@ where
         };
 
         if let Err(e) =
-            validate_psbt_by_ids(self.storage.client.clone(), self.storage.btc_network, &psbt).await
+            validate_psbt_by_ids(&self.storage.client.clone(), self.storage.btc_network, &psbt)
```
shumkov: Do we need to clone here even if we expect reference?

lamafab: Probably not.

scottmillner: removed, thanks!

PR #786, opened 2025-06-18
refactor(finalized-pegouts): add prune_finalized_pegouts_ids() and test

scottmillner:
Issue: https://github.com/botanix-labs/botanix/issues/1054
The btc-server should only store the most recent 3 months of finalized pegouts so it doesn't store an unbounded list.

## What was done?
- Added a method to truncate the `finalized_pegout_ids` tree using a sliding window of the last 3 months
- Call the method in pegout scheduler `finalize_block()`

## How Has This Been Tested?
- Added unit test covering the 3 types of finalized pegouts:
  - pegout with no timestamp, one with a timestamp within sliding window, and one outside sliding window
- Manual tested
## Breaking Changes
None bc the timestamp field on the finalized pegout is Option

## Checklist:
- [X] I have added or updated relevant unit/integration/functional/e2e tests
- [X] I have tested my changes running a local network, and they work as expected
- [X] I have performed a self-review
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have added "!" to the title and described breaking changes in the corresponding section if my code contains any
- [X] I have made corresponding changes to the documentation if needed

## Summary by CodeRabbit

- **New Features**
	- Introduced automatic pruning of finalized pegout IDs older than 90 days, helping to keep the database clean and efficient.
- **Bug Fixes**
	- Improved handling of pegout IDs missing timestamps by updating them accordingly during pruning.
- **Tests**
	- Added a test to verify correct behavior of the pegout ID pruning process.

### DIFF
--- a/bin/btc-server/src/database/mod.rs
+++ b/bin/btc-server/src/database/mod.rs
@@ -1,4 +1,9 @@
-use std::{collections::BTreeMap, io::Write, path::Path};
+use std::{
+    collections::BTreeMap,
+    io::Write,
+    path::Path,
+    time::{SystemTime, UNIX_EPOCH},
+};
 
 use crate::{
     pegout_id::PegoutId,
@@ -15,6 +20,7 @@ use bitcoin::{
 use client::SigningStatus;
 use frost_secp256k1_tr as frost;
 use futures::Stream;
+use log::info;
 use miniscript::psbt::PsbtExt;
 use serde::{Deserialize, Serialize};
 use sled::transaction::{ConflictableTransactionError, TransactionError};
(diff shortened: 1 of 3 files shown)


shumkov:
@coderabbitai review

bin/btc-server/src/database/mod.rs
```diff
@@ -904,6 +913,57 @@ impl Db {
...
+            .map_err(Error::DurationSinceEpoch)?
+            .as_secs();
+
+        let cutoff_timestamp = now.saturating_sub(PRUNING_WINDOW_SECONDS);
+
+        let finalized_pegouts = self.get_finalized_pegout_ids()?;
+
+        self.finalized_pegout_ids
+            .transaction(|database_tx| {
+                for pegout in finalized_pegouts.iter() {
```
shumkov, line 935: Minor performance optimization: instead of having two loops (first in `get_finalized_pegout_ids` and second here) and allocating a big vector, we could just iterate through `self.finalized_pegout_ids.iter()` right here.

scottmillner: I originally tried that way but I'm inside the `database_tx` and you can't iterate through the sled tree inside the `database_tx`

rwlockbg (approved):

bin/btc-server/src/database/mod.rs
shumkov: Oh interesting. Maybe we should leave some comments here to explain why we do it this way?

scottmillner: comment added

shumkov (approved):

PR #793, opened 2025-06-24
fix(rpc): periodically wake up and poll TransactionManager

scottmillner:
RPC nodes sometimes don't gossip their pending txs. This could be bc the future that handles this functionality is starved.
I was not able to reproduce this locally.

Another cause could be the node doesn't gossip the pending txs successfully but thinks it has.

## What was done?

- Every 3 seconds wake the future so it's queues (including pending txs) are drained and pending txs are gossiped.

## How Has This Been Tested?

- I've tested this locally.

## Breaking Changes

- None

## Checklist

### DIFF
--- a/crates/net/network/src/transactions/mod.rs
+++ b/crates/net/network/src/transactions/mod.rs
@@ -249,6 +249,8 @@ pub struct TransactionsManager<Pool> {
     max_transactions_seen_by_peer_history: u32,
     /// `TransactionsManager` metrics
     metrics: TransactionsManagerMetrics,
+    // Optional periodic waker task to prevent starvation
+    periodic_waker: Option<tokio::task::JoinHandle<()>>,
 }
 
 impl<Pool: TransactionPool> TransactionsManager<Pool> {
@@ -300,6 +302,7 @@ impl<Pool: TransactionPool> TransactionsManager<Pool> {
             max_transactions_seen_by_peer_history: transactions_manager_config
                 .max_transactions_seen_by_peer_history,
             metrics,
+            periodic_waker: None,
         }
     }
 }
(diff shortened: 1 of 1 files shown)


crates/net/network/src/transactions/mod.rs
```diff
@@ -1236,6 +1236,16 @@ where
 
         let this = self.get_mut();
 
+        // Ensure periodic wake-up in case of starvation
```
rwlockbg, line 1242: @scottmillner I dont think this is very efficient . Everytime we call the poll , a new task is spawned which is not what we ultimately want. Lets try and do someting like:
```
pub struct TransactionsManager<Pool> {
    // ...
    
    // Optional periodic waker task to prevent starvation
    periodic_waker: Option<tokio::task::JoinHandle<()>>,
}
```
and init with:
```
impl<Pool> TransactionsManager<Pool> {
    pub fn new(...) -> Self {
        Self {
            //  ...
            periodic_waker: None,
        }
    }
}
```
and instead of the impl after `let this = self.get_mut();`, lets add this:
```
if this.periodic_waker.is_none() {
    let waker = cx.waker().clone();
    this.periodic_waker = Some(tokio::spawn(async move {
        let mut interval = tokio::time::interval(Duration::from_secs(3));
        loop {
            interval.tick().await;
            waker.wake_by_ref();
        }
    }));
}
```
which would basically make sure the periodic waker task is init only once and it keeps running in the background throughout the polling duration.

I think we also need to clean it using Drop :
```
impl<Pool> Drop for TransactionsManager<Pool> {
    fn drop(&mut self) {
        if let Some(handle) = self.periodic_waker.take() {
            handle.abort();
        }
    }
}
```

Pls check if htat makes sense and is feasible. I just wrote the code, it might not compile.

shumkov: Good suggestion.

scottmillner: ah yes, that's a lot better, thanks!

changes made and tested!

shumkov (approved):

rwlockbg (approved):

PR #795, opened 2025-06-24
feat/truncate finalized outputs latest 3 months

scottmillner:

BTC-Server should only store the last 3 months of finalized pegouts for 2 reasons:
- to prevent storing an unbounded list
- to prevent gossiping excessively large payloads during wallet sync

## What was done?

- Truncate the finalized outputs sled tree to only contain the most recent 3 months of data
- Only gossip the last 3 months of finalized outputs during wallet sync
- Only sign for pegouts that are within the last 3 months since we cross-check pending pegouts against the finalized pegouts list

## How Has This Been Tested?

- Unit tests added

## Breaking Changes

- None

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added timestamp tracking to pegout requests and finalized pegouts, enabling precise temporal metadata.
  - Implemented automatic pruning of finalized pegout records older than 90 days to optimize database storage.
  - Introduced validation ensuring pegout-related blocks are within a configurable maximum age cutoff (default 3 months).
  - Enhanced error reporting with new error types for outdated or missing pegout data.

- **Improvements**
  - Expanded test coverage and utilities for timestamp handling and pruning mechanisms.
  - Renamed fields in test suite for clearer process management.
  - Converted select async functions to synchronous to simplify usage.

- **Dependency Updates**
  - Upgraded and reorganized dependencies for better maintainability and performance.


crates/consensus/authority/src/utils.rs
```diff
@@ -334,13 +343,34 @@ pub(crate) async fn epoch_pegouts(
...
+///
+/// # Arguments
+///
+/// * `timestamp` - The timestamp of the block in seconds since the UNIX epoch.
+/// * `max_age_cutoff` - The maximum acceptable age of the block as a `Duration`.
+///
+/// # Returns
+///
+/// Returns `true` if the block's age is less than the specified max cutoff age, otherwise `false`.
+pub fn is_block_age_acceptable(timestamp: u64, max_age_cutoff: Duration) -> bool {
```
shumkov, line 356: Don't we have this method in the btc sever utils as well? I guess this should be resolved with the pbst validation refactoring?

shumkov (dismissed):

bin/btc-server/src/util.rs
```diff
@@ -11,31 +15,39 @@ use bitcoin::{
...
+///
+/// Returns `true` if the block's age is less than the specified max cutoff age, otherwise `false`.
+pub fn is_block_age_acceptable(timestamp: u64, max_age_cutoff: Duration) -> bool {
+    let now = match SystemTime::now().duration_since(UNIX_EPOCH) {
+        Ok(now) => now.as_secs(),
+        Err(_) => return false,
+    };
+
+    let threshold = now.saturating_sub(max_age_cutoff.as_secs());
+    timestamp > threshold
```
DariusParvin: nit: the existing logic is correct, but since the method refers to 'age', i think this way round might be clearer
```suggestion
    let age_secs = now.saturating_sub(timestamp);
    age_secs < max_age_cutoff.as_secs()
```

bin/btc-server/src/database/mod.rs
```diff
@@ -52,6 +58,9 @@ const KEY_FINALIZED_PEGOUT_IDS_MERKLE_ROOT: &[u8; 9] = b"pegoutids";
 /// sled tree for pending pegout requests
 const TREE_PENDING_PEGOUTS: &[u8; 7] = b"pegouts";
 
+/// Sliding window duration in seconds (90 days)
+const PRUNING_WINDOW_SECONDS: u64 = 90 * 24 * 60 * 60;
+
```
DariusParvin: is it possible or worth it to import `MAX_BLOCK_TS_CUTOFF_DURATION_SECS`, or have them defined in the same place?

bin/btc-server/src/database/mod.rs
```diff
@@ -82,11 +91,14 @@ pub struct FinalizedPegout {
     pub id: PegoutId,
     /// The Botanix block number.
     pub block_number: u64,
+    /// The timestamp of the pegout, if available.
+    #[serde(default)]
+    pub timestamp: Option<u64>,
 }
```
DariusParvin: nit: I think it might be helpful to clarify what 'timestamp' refers to (e.g. when the request was initiated, L2 confirmation, L1 confirmation)

bin/btc-server/src/database/mod.rs
```diff
@@ -1431,14 +1510,70 @@ mod tests {
...
+                FinalizedPegout { id: pegout_id, block_number: 100, timestamp: None };
+            finalized_pegout_ids.push(finalized_pegout);
+        }
+
+        // update finalized pegout to be within the pruning window
+        let now = SystemTime::now().duration_since(UNIX_EPOCH).unwrap().as_secs();
+        finalized_pegout_ids[0].timestamp = Some(now);
+
+        // update finalized pegout to be outside the pruning window
+        finalized_pegout_ids[1].timestamp = Some(now.saturating_sub(PRUNING_WINDOW_SECONDS + 1));
```
DariusParvin: nit: 'pruning window' sounds a bit like that is a window during which pegouts get pruned, but it's actually the opposite. Maybe 'retention window' would be clearer?

DariusParvin (dismissed):
Overall it looks good to me! I'm still new to the codebase, so I left some non-blocking comments and questions in case they're helpful. 

A couple of general question:

- It looks like we're fully deleting the pegouts here, but would it make sense to do a soft delete or archive them? It might be useful for debugging or metrics, unless we already have another way to get that information.

- I noticed we're using raw `u64` for the timestamp, is there a reason we dont use a more specific Timestamp type? That way we could also name it something more descriptive like `l2_confirmation_ts`.

scottmillner:
> Overall it looks good to me! I'm still new to the codebase, so I left some non-blocking comments and questions in case they're helpful.
> 
> A couple of general question:
> 
> * It looks like we're fully deleting the pegouts here, but would it make sense to do a soft delete or archive them? It might be useful for debugging or metrics, unless we already have another way to get that information.
> * I noticed we're using raw `u64` for the timestamp, is there a reason we dont use a more specific Timestamp type? That way we could also name it something more descriptive like `l2_confirmation_ts`.

I get wanting to soft delete but the original intent of this pr is not to store an unbounded list in the sled db bc eventually we'd have a memory issue.  In hindsight, sled db isn't the optimal choice for what we need. 

Ultimately we need a different db or approach that resolves this issue which leads me to...
@lamafab is working on pegout proofs which should help: his spec describes adding proof data as an `OP_RETURN` output in the PSBT. We could build a history of finalized pegouts using this proof data.

As far as the timestamp, it was probably to be inline with the Header struct that uses u64 for it's timestamp field.

crates/consensus/authority/src/utils.rs
scottmillner: removed, thanks!

bin/btc-server/src/util.rs
scottmillner: removed this method since it wasn't being used

bin/btc-server/src/database/mod.rs
scottmillner: That const is in a crate that isn't a dependency of btc-server
I added the separate const here so I didn't have to add another dependency

But yeah, I get your point that it feels wrong to have the same const in two places.

bin/btc-server/src/database/mod.rs
scottmillner: good point, updated the comment to be more descriptive

bin/btc-server/src/database/mod.rs
scottmillner: you're right! changed the const name to use RETENTION, thanks!

DariusParvin (approved):

shumkov (approved):

rwlockbg (approved):
Well known code, looks fin! Thanks @scottmillner Just make sure the linter is happy and good to go.

PR #800, opened 2025-06-26
feature(priority-fee): set default to 0.01 gwei

scottmillner:

This PR sets the default `eth_maxPriorityFeePerGas` to 0.01 gwei.
This currently defaults to 1 gwei.
Our initial basefee is 7 wei.  So `eth_gasPrice` would return a suggested gas price of `1_000_000_007` which `1.000000007` gwei.  This is way overpriced so setting the default to 0 would result in a gas price of 7 wei.  The gas price is dynamic but only considers the last 20 blocks (this is a constant that can be changed). Some libraries like ethers may hardcode the priority fee to 1 gwei which we can't control.  But if they hit our node directly they will get the accurate estimate.

For a deeper explanation on how the gas oracle that determines gas price estimates work, you can see the original GH issue from the ethereum repo at https://github.com/ethereum/pm/issues/328#issuecomment-853234014

TL;DR
```
Edit: Geth's gas price oracle retrieves the cheapest 3 transactions from the past X blocks, and uses the 60th percentile as the suggestion for the price. The 60th percentile ensures we're aiming for inclusion in 2-3 blocks, whereas looking at the minimums ensures we're trying to push prices downward instead of up.
```

## What was done?

- Set the initial `eth_maxPriorityFeePerGas` to 0.

## How Has This Been Tested?

- Tested locally

## Breaking Changes

- None

## Checklist

### DIFF
--- a/crates/rpc/rpc-eth-types/src/gas_oracle.rs
+++ b/crates/rpc/rpc-eth-types/src/gas_oracle.rs
@@ -4,7 +4,7 @@
 use std::fmt::{self, Debug, Formatter};
 
 use derive_more::{Deref, DerefMut, From, Into};
-use reth_primitives::{constants::GWEI_TO_WEI, BlockNumberOrTag, B256, U256};
+use reth_primitives::{constants::DEFAULT_SUGGESTED_TIP, BlockNumberOrTag, B256, U256};
 use reth_rpc_server_types::constants;
 use reth_storage_api::BlockReaderIdExt;
 use schnellru::{ByLength, LruMap};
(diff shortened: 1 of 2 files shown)


shumkov (dismissed):

jeffgeudens (dismissed):

crates/rpc/rpc-eth-types/src/gas_oracle.rs
```diff
@@ -290,7 +290,10 @@ pub struct GasPriceOracleResult {
 
 impl Default for GasPriceOracleResult {
     fn default() -> Self {
-        Self { block_hash: B256::ZERO, price: U256::from(GWEI_TO_WEI) }
+        // Set default to 0 instead of 1 Gwei (previously used GWEI_TO_WEI constant).
+        // The initial base fee is 7 wei so a default of 1 Gwei is not practical.
+        // Using a default of 0.01 Gwei.
+        Self { block_hash: B256::ZERO, price: U256::from(10_000_000) }
```
rwlockbg: @scottmillner SHould we not make this a constant ?

rwlockbg (dismissed):

crates/rpc/rpc-eth-types/src/gas_oracle.rs
scottmillner: yes, done!

shumkov (approved):
👍

PR #808, opened 2025-06-27
feat(wallet): (3/4) optimize pegout change target

DariusParvin:
indirectly related to https://github.com/botanix-labs/botanix/issues/891

- set the pegout change target to a percentage of pegout value, with configurable min/max values. Over time this would help reduce utxo bloat in the federation wallet.

## What was done?

- set default pegout change to 50% the value of the pegout. This is a balance between:
    - too large --> locking up liquidity in remaining utxos
    - too small --> increasing utxo bloat (lots of small change outputs)
- maximum target change set to 5% of remaining utxo value
- minimum target change set to 0.0001 btc (equal to minimum pegout defined in sidecar). Enforcing the minimum takes precedence over the maximum. 

## How Has This Been Tested?

- unit testing of every path within the new method.

## Breaking Changes

- None

## Checklist

- [ ] I have tested my changes running a local network, and they work as expected
- [x ] I have added "!" to the title and described breaking changes in the corresponding section if my code contains any

## Summary by CodeRabbit

* **New Features**
  * Improved fee handling and dust output filtering for pegout transactions, ensuring only valid outputs are included.
  * Dynamic calculation of change amount based on transaction value and available funds.
  * Added utility to insert dummy Taproot signatures into PSBTs for testing purposes.

* **Bug Fixes**
  * Enhanced detection and handling of insufficient funds and dust outputs.

* **Tests**
  * Added and extended tests to cover new fee application, dust filtering, and change calculation scenarios.

* **Other**
  * Improved logging for better debugging of filtered outputs.

### DIFF
--- a/bin/btc-server/src/wallet/coin_selection.rs
+++ b/bin/btc-server/src/wallet/coin_selection.rs
@@ -56,6 +56,9 @@ impl PartialEq for CoinSelectionError {
     }
 }
 
+// Change calculation constants
+const TARGET_CHANGE_PERCENT: u64 = 50; // 50% of total pegout value
+const MAX_CHANGE_PERCENT: u64 = 5; // 5% of total UTXOs
 const MIN_CHANGE_SATS: u64 = 10_000; // minimum pegout value (0.0001 BTC)
 
 /// Coin selection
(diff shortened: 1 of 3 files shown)


scottmillner (approved):

bin/btc-server/src/wallet/coin_selection.rs
```diff
@@ -198,6 +207,49 @@ fn calculate_required_fee(
...
+/// **Benefits of smaller change:**
+/// - Keeps more liquidity available while the current pegout is waiting to be confirmed
+///
+/// The target change is calculated as a percentage of the pegout value, with a ceiling
+/// to prevent excessive liquidity lockup and a floor to prevent the change from being too small.
+fn calculate_target_change(total_pegout_value: Amount, remaining_utxos_value: Amount) -> Amount {
+    // default target change is a percentage of the total pegout value
+    let mut target_change = total_pegout_value
+        .checked_mul(TARGET_CHANGE_PERCENT)
+        .expect("Bitcoin amounts should never overflow u64")
```
rwlockbg: Can we handle these gracefully so that calculate_target_change returns Result<Amount, Error> ?

Same applies to all expects ?

lamafab: We could also just use `.unwrap_or()` with `u64::MAX` as the amount.

lamafab: Also, shouldn't this be `total_pegout_value / 100 * TARGET_CHANGE_PERCENT`? Because right now you multiply by `50`, not `0.5`.

shumkov:
@coderabbitat review

rwlockbg (approved):

bin/btc-server/src/wallet/coin_selection.rs
DariusParvin: I wasnt sure what the fallback value would be for division, so instead i replaced the `.expect` calls with `.ok_or(CoinSelectionError::FeeRateOverflow)?` to be consistent.

@lamafab you're right, the division on the line after the comment, as in:

```
    let max_change_value = remaining_utxos_value
        .checked_mul(MAX_CHANGE_PERCENT)
        .ok_or(CoinSelectionError::FeeRateOverflow)?
        .checked_div(100)
        .ok_or(CoinSelectionError::FeeRateOverflow)?;
```

PR #816, opened 2025-07-04
fix(priority fee): enforce minimum

scottmillner:

We want to enforce a minimum priority fee 

This has two main benefits:
- enforcing a minimum will prevent spam
- while txs will still be very cheap, block fees will increase which helps the overall ecosystem of the network
  - node operators will be better rewarded for securing the network
  - more fees will enter the yield contract which is spread across all stakers which rewards participation

tracking issue: https://github.com/botanix-labs/botanix/issues/1074

## What was done?

Added `--txpool.minimum-priority-fee` parameter and set its default to DEFAULT_SUGGESTED_TIP
That default is what the gas oracle uses so it makes sense to make the default at least that value.

## How Has This Been Tested?

Tested locally and ensured our bridge works as expected. The dapp and sidecar both use legacy txs so they aren't impacted by this ticket.  

## Breaking Changes

None. This only rejects txs that don't meet the minimum tip and returns that message so the user can adjust their price.

These change exist upstream in Reth repo:
https://github.com/paradigmxyz/reth/pull/17183

## Checklist

### DIFF
--- a/crates/transaction-pool/src/validate/eth.rs
+++ b/crates/transaction-pool/src/validate/eth.rs
@@ -124,7 +124,7 @@ pub(crate) struct EthTransactionValidatorInner<Client, T> {
     /// The current max gas limit
     block_gas_limit: u64,
     /// Minimum priority fee to enforce for acceptance into the pool.
-    minimum_priority_fee: Option<u128>,
+    minimum_priority_fee: u128,
     /// Stores the setup and parameters needed for validating KZG proofs.
     kzg_settings: EnvKzgSettings,
     /// How to handle [`TransactionOrigin::Local`](TransactionOrigin) transactions.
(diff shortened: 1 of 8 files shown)


crates/transaction-pool/src/validate/eth.rs
```diff
@@ -252,8 +252,12 @@ where
         {
```
shumkov, line 252: We compare two `Option` values here? If one of them is `None`, then the clause returns `false`?

scottmillner: yes
it's bc `max_priority_fee_per_gas` and `minimum_priority_fee` are only related to eip1559 txs (type 2) and not legacy txs (type 0)

shumkov: Thank you for the clarification.

shumkov (approved):
LGTM! 👍

rwlockbg (approved):
Nice tests too!

PR #785, opened 2025-06-18
fix(storage): `ProviderFactory` methods aren't persisting data

shumkov:
## Issue being fixed or feature implemented
https://github.com/botanix-labs/botanix/issues/1064

Implementation of
`SnapshotWriter`, `WalletStateSyncWriter`, and `StagedHeader` for `ProviderFactory` do not commit database transactions. 

## What was done?
- Commit database transactions in `SnapshotWriter`, `WalletStateSyncWriter`, and `StagedHeader` implementations.

## How Has This Been Tested?
- Testing scenario described in https://github.com/botanix-labs/botanix/issues/1064

## Breaking Changes
None

## Checklist:
- [x] I have added or updated relevant unit/integration/functional/e2e tests
- [x] I have tested my changes running a local network, and they work as expected
- [x] I have performed a self-review
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have added "!" to the title and described breaking changes in the corresponding section if my code contains any
- [x] I have made corresponding changes to the documentation if needed

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability of database write operations for snapshots, state sync records, and staged headers.
  - Ensured changes are properly saved after each write operation, reducing the risk of missing or incomplete updates.

- **Other**
  - Minor correction to calculation of total difficulty value.

PR #825, opened 2025-07-08
fix/pegouts/recover-missed-pegouts

scottmillner:
**NOTE**
These changes have already been reviewed and merged into develop. They need to be deployed to production.

These changes handle any missed pegin/pegouts if reth isn't able to communicate with the btc-server.
Currently these pegins/pegouts are not recovered: added to the btc-server db when connection is reestablished.
This is most problematic for pegouts since this means those pegouts will never be handled by the coordinator since it's unaware it has missed processing them.  For pegins, the user has received their L2 funds but the utxo hasn't been stored in the btc-server so it's not part of the utxo set available to honor pegouts.

I'm cherry picking 3 prs that are already merged into develop:
- https://github.com/botanix-labs/Macbeth/pull/759
- https://github.com/botanix-labs/Macbeth/pull/756
- https://github.com/botanix-labs/Macbeth/pull/785

The last pr adds `commit()` to db txs since staged headers (and some snapshot entries) were not being persisted correctly. The pr is a single commit so I cherry-picked the entire commit instead of pulling out just the stage header relevant code in fear I would miss something.

## What was done?

- The concept of staged headers has been added: reth keeps track of headers for which processing (in terms of the btc-server) was not completed. In short, headers whose pegins/pegouts were not stored in the btc-server db.  Once reth can connect to the btc-server again, it will parse the unprocessed headers and send the pegins/pegouts to btc-server. Once the btc-server has acked it stored them, the staged header is considered processed and no longer tracked.

- Also, `fn track_mempool()` was refactored to check that a tracked_tx is indeed deeply confirmed before untracking it. Currently there is an edge case where a tracked_tx is untracked.  This means it's never `finalized` and becomes part of the finalized pegouts list which it should be.

- Both these changes need to go into together:
  - Now that we reprocess missed pegouts, there is the possibility we could retry an already honored pegout if we introduce a bug. So making the finalized pegouts list tracking function correctly defends against signing for a finalized pegout

## How Has This Been Tested?

- Unit tests were added for this logic
- This has been tested manually
- All integration tests pass

## Breaking Changes

- There are no breaking changes and this only needs to be deployed on our coordinator node since the coordinator creates the psbt that honors pegouts.

## Checklist


rwlockbg (approved):
All code already checked and approved previously. Looks LGTM.

lamafab (approved):

PR #841, opened 2025-07-15
hotfix(hard-fork): Initiate `ActivationManager` with upgrade conditions & floor base fee per gas

lamafab:
So, this hotfix is responsible to setting a minimum base fee per gas (_floor value_), which is a breaking change (hard-fork). It utilizes the `ActivationManager` to coordinate the readiness and timing to set the transition into motion without causing a chain halt.

In order to set this floor value, we cannot just modify the `ChainSpec::base_fee_params_at_timestamp(..)` method since it's part of the higher EIP-1559 fee calculation; we have to bypass that formula entirely. Additionally, since the `ActivationManager` is responsible for coordinating the hard-fork, we do not yet know at what `timestamp` the transition takes place - that's only something we can determine and hard-code retroactively.

Important changes:

* The conditions for the hard-fork to take place: https://github.com/botanix-labs/Macbeth/pull/841/commits/231a6668d8a5d445e8dc89cd826ef364425e1eb7

```rust
if *is_testnet {
    quorum = 100; // 100% ~= 3/3 members must approve
    min_validator_count = 3;
    target_height = 1; // Activate as soon as possible
    our_vote = Vote::Aye;
} else {
    quorum = 100; // 100% ~= 15/15 members must approve
    min_validator_count = 15;
    target_height = 1; // Activate as soon as possible
    our_vote = Vote::Aye;
}
```

* Update the NDD to version 2: https://github.com/botanix-labs/Macbeth/pull/841/commits/67c6c042e6a9fc86ae32b124d0dea06efdba7a50
  * Includes a Botanix _runtime version_ indicator that signals when the floor value has to be set.
  * Includes an (optional) network upgrade payload such that the `ActivationManager` can determine readiness.
  * NOTE: There's currently a _hack_ that version 2 gets (de-)serialized as version 1 in a backwards-compatible way; this should be removed post-fork.
* Implement a simple vote tracker: https://github.com/botanix-labs/Macbeth/pull/841/commits/c3d1329560911383fa63c691508a7b08b4851169
  * This simply tracks the last vote by a Fed member. In a permissionless context, this would track votes over time to determine community sentiment, but it's too early for that.
  * This is used by the `ActivationManager` underneath.
* Update the `base_fee_per_gas` field in the Botanix header before block execution: https://github.com/botanix-labs/Macbeth/pull/841/commits/5b9cd1fc247ff34b83c3f34e211f5e09e3092711
  * Used by ABCI's `process_proposal` and `finalize_block`.
* ABCI state transitions:
  * `prepare_proposal`: https://github.com/botanix-labs/Macbeth/pull/841/commits/dfb2bc591c1e77ef293de26f20269f1531ae97cc
    * Note that this sets the floor value to the attributes that are going to be passed to the default Ethereum payload builder. 
  * `process_proposal`: https://github.com/botanix-labs/Macbeth/pull/841/commits/9f4f715baaa4de347c5a2a4de2f5c6cae59dacea
  * `finalize_block`: https://github.com/botanix-labs/Macbeth/pull/841/commits/b93a595bb52737b3b253c41ce30dbae40aa22fc5

### DIFF
--- a/bin/reth/src/commands/poa/mod.rs
+++ b/bin/reth/src/commands/poa/mod.rs
@@ -13,7 +13,11 @@ use eyre::Context;
 use fdlimit::raise_fd_limit;
 use futures::TryFutureExt;
 use reth_authority_consensus::{
-    comet_bft::abci::ABCIDriver,
+    activation_manager::ActivationManagerBuilder,
+    comet_bft::{
+        abci::{ABCIDriver, RUNTIME_VERSION_ACTIVE, RUNTIME_VERSION_UPGRADE},
+        vote_tracker::VoteWatcher,
+    },
     random_source_provider::RandomSourceProvider,
     snapshot_manager::SnapshotRunnable,
     utils::{is_known_minting_contract, retry_exec},
@@ -57,7 +61,7 @@ use reth_chainspec::{BOTANIX_MAINNET_CHAIN_ID, BOTANIX_TESTNET_CHAIN_ID};
 use reth_cli_runner::CliContext;
 use reth_config::{config::StageConfig, Config};
 use reth_consensus_common::utils;
-use reth_db::{database::Database, init_db, DatabaseEnv};
+use reth_db::{database::Database, init_db, models::Vote, DatabaseEnv};
 use reth_exex::ExExManagerHandle;
 use reth_network::{
     frost::{manager::FrostConfig, protocol::FrostProtoHandler},
@@ -771,12 +775,42 @@ impl<Ext: clap::Args + fmt::Debug> PoaNodeCommand<Ext> {
             .with_port(*cometbft_rpc_port)
             .with_host(cometbft_rpc_host);
 
+        let quorum;
+        let min_validator_count;
+        let target_height;
+        let our_vote;
+
+        // ActivationManager: setup parameters and conditions.
+        if *is_testnet {
+            quorum = 100; // 100% ~= 3/3 members must approve
+            min_validator_count = 3;
+            target_height = 1; // Activate as soon as possible
+            our_vote = Vote::Aye;
+        } else {
+            quorum = 100; // 100% ~= 15/15 members must approve
+            min_validator_count = 15;
+            target_height = 1; // Activate as soon as possible
+            our_vote = Vote::Aye;
+        }
+
+        // ActivationManager: setup compliance and vote inclusion.
+        let activation_manager =
+            ActivationManagerBuilder::new(VoteWatcher::default(), RUNTIME_VERSION_ACTIVE)
+                .build_COMPLIANT_network_upgrade(
+                    RUNTIME_VERSION_UPGRADE,
+                    quorum,
+                    min_validator_count,
+                    target_height,
+                    Some(our_vote),
+                );
+
         // Build authority Consensus
         let (abci_started_tx, abci_started_rx) = tokio::sync::oneshot::channel::<()>();
         let (frost_task, abci_client_builder, snapshot_manager, wallet_sync) =
             match AuthorityConsensusBuilder::try_new(
                 Arc::clone(&chain_arc.clone()),
                 blockchain_db.clone(),
+                activation_manager,
                 btc_server_factory.clone(),
                 bitcoin_block_header_clone.clone(),
                 secret_key,
--- a/bin/reth/src/commands/poa/mod.rs
+++ b/bin/reth/src/commands/poa/mod.rs
@@ -787,8 +787,8 @@ impl<Ext: clap::Args + fmt::Debug> PoaNodeCommand<Ext> {
             target_height = 1; // Activate as soon as possible
             our_vote = Vote::Aye;
         } else {
-            quorum = 100; // 100% ~= 15/15 members must approve
-            min_validator_count = 15;
+            quorum = 100; // 100% ~= 16/16 members must approve
+            min_validator_count = 16;
             target_height = 1; // Activate as soon as possible
             our_vote = Vote::Aye;
         }
--- a/crates/consensus/authority/src/activation_manager/mod.rs
+++ b/crates/consensus/authority/src/activation_manager/mod.rs
@@ -260,7 +260,7 @@ impl ConditionList {
 /// This builder provides methods to configure how a validator will handle
 /// network upgrades - whether it will ignore them, signal votes without
 /// being compliant, or fully accept upgrades when conditions are met.
-pub struct ActivationManagerBuilder<DB> {
+pub struct ActivationManagerBuilder<DB, Auth> {
     /// The database client implementing ActivationManagerReaderWriter
     client: DB,
 
@@ -272,11 +272,13 @@ pub struct ActivationManagerBuilder<DB> {
 
     /// Configuration for a pending network upgrade, if any
     upgrade: Option<NetworkUpgrade>,
+
+    _p: std::marker::PhantomData<Auth>,
 }
 
-impl<DB> ActivationManagerBuilder<DB>
+impl<DB, Auth> ActivationManagerBuilder<DB, Auth>
 where
-    DB: ActivationManagerReaderWriter,
+    DB: ActivationManagerReaderWriter<Auth>,
 {
     /// Creates a new ActivationManagerBuilder with specified database client and active version.
     ///
@@ -295,6 +297,7 @@ where
             active_version,
             vote_retention_period: VOTE_RETENTION_PERIOD,
             upgrade: None,
+            _p: std::marker::PhantomData,
         }
     }
 
@@ -327,7 +330,7 @@ where
     ///
     /// # Returns
     /// * An ActivationManager configured to ignore network upgrades
-    pub fn build_ignore_nework_upgrade(self) -> ActivationManager<DB> {
+    pub fn build_ignore_nework_upgrade(self) -> ActivationManager<DB, Auth> {
         debug_assert!(self.upgrade.is_none());
         self._finalize()
     }
@@ -356,7 +359,7 @@ where
         mut self,
         upgrade_version: RuntimeVersion,
         our_vote: Vote,
-    ) -> ActivationManager<DB> {
+    ) -> ActivationManager<DB, Auth> {
         assert!(self.active_version < upgrade_version);
 
         #[rustfmt::skip]
@@ -408,7 +411,7 @@ where
         min_validator_count: usize,
         target_height: u64,
         our_vote: Option<Vote>,
-    ) -> ActivationManager<DB> {
+    ) -> ActivationManager<DB, Auth> {
         assert!(
             self.active_version < upgrade_version,
             "upgrade version is older than the active version"
@@ -437,12 +440,13 @@ where
         self._finalize()
     }
 
-    fn _finalize(self) -> ActivationManager<DB> {
+    fn _finalize(self) -> ActivationManager<DB, Auth> {
         ActivationManager {
             client: self.client,
             vote_retention_period: self.vote_retention_period,
             active_version: Arc::new(RwLock::new(self.active_version)),
             upgrade: Arc::new(RwLock::new(self.upgrade)),
+            _p: std::marker::PhantomData,
         }
     }
 }
@@ -467,7 +471,7 @@ where
 /// state through atomic reference counting and read-write locks, allowing it to
 /// be shared between concurrent contexts.
 #[derive(Clone)]
-pub struct ActivationManager<DB> {
+pub struct ActivationManager<DB, Auth> {
     /// Database client for persisting and retrieving votes
     client: DB,
 
@@ -481,11 +485,13 @@ pub struct ActivationManager<DB> {
     /// Configuration for a pending network upgrade, if any
     /// Protected by a read-write lock for thread-safe updates
     upgrade: Arc<RwLock<Option<NetworkUpgrade>>>,
+
+    _p: std::marker::PhantomData<Auth>,
 }
 
-impl<DB> ActivationManager<DB>
+impl<DB, Auth> ActivationManager<DB, Auth>
 where
-    DB: ActivationManagerReaderWriter,
+    DB: ActivationManagerReaderWriter<Auth>,
 {
     /// Prepares a proposal decision based on the current upgrade state.
     ///
@@ -633,7 +639,7 @@ where
         &self,
         block_version: RuntimeVersion,
         block_height: u64,
-        proposer_address: secp256k1::PublicKey,
+        proposer_address: Auth,
         proposer_vote: Option<NetworkUpgradePayload>,
     ) -> ProviderResult<OnFinalizeBlockDecision> {
         // Track the proposer's vote for upgrade intention, if provided
@@ -757,7 +763,7 @@ where
     fn _track_vote(
         &self,
         block_height: u64,
-        addr: secp256k1::PublicKey,
+        addr: Auth,
         vote: Option<NetworkUpgradePayload>,
     ) -> ProviderResult<()> {
         // If the proposer made a vote...
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10412,7 +10412,6 @@ dependencies = [
  "reth-trie",
  "reth-trie-db",
  "revm",
- "secp256k1 0.29.1",
  "strum 0.26.3",
  "tempfile",
  "tokio",
--- a/crates/consensus/authority/src/activation_manager/tests/mod.rs
+++ b/crates/consensus/authority/src/activation_manager/tests/mod.rs
@@ -1,8 +1,7 @@
 use super::*;
-use rand::thread_rng;
 use reth_provider::ProviderResult;
-use secp256k1::generate_keypair;
 use std::{collections::HashMap, sync::Arc};
+use utils::Address;
 
 mod accept_lagging_validator;
 mod reject_ignored_upgrade;
@@ -18,7 +17,7 @@ mod wait_for_compliance;
 /// In-Memory database that integrates the `ActivationManagerReaderWriter` trait.
 #[derive(Clone)]
 struct Db {
-    votes: Arc<RwLock<HashMap<secp256k1::PublicKey, VoteEntry>>>,
+    votes: Arc<RwLock<HashMap<Address, VoteEntry>>>,
 }
 
 impl Db {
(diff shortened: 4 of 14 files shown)


rwlockbg (dismissed):

lamafab:
Alright - ran a bunch of tests on local dev-net on my side with four Fed member; checked version mismatches, vote mismatches, reboots, etc and the transition was always smooth and uninterrupted:

Block header verification via RPC (`eth_getBlockByNumber` for **block 143**) - note the `"baseFeePerGas"`:

```json
{
  "jsonrpc": "2.0",
  "id": 83,
  "result": {
    "hash": "0x3fdc916f6a20e838e165b52fd250e3eb3360c59d1fd505cd03672282e059f03a",
    "parentHash": "0x5f58b7c8788e99cc713e921d401419e8a47874a348c8d10000e9f6ad1776e0e5",
    "sha3Uncles": "0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347",
    "miner": "0x0000000000000000000000000000000000000000",
    "stateRoot": "0x28e9993cac543c42803a99af9530f24b9524e6432e80f0c9227e17f433c16d44",
    "transactionsRoot": "0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421",
    "receiptsRoot": "0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421",
    "logsBloom": "0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
    "difficulty": "0x0",
    "number": "0x8f",
    "gasLimit": "0x1c9c380",
    "gasUsed": "0x0",
    "timestamp": "0x68781918",
    "totalDifficulty": "0x0",
    "extraData": "0x0000000000000000228f8eab657fab845e77d8a0060e5e9c2f7172c9675a332dd95f126545d6e03303b3d6b59405ffcf4e0e9773a8834bf0540186f2440526dafead940db055870ba9471737851d3a40c61b7dd66e75c394b59f8d051f",
    "mixHash": "0x0000000000000000000000000000000000000000000000000000000000000000",
    "nonce": "0x0000000000000000",
    "baseFeePerGas": "0x7",
    "blobGasUsed": "0x0",
    "excessBlobGas": "0x0",
    "parentBeaconBlockRoot": "0x0000000000000000000000000000000000000000000000000000000000000000",
    "uncles": [],
    "transactions": [],
    "size": "0x280"
  }
}
```

Block header verification via RPC (`eth_getBlockByNumber` for **block 144**) - note the `"baseFeePerGas"`:


```json
{
  "jsonrpc": "2.0",
  "id": 83,
  "result": {
    "hash": "0x32b7b82d3bef9f6aa7391988e744a8765bfc6956ef79cc369f069c74ed1b9733",
    "parentHash": "0x3fdc916f6a20e838e165b52fd250e3eb3360c59d1fd505cd03672282e059f03a",
    "sha3Uncles": "0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347",
    "miner": "0x0000000000000000000000000000000000000000",
    "stateRoot": "0x28e9993cac543c42803a99af9530f24b9524e6432e80f0c9227e17f433c16d44",
    "transactionsRoot": "0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421",
    "receiptsRoot": "0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421",
    "logsBloom": "0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
    "difficulty": "0x0",
    "number": "0x90",
    "gasLimit": "0x1c9c380",
    "gasUsed": "0x0",
    "timestamp": "0x6878191d",
    "totalDifficulty": "0x0",
    "extraData": "0x0000000000000000228f8eab657fab845e77d8a0060e5e9c2f7172c9675a332dd95f126545d6e03303b3d6b59405ffcf4e0e9773a8834bf0540186f2440526dafead940db055870ba9471737851d3a40c61b7dd66e75c394b59f8d051f",
    "mixHash": "0x0000000000000000000000000000000000000000000000000000000000000000",
    "nonce": "0x0000000000000000",
    "baseFeePerGas": "0x4c4b40",
    "blobGasUsed": "0x0",
    "excessBlobGas": "0x0",
    "parentBeaconBlockRoot": "0x0000000000000000000000000000000000000000000000000000000000000000",
    "uncles": [],
    "transactions": [],
    "size": "0x283"
  }
}
```

Mempool fee history via RPC (`eth_feeHistory`)

```json
{
    "jsonrpc": "2.0",
    "id": 83,
    "result": {
        "baseFeePerGas": [
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x7",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x4c4b40",
            "0x42c1d8"
        ],
        "gasUsedRatio": [
            "..."
        ],
        "baseFeePerBlobGas": [
            "..."
        ],
        "blobGasUsedRatio": [
            "..."
        ],
        "oldestBlock": "0x5e",
        "reward": [
            "..."
        ]
    }
}
```

I'd say this is good to go!

bin/reth/src/commands/poa/mod.rs
```diff
@@ -771,12 +775,42 @@ impl<Ext: clap::Args + fmt::Debug> PoaNodeCommand<Ext> {
...
+
+        // ActivationManager: setup parameters and conditions.
+        if *is_testnet {
+            quorum = 100; // 100% ~= 3/3 members must approve
+            min_validator_count = 3;
+            target_height = 1; // Activate as soon as possible
+            our_vote = Vote::Aye;
+        } else {
+            quorum = 100; // 100% ~= 15/15 members must approve
+            min_validator_count = 15;
```
scottmillner: there are 16 federation members

crates/consensus/authority/src/activation_manager/mod.rs
```diff
@@ -272,11 +272,13 @@ pub struct ActivationManagerBuilder<DB> {
 
     /// Configuration for a pending network upgrade, if any
     upgrade: Option<NetworkUpgrade>,
+
+    _p: std::marker::PhantomData<Auth>,
 }
 
-impl<DB> ActivationManagerBuilder<DB>
+impl<DB, Auth> ActivationManagerBuilder<DB, Auth>
```
SatwikPrabhu, line 279: I think you would need to add generic trait bounds to the `Auth` parameter to avoid any potential issues if something else other than hashset will be used. Same comment for all other places with `Auth`

bin/reth/src/commands/poa/mod.rs
SatwikPrabhu: Would it make sense to make these configurable instead of hardcoding?

SatwikPrabhu:
Overall, I think this looks great to me. Good job @lamafab 🔥 . Just a couple of minor comments added from my side. Feel free to pushback on them!

crates/consensus/authority/src/activation_manager/mod.rs
```diff
@@ -408,7 +411,7 @@ where
         min_validator_count: usize,
         target_height: u64,
         our_vote: Option<Vote>,
-    ) -> ActivationManager<DB> {
+    ) -> ActivationManager<DB, Auth> {
```
shumkov, line 414: `MIN_QUORUM` should be 68 because CometBFT majority is 2/3 + 1

crates/consensus/authority/src/activation_manager/mod.rs
```diff
@@ -408,7 +411,7 @@ where
         min_validator_count: usize,
         target_height: u64,
         our_vote: Option<Vote>,
```
shumkov, line 413: Why do we need to vote on minor version (non breaking change)? People can install new version of software or not (if they don't want)

shumkov:
> we do not yet know at what timestamp the transition takes place - that's only something we can determine and hard-code retroactively.

To solve this, we should deterministically decide on which block height or timestamp we upgrade. To do so, we can use epochs, so when we have a threshold of votes, we lock in this protocol version, and it should switch on automatically at the next epoch. We also could utilize app_version in consensus params to persist current version in cometbft chain

I have a few questions:
1. How do we validate votes and switch to new versions if they aren't part of consensus?
2. How do we persist the current version in time to handle historical sync?
3. Why do we need MinorVersion? It doesn't affect consensus so you can update software or stay on the same version without any problems.

shumkov (dismissed):

lamafab:
@shumkov 

> To solve this, we should deterministically decide on which block height or timestamp we upgrade. To do so, we can use epochs, so when we have a threshold of votes, we lock in this protocol version, and it should switch on automatically at the next epoch.

So we do have that `target_height` parameter that sets the height at which the upgrade _should_ happen, but that can be pushed back if not all requirements are met. Usually - and in a permissionless context - when you can plan ahead weeks or months, that would be acceptable. But in our case we want to make sure that **every** Fed member is ready ASAP, which is why we push that quorum requirement to 100%. And considering that rollouts to our Fed member has historically been slow and messy, we don't bother thinking too much about`target_height`.

> 1. How do we validate votes and switch to new versions if they aren't part of consensus?

This mechanism purely tracks votes of Fed member (and in the future, DynaFed), but all nodes do count votes. Generally, to measure sentiment you must wait for some time until all validators get selected by CometBFT for block production; casting votes **only** happens in `ABCI::prepare_proposal` (only Fed members), and counting votes **only** happens in `ABCI::finalize_block` (all nodes).

> 2. How do we persist the current version in time to handle historical sync?

See this commit: https://github.com/botanix-labs/Macbeth/pull/841/commits/5b9cd1fc247ff34b83c3f34e211f5e09e3092711 - The `BlockWithContext` structure now comes with a `runtime_version` parameter, which is pulled from the NDD. In other words, the block proposer signals what runtime version is used, and the other Fed members validate that value. See `ActivationManager::on_process_proposal()`:

```rust
/// Processes an incoming block proposal and determines if it should be accepted.
///
/// This method is called during CometBFT's `process_proposal` phase to
/// determine whether to accept or reject a block proposal based on its
/// version and the current upgrade state.
///
/// ## Behavior
///
/// - Accepts blocks with the active version
/// - Accepts blocks with an upgrade version if the validator is compliant with the upgrade AND
///   all conditions are met
/// - Rejects blocks with unknown versions or when conditions aren't met
///
/// # Parameters
/// * `block_height` - The height of the block being processed
/// * `block_version` - The runtime version of the block being processed
///
/// # Returns
/// * `OnProcessProposalDecision::Process` if the block should be accepted
/// * `OnProcessProposalDecision::RejectBlock` if the block should be rejected
pub fn on_process_proposal(
	&self,
	block_height: u64,
	block_version: RuntimeVersion,
) -> ProviderResult<OnProcessProposalDecision>;
```

Additionally, do have a look at `ActivationManager::on_finalize_block()`; it specifically skips requirements checks:

```rust
// NOTE: Upgrade condition checks are deliberately omitted here. These
// validations have already occurred during the backing phase in
// `on_process_proposal`. When CometBFT finalizes an upgraded block, it
// confirms that the required majority of validators have endorsed the
// upgrade. Additionally, syncing nodes lack historical context about
// previous forks (including migrations and code changes), preventing
// them from independently validating past version transitions.
// Consequently, any block finalized by CometBFT must be accepted by all
// nodes.
//
// However, nodes can resist future upgrades, and their consequent
// behavior is considered undefined.
```

> Why do we need MinorVersion? It doesn't affect consensus so you can update software or stay on the same version without any problems.

It's just semantics; _major_ implies hard-fork, _minor_ implies soft-fork. While you could argue that soft-forks are backwards-compatible, they can still fracture the ecosystem where some nodes support an operation and other don't. We sill want to measure sentiment and community support for soft-forks.

crates/consensus/authority/src/activation_manager/mod.rs
lamafab: Yeah fair enough, but I just realized that changing it from `67` to `68`  screws up my unit tests where we do 2/3 Fed member tests 😬 Lets say that as the hard-coded minimum value, `MIN_QUORUM = 67` is acceptable.

crates/consensus/authority/src/activation_manager/mod.rs
lamafab: Explained in my comment below.

scottmillner:
utACK

really nice job!! 🔥 🔥 🔥 

had a small comment but it has no impact since we're forcing all fed members to upgrade immediately

going to test now

shumkov:
@lamafab 
2. So we always have an active version persistent in the blockchain, and we know it for each block, right?
4. If the node doesn't know about the next version, it will vote Nope, right? And when we switch to a new version, this outdated  node just halts because it won't be able to process a block with the new version, which it isn't aware of?

bin/reth/src/commands/poa/mod.rs
lamafab: When actually rolling out software that is specifically designed to handle the upgrade migration, those values _should_ be hard-coded such that they're equal for all participants. Also do note that we've had problems in the past when we asked Fed members to specify certain parameters ^_^

lamafab:
@shumkov 

> 2. So we always have an active version persistent in the blockchain, and we know it for each block, right?

Correct! Do note that for the "active version" before this PR, we retroactively apply this value:

```rust
// -> crates/consensus/authority/src/comet_bft/non_deterministic_data.rs

/// The implied Botanix runtime version at mainnet launch, created
/// retroactively.
pub(crate) const GENESIS_RUNTIME_VERSION: RuntimeVersion = RuntimeVersion::new(0, 1);
```

> 4. If the node doesn't know about the next version, it will vote Nope, right? And when we switch to a new version, this outdated node just halts because it won't be able to process a block with the new version, which it isn't aware of?

Correct! Also have a look at `ActivationManagerBuilder`; it essentially has three different modes.

```rust
/// Finalizes the builder to create an ActivationManager that ignores
/// network upgrades.
///
/// This creates a manager that will:
/// * Not participate in upgrade voting
/// * Not include any NetworkUpgradePayload in proposals
/// * Reject any blocks with versions other than the current active version
///
/// This is the default operating mode for nodes that are not participating
/// in the upgrade process and will result in the node rejecting blocks if
/// an upgrade is activated on the network.
/// ...
pub fn build_ignore_nework_upgrade(self) -> ActivationManager<DB, Auth>;

/// Finalizes the builder to create an ActivationManager that signals a vote
/// on an upgrade.
///
/// This creates a manager that will:
/// * Include the specified vote in the NetworkUpgradePayload of proposals
/// * Still reject any blocks with the upgrade version (is_compliant = false)
/// * Track votes from other validators
///
/// This mode allows validators to signal their support or opposition to an
/// upgrade before they are ready to accept upgraded blocks.
/// ...
pub fn build_signal_network_upgrade(self) -> ActivationManager<DB, Auth>;

/// Finalizes the builder to create an ActivationManager that accepts an upgrade when conditions
/// are met.
///
/// This creates a manager that will:
/// * Include the specified vote in the NetworkUpgradePayload of proposals
/// * Accept and process blocks with the upgrade version once all conditions are met
/// * Track votes from other validators and calculate approval rates
/// * Transition to the upgrade version automatically when an upgraded block is finalized
///
/// This mode should be used when validators are prepared for an upgrade and have
/// the necessary code and migrations in place to handle the new version.
/// ...
pub fn build_COMPLIANT_network_upgrade(self) -> ActivationManager<DB, Auth>;
```

crates/consensus/authority/src/activation_manager/mod.rs
lamafab: We really don't need to add any bounds here; the `ActivationManagerReaderWriter` (database layer) decides what the actual type is and how it should be handled, since the `ActivationManager` just passes that value down to the database layer.

```rust
pub struct ActivationManager<DB, Auth>;

impl<DB, Auth> ActivationManager<DB, Auth>
where
    DB: ActivationManagerReaderWriter<Auth>,
{}
```

In our `ABCIClient`, we just use `ActivationManager<VoteWatcher, Address>`, for example.

crates/consensus/authority/src/activation_manager/mod.rs
shumkov: Up to you :)

shumkov (dismissed):
Good job! 💯

scottmillner (approved):

PR #843, opened 2025-07-16
fix/change outputs/add back missing outputs

scottmillner:

Pegout scheduler is incorrectly removing change outputs from the utxo set.  Another PR fixes this: https://github.com/botanix-labs/Macbeth/pull/842

But we still need to add those change outputs back to the utxo set. 

## What was done?

On startup, btc server.sync_until() will scan the L1 for unspent utxos for the multisig change address and add back any missing ones to the db.

## How Has This Been Tested?

- Added a unit test and tested locally

## Breaking Changes

- None

## Checklist

### DIFF
--- a/bin/btc-server/src/pegout_scheduler/mod.rs
+++ b/bin/btc-server/src/pegout_scheduler/mod.rs
@@ -12,7 +12,7 @@ use std::{
 };
 
 use crate::{
-    database::FinalizedPegout,
+    database::{FinalizedPegout, Utxo},
     pegout_id::PegoutId,
     telemetry::Telemetry,
     update_telemetry_error,
@@ -22,7 +22,7 @@ use crate::{
     },
 };
 use bitcoin::{Amount, Block, BlockHash, OutPoint, ScriptBuf, Transaction, TxOut, Txid};
-use bitcoincore_rpc::RpcApi;
+use bitcoincore_rpc::{json::ScanTxOutRequest, RpcApi};
 use log::{debug, error, info, trace, warn};
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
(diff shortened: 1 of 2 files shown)


bin/btc-server/src/pegout_scheduler/mod.rs
```diff
@@ -687,6 +690,65 @@ impl PegoutScheduler {
...
+        for unspent in change_utxos {
+            let utxo = Utxo {
+                outpoint: OutPoint { txid: unspent.txid, vout: unspent.vout },
+                output: TxOut { value: unspent.amount, script_pubkey: unspent.script_pub_key },
+                eth_address: None,     // Assuming this is None for change outputs
+                version: utxo_version, // Set to appropriate default or version
+            };
+
+            // Check if this UTXO already exists in the database
+            if !utxos.contains(&utxo) {
```
DariusParvin: It looks good to me but im a bit nervous that this might return false, not because it's not already in the db but because there's some slight mismatch in how we do the conversion, or maybe it's comparing the pointer rather than the raw values or something like that.

DariusParvin: maybe making the equality check specific to the OutPoint value might reduce the chance of that happening?

also, just to be safe (since it would be a hassle to undo any errors here), if the OutPoint _does_ exist in the db, perhaps we can do an equality check to make sure it's values are the same as the `utxo` we had just created, and if it's not then throw an error before adding the `missing_utxos`. That way i think we can be confident that we're adding the `missing_utxos` with the correct format.

scottmillner: it's not comparing the pointers
`contains()` is using `PartialEq` bc `struct Utxo` is using it with the derive macro

```
[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
pub struct Utxo {
```

DariusParvin (dismissed):
looks good! Just had a suggestion that might make the change safer to implement, but I'm happy to go with your judgement

bin/btc-server/src/pegout_scheduler/mod.rs
scottmillner: great suggestion
just pushed changes to do that

DariusParvin (approved):
🙌

PR #847, opened 2025-07-16
feat(reth): Created new crates botanix-chainspec and botanix-hardforks

rwlockbg:

Added 2 new crates:

- botanix-chainspec
- botanix-hardforks

Deprecated botanix-specific stuff from reth-chainspec and made use of botanix-chainspec throughout the entire code thus isolating the botanix-specific entities. The latter enables us to directly make use of the new crates as part of the reth-upgrades task.

## What was done?
- botanix-chainspec
- botanix-hardforks

Deprecated botanix-specific stuff from reth-chainspec and made use of botanix-chainspec throughout the entire code thus isolating the botanix-specific entities

## How Has This Been Tested?

CI/CD

## Breaking Changes

Some marginal changes to the reth evm execution crates as now the `BotanixChainSpec` had to be passed for direct use

## Checklist

## Summary by CodeRabbit

- New Features
  - Added Botanix chain specs (Mainnet/Testnet) with genesis templates and CLI/parser support for names, file paths, or raw JSON.
  - Introduced Botanix hardforks enum including Arbitrum Sepolia support and a Devnet option.

- Refactor
  - Public APIs now use a BotanixChainSpec wrapper; callers updated to access the inner chain spec and chain IDs.

- Chores
  - Unified askama to workspace, added botanix-chainspec dependency across crates; tests and examples updated to new imports.


crates/botanix-chainspec/genesis/botanix_mainnet.json
DariusParvin, line 1: just for my own understanding, what are these jsons used for? 

also, is there any way to validate the values. I'm a bit concerned that with so many mainnet config values moving, if there's a mistake somewhere, we would only find out after it's already deployed on mainnet. Is there any way we can be safe against such mistakes or we just need to manually check all the all values are the same as before?

rwlockbg: I have triple-checked them @DariusParvin. I am afraid there is no other validation than that. Your feedback here is priceless.

DariusParvin: okay yeah, i'll check them too against what's in develop just to be safe!

DariusParvin: I am checking the values too, but do you know what the risks are if there is a mismatch in one of the values? Would we be able to recover by just rolling back to the previous version, or could it be a super messy situation where we accidentally create a chain split? Maybe this is the only way, but i just want to get a sense of the risk here

bin/reth/src/commands/poa/mod.rs
```diff
@@ -598,7 +598,8 @@ impl<Ext: clap::Args + fmt::Debug> PoaNodeCommand<Ext> {
         // Config executor factory
         let evm_config = EthEvmConfig::default();
         let executor_factory = EthExecutorProvider::new(
-            Arc::new(chain.clone()),
```
scottmillner, line 601: - we're basically passing the same arg twice here: `chain` is the same and we could just do the conversions inside `new()`

- also, im not understanding why `EthExecutorProvider` needs both `chain_spec` and `botanix_chain_spec` since `botanix_chain_spec.inner` is `chain_spec`

maybe some comments to explain would be helpful?

DariusParvin (approved):
If i understand correctly, this more like a 'refactor' rather than a 'feat' right? 

I double checked the new chainspec values and they matched as far as i could tell 👍  

Im wondering if it's possible to have some kind of configuration regression test that does the following: instantiates a mainnet node, dumps all the configuration settings into a temporary file, then compares that to an existing file of what we expect. 

That way we can be sure that this big refactor didn't accidentally change something.

For me ,without understanding exactly how bad the consequences of having an incorrect configuration is, my inclination is that we should have such a regression test, but I'm happy to defer to other's judgement if it isn't worth delaying these changes for that.

DariusParvin (dismissed):

DariusParvin (commented):
sorry my previous review was meant as a 'comment' rather than 'approve'

crates/botanix-chainspec/genesis/botanix_mainnet.json
rwlockbg: Well, this PR does not do much apart from wrapping the old chainspec into a botanixchainspec and using it instead. It does a bit of reorganization though as the botanix-chainspec-related props are now a separate crate and is no longer part of the old chainspec (the reth crate). The new reth architecture needs exactly this change so I have made sure its properly composed. If the values for both Botanix testnet and mainnet are correct I dont see any danger. Also, since we dont care about hardforks etc. the values for arbitrum, celo and all the other networks are irrelevant to us. cc @scottmillner @DariusParvin @shumkov

bin/reth/src/commands/poa/mod.rs
rwlockbg: @scottmillner Good observation, was not needed. I fixed that as the one latter already contains the Chainspec via inner(). Thanks for spotting this.

crates/botanix-chainspec/src/lib.rs
```diff
@@ -0,0 +1,128 @@
...
+            bitcoin_checkpoint_confirmation_depth: 0,
+            weak_bitcoin_checkpoints_count: 0,
+            historical_bitcoin_checkpoints_count: 0,
+            epoch_length: 0,
+        }
+    }
+}
+
+impl BotanixChainSpec {
+    pub fn inner(&self) -> &ChainSpec {
```
shumkov, line 81: Won't it be better to implement `Deref`?

crates/botanix-chainspec/src/lib.rs
```diff
@@ -0,0 +1,128 @@
...
+
+use crate::constants::{BOTANIX_INITIAL_BASE_FEE, BOTANIX_TESTNET};
+pub mod constants;
+pub mod parser;
+
+/// Botanix chain spec type.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct BotanixChainSpec {
+    /// [`ChainSpec`].
+    pub inner: ChainSpec,
```
shumkov, line 15: From a coding pattern point of view, it's inner. Shall we call it `base` since it's the base properties of the chain?

crates/botanix-hardforks/src/lib.rs
```diff
@@ -0,0 +1,486 @@
...
+
+        None
+    }
+
+    /// Retrieves the activation block for the specified hardfork on the Ethereum mainnet.
+    pub const fn mainnet_activation_block(&self) -> Option<u64> {
+        match self {
+            Self::Frontier => Some(0),
+            Self::Homestead => Some(1150000),
+            Self::Dao => Some(1920000),
```
shumkov, line 81: Are these block heights actual for us?

shumkov (changes requested):

crates/botanix-chainspec/src/lib.rs
rwlockbg: I dont like the base name. It feels weird here. I suggest we keep it inner as it clearly shows that the chainspec is an inner property ?

crates/botanix-chainspec/src/lib.rs
rwlockbg: Might be a good idea, indeed. I will check it.

crates/botanix-hardforks/src/lib.rs
rwlockbg: We dont really care about these at the moment as we dont have any hardforls. I took those values from the MAINNET as we were using them before.

scottmillner (dismissed):

DariusParvin (dismissed):

crates/botanix-chainspec/src/lib.rs
rwlockbg: I have added he Deref, but have explicitly not changed it in the code as inner() gives better indication as to where the Chainspec is coming from. Deref is more for wrapped type structures like `Address(pub String)` where we want to outer type to give us direct ref to the inner by *, here the inner type is quite complex and ChainSpec is just one part of it.

crates/botanix-hardforks/src/lib.rs
shumkov: I just wondering if some hardfork will be enabled for us unexpectedly on specific height?

shumkov (approved):

crates/botanix-chainspec/src/lib.rs
shumkov: Oh I see. Thank you

crates/botanix-chainspec/src/lib.rs
shumkov: I don't have strong preferences.

DariusParvin (approved):

scottmillner (approved):

PR #852, opened 2025-07-17
Feat: Emergency wallet sweeping feature

SatwikPrabhu:
PR for the emergency wallet sweeping feature. 

Checklist:

- [Feature(wallet sweeping): Create emergency tool and UTXO discovery #1078](https://github.com/botanix-labs/botanix/issues/1078) -- Done
- [Feat(wallet sweeping): Import/export utxo set and comparison #1079](https://github.com/botanix-labs/botanix/issues/1079) -- Done
- [Feat(wallet sweeping): Sweep PSBT construction #1080](https://github.com/botanix-labs/botanix/issues/1080) -- Done
- [Feat(wallet sweeping): Introduce new sweep event #1081](https://github.com/botanix-labs/botanix/issues/1081)
- [Feat(wallet sweeping): sweep event inititation #1082](https://github.com/botanix-labs/botanix/issues/1082)
- [Feat(wallet sweeping): Integrate sweep initiation into emergency tool #1083](https://github.com/botanix-labs/botanix/issues/1083)
- [Feat(wallet sweeping): Adapt commit-l1 call to include sweep event #1084](https://github.com/botanix-labs/botanix/issues/1084)
- [Feat(wallet sweeping): Graceful recovery and monitoring #1085](https://github.com/botanix-labs/botanix/issues/1085)

## Summary by CodeRabbit

* **New Features**
  * Introduced a command-line emergency tool for Bitcoin UTXO management and emergency sweep PSBT creation, supporting manual and consensus-based modes with detailed validation and reporting.
  * Added configuration options for Bitcoin node RPC connection and federation coordination.
  * Enabled listing, exporting, and comparing UTXO sets, computing consensus safe subsets, and constructing sweep transactions with audit reports.
  * Added a comprehensive engineering specification document detailing the emergency wallet sweep design, workflow, and future Trusted Execution Environment (TEE) integration.
  * Extended btc-server with new emergency coordination messages and an RPC endpoint to authorize emergency signing.

* **Bug Fixes**
  * None.

* **Style**
  * Improved formatting and readability in contract ABI binding source files; no changes to logic or functionality.

PR #876, opened 2025-07-25
docs: Add emergency wallet sweep engineering specification

SatwikPrabhu:

This PR adds a comprehensive engineering specification for the Emergency Wallet Sweep feature, which addresses critical security scenarios in the Botanix federation. The specification is needed to:

- Provide a standardized approach for emergency fund consolidation when federation keys are compromised
- Enable proactive key rotation when federation members go offline and quorum is at risk
- Establish a secure, auditable protocol for coordinating emergency operations across federation members
- Ensure forward compatibility with future TEE integration

## What was done?

- Added `docs/design/emergency_wallet_sweep_spec.md`

## How Has This Been Tested?

-

## Breaking Changes

-

## Checklist

## Summary by CodeRabbit

* **Documentation**
  * Added a detailed specification document outlining the Emergency Wallet Sweep feature for the Bitcoin wallet, including workflow phases, operator controls, CLI tool commands, and future TEE integration plans.

* **New Features**
  * Introduced emergency wallet sweep functionality enabling secure manual fund transfer under emergency conditions.
  * Added CLI commands to initiate, accept, and sign emergency sweep transactions.
  * Implemented secure coordinator-peer communication with consensus-based UTXO selection and threshold signing.
  * Extended the Bitcoin server with new endpoints for emergency session authorization and UTXO state retrieval.

### DIFF
--- a/docs/design/emergency_wallet_sweep_spec.md
+++ b/docs/design/emergency_wallet_sweep_spec.md
@@ -1,0 +1,144 @@
+**This file provides a comprehensive, auditable, and forward-compatible specification for the Botanix Emergency Wallet Sweep feature.**
+
+# Emergency Wallet Sweep: Engineering Specification (v1.0)
+
+## 1. Introduction & Guiding Principles
+
+This document specifies the design and implementation of an emergency wallet sweep feature for the Botanix federation. This is a critical security and operational tool of last resort, designed to secure all funds in the federation's Bitcoin wallet by transferring them to a new, safe address.
+
+The design is governed by the following non-negotiable principles:
+
+- **Zero Public Exposure**: The mechanism will not introduce any new public-facing endpoints or on-chain events. All coordination communication will be encrypted and authenticated.
+- **Deliberate Operator Control**: The sweep is a manual procedure, initiated and controlled by human operators through a dedicated command-line tool.
+- **Leverage Proven Infrastructure**: The solution will be built upon the `btc-server`'s existing, battle-tested components, including its database, secure peer-to-peer communication channels, and FROST threshold signing engine.
+- **Forward Compatibility**: The design is modular and anticipates future integration with a Trusted Execution Environment (TEE).
+
+---
+
+## 2. Core Scenarios & Threat Model
+
+This feature is designed to address two primary emergency scenarios:
+
+- **Scenario A: Active Key Compromise / Imminent Threat**  
+  A situation where one or more of the federation's signing keys are believed to be compromised, requiring an immediate consolidation of funds to prevent theft.
+
+- **Scenario B: Proactive Key Rotation due to Loss of Quorum**  
+  A scenario where multiple federation members have gone permanently offline, risking the loss of signing quorum (`t` of `n`). A sweep allows the remaining members to move funds to a new wallet controlled by a reconstituted federation.
+
+---
+
+## 3. High-Level Architecture: Integrated Emergency Coordination
+
+To address the challenges of divergent UTXO sets and offline members, this feature uses a Coordinator-Peer model facilitated by the existing DKG communication channel.
+
+- **Secure Communication**: Operators will use new commands that send signed, encrypted messages through the existing `NewDkgPayload`/`GetDkgPayloads` gRPC endpoints. This provides an auditable, integrated, and robust channel for coordination.
+- **Coordinator-Led Consensus**: To resolve inevitable state discrepancies between members, the Coordinator is responsible for collecting the state of all peers, computing a "safe subset" of UTXOs that have reached consensus, and generating an auditable report of any excluded funds.
+- **Coordinator-Generated PSBT**: The designated Coordinator generates a single, definitive Partially Signed Bitcoin Transaction (PSBT) for the sweep based on the consensus UTXO set. This ensures all members sign the exact same transaction.
+- **Threshold-Based Signing**: The existing FROST signing engine naturally handles offline or lost members, proceeding as long as the minimum signature threshold (t) is met.
+
+---
+
+## 4. The Step-by-Step Emergency Workflow
+
+### (All Members) Phase 1: Report Local State
+
+- Each federation member operator (including the coordinator) runs a command to broadcast their view of the wallet's state:  
+  `emergency-tool sweep report-state`
+- This command reads the local `btc-server` database to get its list of UTXOs, constructs a signed `ReportState` message, and sends it to the designated Coordinator via the `NewDkgPayload` gRPC endpoint.
+
+### (Coordinator) Phase 2: Build Consensus and Initiate Sweep
+
+- The designated Coordinator operator runs the `initiate` command with a parameter to define consensus:  
+  `emergency-tool sweep initiate --destination <SECURE_ADDR> --fee-rate <RATE> --consensus-threshold 80`
+- This command performs several critical actions:
+    1. **Collect States**: It calls `GetDkgPayloads` to retrieve all `ReportState` messages from peers.
+    2. **Compute Safe Subset**: It uses the collected UTXO lists as input to a `compute_safe_subset` function, which produces a final set of UTXOs recognized by at least the percentage of members defined by `--consensus-threshold`.
+    3. **Generate Audit Report**: The function generates a detailed `excluded_utxos.json` report listing every UTXO that did not meet the threshold and which members reported it. This file is saved locally for auditing.
+    4. **Generate Definitive PSBT**: It uses only the "safe subset" of consensus UTXOs to generate the single, definitive PSBT for the sweep.
+    5. **Broadcast `StartEmergencySweep`**: It constructs and broadcasts a `StartEmergencySweep` message containing the definitive PSBT and a hash of the audit report.
+
+### (Peers) Phase 3: Acknowledge and Validate
+
+- Peer operators run a command to check for and process incoming coordination messages:  
+  `emergency-tool sweep check-messages`
+- Upon receiving a valid `StartEmergencySweep` message, each Peer's tool:
+    1. Validates the Coordinator's signature.
+    2. Validates that every UTXO used as an input in the definitive PSBT exists in its own local database. If the peer is missing a required input, it cannot sign and will raise a critical error.
+    3. Sends an `AcknowledgeEmergencySweep` message back to the Coordinator via `NewDkgPayload`.
+
+### (Coordinator) Phase 4: Distribute Definitive PSBT and Sign
+
+- The Coordinator's `emergency-tool` collects the `AcknowledgeEmergencySweep` messages. Once a sufficient number of peers have acknowledged, it submits its definitive PSBT to its local `btc-server` to start the FROST signing process.
+
+### (All) Phase 5: Unified Threshold Signing
+
+- All acknowledged Peers fetch the definitive PSBT from the Coordinator and join the FROST signing session. This follows the existing, unmodified signing flow. As soon as `t` signatures are collected, the transaction is finalized and broadcast.
+
+---
+
+## 5. Detailed Implementation Specification
+
+### Epic 1: `emergency-tool` — The Secure Coordination CLI
+
+- **[ET-DB-01] Implement Database-as-a-Source**
+    - **Task**: The tool must read UTXOs directly from a `btc-server` sled database.
+    - **Interface**: Add a `--db-path <PATH>` argument.
+
+- **[ET-CONSENSUS-01] Implement UTXO Consensus Logic**
+    - **Task**: Implement the `compute_safe_subset` function as described in the recent commit. It must take a list of UTXO sets from members and a consensus threshold, and output one "safe" list of UTXOs and one `excluded_utxos.json` report.
+
+- **[ET-PSBT-01] Implement High-Integrity PSBT Construction**
+    - **Task**: The PSBT creation logic must correctly populate all required proprietary and witness fields (`witness_utxo`, `eth_address`, `version`) using the consensus UTXO list.
+
+- **[ET-COMMS-01] Implement Emergency Message Types**
+    - **Task**: Define new serializable structs for emergency coordination, to be wrapped in the `DkgPayload`.
+    - **Structs**:
+        - `ReportState { member_id: u16, utxos: Vec<Utxo>, signature: Vec<u8> }`
+        - `StartEmergencySweep { session_id: [u8; 32], destination_address: String, fee_rate: u64, definitive_psbt: Vec<u8>, audit_report_hash: [u8; 32], signature: Vec<u8> }`
+        - `AcknowledgeEmergencySweep { session_id: [u8; 32], member_id: u16, signature: Vec<u8> }`
+
+- **[ET-CLI-01] Implement the `sweep` Subcommands**
+    - `emergency-tool sweep report-state`: For all members. Reads local UTXOs and sends a `ReportState` message.
+    - `emergency-tool sweep initiate ...`: For the Coordinator. Collects `ReportState` messages, runs consensus logic, generates PSBT, and broadcasts `StartEmergencySweep`.
+    - `emergency-tool sweep check-messages`: For Peers. Calls `GetDkgPayloads`, validates messages, and sends `AcknowledgeEmergencySweep`.
+    - `emergency-tool sweep sign ...`: For all acknowledged members. Authorizes the local `btc-server` to participate in the FROST signing.
+
+---
+
+### Epic 2: `btc-server` — Minor Adaptations for Emergency Payloads
+
+- **[BS-PAYLOAD-01] Handle New Emergency Message Types**
+    - **Task**: In the `new_dkg_payload` gRPC method, add logic to recognize and handle the `ReportState`, `StartEmergencySweep`, and `AcknowledgeEmergencySweep` message types.
+    - **Logic**: Instead of processing them through the DKG state machine, simply store them in a new, separate database table/tree indexed by `session_id`. This makes them retrievable by the `emergency-tool`.
+
+- **[BS-RPC-01] Implement an `AuthorizeEmergencySigning` Endpoint**
+    - **Task**: Create a new, simple, authenticated gRPC endpoint.
+    - **Endpoint**: `rpc AuthorizeEmergencySigning(AuthorizeRequest) returns (AuthorizeResponse);`
+    - **Logic**: This endpoint takes a `session_id`. When called by an authenticated operator (via the `emergency-tool sweep sign` command), it flips a switch that "unlocks" the corresponding emergency PSBT, allowing it to be used in the standard `GetRound1/2SigningPackage` flow.
+
+---
+
+## 6. Future Work: Adaptation for TEE Integration
+
+The implementation of the Pegout Proofs and TEE Environment will shift the locus of trust for signing from the FROST engine within `btc-server` to the isolated TEE. The emergency sweep feature must be adapted accordingly.
+
+The `emergency-tool` will evolve from a signing *coordinator* to a TEE *payload constructor*. The `btc-server` will become a facilitator for the TEE.
+
+The following tasks from the v1.0 specification would need to be re-implemented:
+
+- **Re-implement [ET-PSBT-01] as [TEE-PAYLOAD-01]: Construct TEE `p_L1` Payload**
+    - The tool's primary responsibility will no longer be to create a simple PSBT. It must construct the entire `p_L1` input parameter required by the TEE's `commit-l1` function. This involves gathering not just UTXOs, but also the required Botanix blocks, validator sets, and state proofs (`P_x`, `P_d`). The emergency sweep itself would need to be defined as a transaction type the TEE understands is valid without a corresponding user-initiated pegout.
+
+- **Re-implement [ET-COMMS-01]: Evolve `StartEmergencySweep` Message**
+    - The `StartEmergencySweep` message will need to be expanded to carry all the data necessary for every peer to deterministically reconstruct an identical `p_L1` payload to present to their local TEE. This is far more comprehensive than just a PSBT.
+
+- **Re-implement [ET-CLI-01]: Reroute `sweep sign` Command**
+    - The `emergency-tool sweep sign` command will no longer call an authorization endpoint on `btc-server`. Instead, it will submit the final, agreed-upon `p_L1` payload directly to the local TEE instance for validation and signing.
+
+- **Replace [BS-RPC-01] with [TEE-RPC-01]: Implement `RequestTEESign` Endpoint**
+    - The `AuthorizeEmergencySigning` RPC endpoint on `btc-server` will be deprecated. It will be replaced by a new method that passes the entire `p_L1` payload from the tool into the TEE. All authorization logic will be handled by the code programmed into the TEE itself.
+
+- **New Task [TEE-FROST-01]: Solve Multi-Round Signing for Networkless TEE**
+    - The TEE's "networkless" constraint poses a challenge for multi-round protocols like FROST. A new sub-protocol must be designed. This will likely involve the TEE consuming the `p_L1` payload, performing its validation and round-one calculations, and then outputting a signed `Round1Package`. The `emergency-tool` and `btc-server` would then be responsible for relaying these packages among peers and feeding them back into the TEE for the next round, ultimately collecting the partial signatures for final aggregation.
+
+--- 


docs/design/emergency_wallet_sweep_spec.md
```diff
@@ -0,0 +1,144 @@
...
+### (All Members) Phase 1: Report Local State
+
+- Each federation member operator (including the coordinator) runs a command to broadcast their view of the wallet's state:  
+  `emergency-tool sweep report-state`
+- This command reads the local `btc-server` database to get its list of UTXOs, constructs a signed `ReportState` message, and sends it to the designated Coordinator via the `NewDkgPayload` gRPC endpoint.
+
+### (Coordinator) Phase 2: Build Consensus and Initiate Sweep
+
+- The designated Coordinator operator runs the `initiate` command with a parameter to define consensus:  
+  `emergency-tool sweep initiate --destination <SECURE_ADDR> --fee-rate <RATE> --consensus-threshold 80`
```
DariusParvin: I'm guessing generating the `<SECURE_ADDR>` is outside the scope of this spec, but it seems like that would have to be done as a prerequisite step to this workflow, is that right?

docs/design/emergency_wallet_sweep_spec.md
```diff
@@ -0,0 +1,144 @@
...
+- **Secure Communication**: Operators will use new commands that send signed, encrypted messages through the existing `NewDkgPayload`/`GetDkgPayloads` gRPC endpoints. This provides an auditable, integrated, and robust channel for coordination.
+- **Coordinator-Led Consensus**: To resolve inevitable state discrepancies between members, the Coordinator is responsible for collecting the state of all peers, computing a "safe subset" of UTXOs that have reached consensus, and generating an auditable report of any excluded funds.
+- **Coordinator-Generated PSBT**: The designated Coordinator generates a single, definitive Partially Signed Bitcoin Transaction (PSBT) for the sweep based on the consensus UTXO set. This ensures all members sign the exact same transaction.
+- **Threshold-Based Signing**: The existing FROST signing engine naturally handles offline or lost members, proceeding as long as the minimum signature threshold (t) is met.
+
+---
+
+## 4. The Step-by-Step Emergency Workflow
+
+### (All Members) Phase 1: Report Local State
```
shumkov: Please describe why the coordinator couldn't request state via gRPC?

docs/design/emergency_wallet_sweep_spec.md
```diff
@@ -0,0 +1,144 @@
...
+
+- Each federation member operator (including the coordinator) runs a command to broadcast their view of the wallet's state:  
+  `emergency-tool sweep report-state`
+- This command reads the local `btc-server` database to get its list of UTXOs, constructs a signed `ReportState` message, and sends it to the designated Coordinator via the `NewDkgPayload` gRPC endpoint.
```
shumkov: Is the coordinator saving this in the BTC server? Do we have some kind of TTL for these messages?

docs/design/emergency_wallet_sweep_spec.md
shumkov: Yes, correct.

docs/design/emergency_wallet_sweep_spec.md
shumkov: `NewDkgPayload` accepts: 

```
message DkgPayload {
    bytes sender = 1;
    bytes recipient = 2;
    bytes payload = 3;
}
```

shumkov: What should the `ReportState` message look like?

docs/design/emergency_wallet_sweep_spec.md
```diff
@@ -0,0 +1,144 @@
...
+
+### (Coordinator) Phase 2: Build Consensus and Initiate Sweep
+
+- The designated Coordinator operator runs the `initiate` command with a parameter to define consensus:  
+  `emergency-tool sweep initiate --destination <SECURE_ADDR> --fee-rate <RATE> --consensus-threshold 80`
+- This command performs several critical actions:
+    1. **Collect States**: It calls `GetDkgPayloads` to retrieve all `ReportState` messages from peers.
```
shumkov: Why do we exchange UTXOs twice: phase 1 and phase 2?

docs/design/emergency_wallet_sweep_spec.md
```diff
@@ -0,0 +1,144 @@
...
+    2. **Compute Safe Subset**: It uses the collected UTXO lists as input to a `compute_safe_subset` function, which produces a final set of UTXOs recognized by at least the percentage of members defined by `--consensus-threshold`.
```
shumkov: If we are referring to a function, then we need to describe it.

docs/design/emergency_wallet_sweep_spec.md
```diff
@@ -0,0 +1,144 @@
...
+    3. **Generate Audit Report**: The function generates a detailed `excluded_utxos.json` report listing every UTXO that did not meet the threshold and which members reported it. This file is saved locally for auditing.
+    4. **Generate Definitive PSBT**: It uses only the "safe subset" of consensus UTXOs to generate the single, definitive PSBT for the sweep.
+    5. **Broadcast `StartEmergencySweep`**: It constructs and broadcasts a `StartEmergencySweep` message containing the definitive PSBT and a hash of the audit report.
```
shumkov: We need more details on how the `StartEmergencySweep` broadcasting works

docs/design/emergency_wallet_sweep_spec.md
```diff
@@ -0,0 +1,144 @@
...
+- This command reads the local `btc-server` database to get its list of UTXOs, constructs a signed `ReportState` message, and sends it to the designated Coordinator via the `NewDkgPayload` gRPC endpoint.
+
+### (Coordinator) Phase 2: Build Consensus and Initiate Sweep
+
+- The designated Coordinator operator runs the `initiate` command with a parameter to define consensus:  
+  `emergency-tool sweep initiate --destination <SECURE_ADDR> --fee-rate <RATE> --consensus-threshold 80`
+- This command performs several critical actions:
+    1. **Collect States**: It calls `GetDkgPayloads` to retrieve all `ReportState` messages from peers.
+    2. **Compute Safe Subset**: It uses the collected UTXO lists as input to a `compute_safe_subset` function, which produces a final set of UTXOs recognized by at least the percentage of members defined by `--consensus-threshold`.
+    3. **Generate Audit Report**: The function generates a detailed `excluded_utxos.json` report listing every UTXO that did not meet the threshold and which members reported it. This file is saved locally for auditing.
```
shumkov: Do we do a threshold per input? Or do we do threshold on the entire intersection?

docs/design/emergency_wallet_sweep_spec.md
```diff
@@ -0,0 +1,144 @@
...
+    4. **Generate Definitive PSBT**: It uses only the "safe subset" of consensus UTXOs to generate the single, definitive PSBT for the sweep.
+    5. **Broadcast `StartEmergencySweep`**: It constructs and broadcasts a `StartEmergencySweep` message containing the definitive PSBT and a hash of the audit report.
+
+### (Peers) Phase 3: Acknowledge and Validate
+
+- Peer operators run a command to check for and process incoming coordination messages:  
+  `emergency-tool sweep check-messages`
+- Upon receiving a valid `StartEmergencySweep` message, each Peer's tool:
+    1. Validates the Coordinator's signature.
+    2. Validates that every UTXO used as an input in the definitive PSBT exists in its own local database. If the peer is missing a required input, it cannot sign and will raise a critical error.
```
shumkov: As far as I understand we might some input might be missed that's why we have consensus threshold? If set of participants for each input might be different then there is a high chance that nobody will be able to sign the transaction.

docs/design/emergency_wallet_sweep_spec.md
```diff
@@ -0,0 +1,144 @@
...
+- This command performs several critical actions:
+    1. **Collect States**: It calls `GetDkgPayloads` to retrieve all `ReportState` messages from peers.
+    2. **Compute Safe Subset**: It uses the collected UTXO lists as input to a `compute_safe_subset` function, which produces a final set of UTXOs recognized by at least the percentage of members defined by `--consensus-threshold`.
+    3. **Generate Audit Report**: The function generates a detailed `excluded_utxos.json` report listing every UTXO that did not meet the threshold and which members reported it. This file is saved locally for auditing.
+    4. **Generate Definitive PSBT**: It uses only the "safe subset" of consensus UTXOs to generate the single, definitive PSBT for the sweep.
+    5. **Broadcast `StartEmergencySweep`**: It constructs and broadcasts a `StartEmergencySweep` message containing the definitive PSBT and a hash of the audit report.
+
+### (Peers) Phase 3: Acknowledge and Validate
+
+- Peer operators run a command to check for and process incoming coordination messages:
```
shumkov: Don't we need to validate the destination? Let's accept it as param

docs/design/emergency_wallet_sweep_spec.md
```diff
@@ -0,0 +1,144 @@
...
+- Upon receiving a valid `StartEmergencySweep` message, each Peer's tool:
+    1. Validates the Coordinator's signature.
+    2. Validates that every UTXO used as an input in the definitive PSBT exists in its own local database. If the peer is missing a required input, it cannot sign and will raise a critical error.
+    3. Sends an `AcknowledgeEmergencySweep` message back to the Coordinator via `NewDkgPayload`.
+
+### (Coordinator) Phase 4: Distribute Definitive PSBT and Sign
+
+- The Coordinator's `emergency-tool` collects the `AcknowledgeEmergencySweep` messages. Once a sufficient number of peers have acknowledged, it submits its definitive PSBT to its local `btc-server` to start the FROST signing process.
+
+### (All) Phase 5: Unified Threshold Signing
```
shumkov: What prevents the coordinator from skipping 3rd and 4th phases and start signing directly? If so, why do we need them then? Otherwise, we need to add validation to the signing process.

shumkov (changes requested):

lamafab:
Looks good, just gonna bring up some thoughts/clarifications/recommendations since you're planning on handling the logic over the existing `NewDkgPayload` interface.

* The message type contains a [`bytes` payload field](https://github.com/botanix-labs/Macbeth/blob/3df3764e193bde2f20f87de3d01a1a69e804d9a8/bin/btc-server/proto/btc_server.proto#L208) which is opaque to the btc-server itself; those bytes are [decoded internally by the DKG state machine](https://github.com/botanix-labs/Macbeth/blob/3df3764e193bde2f20f87de3d01a1a69e804d9a8/bin/btc-server/src/dkg/mod.rs#L119). This state machine is generally designed with a single session in mind, and hence I would not expand the state machine itself but rather [just create a new instance](https://github.com/botanix-labs/Macbeth/blob/3df3764e193bde2f20f87de3d01a1a69e804d9a8/bin/btc-server/src/bin/main.rs#L240). So the new message type (enum?) should be in the Protobuf definition that wraps over the exiting `message DkgPayload`, and you can handle the logic for sweeping on a higher level in the btc-server. The Frost message on the Reth side I'd would not touch either. Note that this is technically a breaking change, but since that does not affect consensus this should not be a problem at all as long as it's organized correctly.

* The tricky part will be how to handle the transition; once the Dkg process has completed, you can just discard the state machine, although it's recommended that you still [keep it running for a little while](https://github.com/botanix-labs/Macbeth/blob/3df3764e193bde2f20f87de3d01a1a69e804d9a8/bin/btc-server/src/bin/main.rs#L2070) after completion. However, the DB must be able to save multiple keys in storage and select the correct one when receiving signing requests post-migration. Additionally, while the DKG state machine itself makes sure that only the coordinator can orchestrate the DKG process, you have to make sure that non-coordinators cannot messup this transition mechanism - that you're going to implement - if they try to initiate a new DKG process wrongly. This is all implementation specific, but just something to keep in mind.

* Regarding the audit report, the effective PSBT should be exportable such that it can be opened in a PSBT-compatible wallet, by both the coordinator and the non-coordinators. The audit report should exclusively retrieve information from that PSBT, although I do think having that `excluded_utxos.json` is a good idea.

EDIT:

> lamafab: However, the DB must be able to save multiple keys in storage and select the correct one when receiving signing requests post-migration.

Also, keep in mind that the coordinator must wait for the sweep to actually occur and finalize before it can process pegouts again. So the signing endpoints should probably block/return-early until this has completed

EDIT-2:

Another point regarding the database; you'll have to group UTXOs and transactions (tracking) by multisig wallet... this will be complicated. Maybe you can figure out a better way, such as creating an entirely new database, or propose some acceptable compromises. Since this sweeping event requires manual intervention anyway, you'll have some wiggle room here.

EDIT-3:

Maybe it's easier to just create an entirely new btc-server instance (and corresponding DB), and handle the necessary logic on the Reth side after all(?)

lamafab (approved):

docs/design/emergency_wallet_sweep_spec.md
```diff
@@ -0,0 +1,144 @@
...
+
+- All acknowledged Peers fetch the definitive PSBT from the Coordinator and join the FROST signing session. This follows the existing, unmodified signing flow. As soon as `t` signatures are collected, the transaction is finalized and broadcast.
+
```
shumkov, line 143: We need to output pegouts and pegins made meanwhile with the old key so we can manually handle it

shumkov (changes requested):

docs/design/emergency_wallet_sweep_spec.md
shumkov: After internal discussion: As an output of this command, let's create just one file with PSBT hash, destination address, and excluded inputs, and then the coordinator shares this file though social communication with other operators. The operators are supposed to verify this file manually and then call the "accept sweep request" command with this file as a parameter. The command will validate this file and enable the siging process. During the signing process, we validate PSBT based on the file content. In this case, we don't need to change DKG message payloads.

shumkov (changes requested):

SatwikPrabhu:
Thanks @lamafab and @shumkov for the detailed comments. I'll update the spec accordingly and keep you posted. Had an offline discussion with @shumkov and a chat with @lamafab on what updates are needed.

docs/design/emergency_wallet_sweep_spec.md
```diff
@@ -0,0 +1,144 @@
...
+- **Coordinator-Generated PSBT**: The designated Coordinator generates a single, definitive Partially Signed Bitcoin Transaction (PSBT) for the sweep based on the consensus UTXO set. This ensures all members sign the exact same transaction.
+- **Threshold-Based Signing**: The existing FROST signing engine naturally handles offline or lost members, proceeding as long as the minimum signature threshold (t) is met.
+
+---
+
+## 4. The Step-by-Step Emergency Workflow
+
+### (All Members) Phase 1: Report Local State
+
+- Each federation member operator (including the coordinator) runs a command to broadcast their view of the wallet's state:
```
shumkov: How slow/fast this operation can be ? https://github.com/botanix-labs/Macbeth/pull/852#discussion_r2216893815

shumkov (changes requested):

SatwikPrabhu:
@shumkov @lamafab I have updated the spec based on our discussions and chats. Please let me know if this looks good now and we can get started in the implementation

docs/design/emergency_wallet_sweep_spec.md
```diff
@@ -0,0 +1,699 @@
...
+---
+
+## 2. Core Scenarios & Threat Model
+
+This feature is designed to address two primary emergency scenarios:
+
+- **Scenario A: Active Key Compromise / Imminent Threat**  
+  A situation where one or more of the federation's signing keys are believed to be compromised, requiring an immediate consolidation of funds to prevent theft.
+
+- **Scenario B: Proactive Key Rotation due to Loss of Quorum**
```
shumkov, line 33: What would happen if the coordinator went offline? Is it possible to change this role?

docs/design/emergency_wallet_sweep_spec.md
```diff
@@ -0,0 +1,699 @@
...
+        - Sends `GetMemberUtxoState` requests with configurable timeout (default: 30 seconds per member)
+        - Implements exponential backoff retry logic (3 attempts, 1s/2s/4s delays) for transient failures
+        - Marks persistently unreachable members as offline after retry exhaustion
+    3. **Compute Safe Subset**: Uses the collected UTXO lists to identify UTXOs recognized by at least the percentage of *reachable* members defined by `--consensus-threshold`. Note: UTXOs from offline members are not considered since their state is unknown.
+    4. **Generate Audit Report**: Creates a detailed `excluded_utxos.json` report listing every UTXO that did not meet the consensus threshold and which specific members reported it.
+    5. **Generate Deterministic PSBT**: Creates the definitive PSBT using deterministic construction rules to ensure all members can independently recreate the identical transaction.
+    6. **Generate Sweep Request File**: Constructs a `SweepRequest` JSON file with PSBT hash, destination address, deterministic construction parameters, and consensus metadata for manual operator verification and sharing.
```
shumkov: Let's put excluded utxos there as well?

docs/design/emergency_wallet_sweep_spec.md
```diff
@@ -0,0 +1,699 @@
...
+
+- The designated Coordinator operator runs the `initiate` command with federation configuration:  
+  `emergency-tool sweep initiate --destination <SECURE_ADDR> --fee-rate <RATE> --consensus-threshold 80 --federation-config <PATH> --coordinator-key <PATH>`
+- This command performs several critical actions:
+    1. **Validate Coordinator Authority**: Verifies that the operator holds the coordinator private key for the active key generation by requiring a cryptographic signature of the emergency session nonce.
+    2. **Query Reachable Members**: Directly queries each federation member's `btc-server` via the `GetMemberUtxoState` gRPC endpoint to collect current UTXO state:
+        - Reads the `FederationTomlConfig` from the specified path
+        - Extracts member socket addresses from `federation_member_public_key[].socket_addr` fields
+        - Establishes authenticated gRPC connections using JWT authentication
+        - Sends `GetMemberUtxoState` requests with configurable timeout (default: 30 seconds per member)
```
shumkov, line 62: How long will it take for peers to collect the state and send it back? Do you think 30 seconds max is enough?

docs/design/emergency_wallet_sweep_spec.md
```diff
@@ -0,0 +1,699 @@
...
+---
+
+## 4. The Step-by-Step Emergency Workflow
+
+### (Coordinator) Phase 1: Collect States and Initiate Sweep
+
+- The designated Coordinator operator runs the `initiate` command with federation configuration:  
+  `emergency-tool sweep initiate --destination <SECURE_ADDR> --fee-rate <RATE> --consensus-threshold 80 --federation-config <PATH> --coordinator-key <PATH>`
+- This command performs several critical actions:
+    1. **Validate Coordinator Authority**: Verifies that the operator holds the coordinator private key for the active key generation by requiring a cryptographic signature of the emergency session nonce.
```
shumkov: Is emergency session nonce stored somewhere? Or is it just a socially communicated number and we pass it through a param? Why do we actually need this?

docs/design/emergency_wallet_sweep_spec.md
```diff
@@ -0,0 +1,699 @@
...
+    3. **Compute Safe Subset**: Uses the collected UTXO lists to identify UTXOs recognized by at least the percentage of *reachable* members defined by `--consensus-threshold`. Note: UTXOs from offline members are not considered since their state is unknown.
+    4. **Generate Audit Report**: Creates a detailed `excluded_utxos.json` report listing every UTXO that did not meet the consensus threshold and which specific members reported it.
+    5. **Generate Deterministic PSBT**: Creates the definitive PSBT using deterministic construction rules to ensure all members can independently recreate the identical transaction.
+    6. **Generate Sweep Request File**: Constructs a `SweepRequest` JSON file with PSBT hash, destination address, deterministic construction parameters, and consensus metadata for manual operator verification and sharing.
+
+### (Peers) Phase 2: Manual Verification and Acceptance
+
+- Peer operators receive the `SweepRequest` JSON file from the coordinator via external communication channels
+- **Manual Verification**: Each operator manually reviews:
+    1. **Coordinator Authority**: Verifies the coordinator signature and session nonce
```
shumkov: I feel it will be easier to verify with the accept-request command rather than manually.

docs/design/emergency_wallet_sweep_spec.md
```diff
@@ -0,0 +1,699 @@
...
+    4. **Context Validation**: Confirms this matches the emergency scenario being addressed
+- **Accept Sweep Request**: After manual verification, operators run:  
+  `emergency-tool sweep accept-request <SWEEP_REQUEST_FILE>`
+- Upon processing the `SweepRequest` file, each member's tool:
+    1. **Validate Authority**: Verifies coordinator signature and emergency session authorization
+    2. **Reconstruct Local PSBT**: 
+        - Queries local `btc-server` database for all available UTXOs
+        - Applies the same consensus filtering using the provided parameters
+        - Constructs PSBT using deterministic emergency PSBT builder with identical rules
+    3. **Structural Validation**: Compares the locally reconstructed PSBT with the coordinator's PSBT using semantic equivalence (same inputs, outputs, fees) rather than byte-for-byte comparison
```
shumkov: Why semantic is better than byte-for-byte which simpler?

docs/design/emergency_wallet_sweep_spec.md
```diff
@@ -0,0 +1,699 @@
...
+    4. **Enable Signing**: If validation passes, enables the signing process for this emergency session
+    5. **Validation Failure Handling**: If PSBTs don't match, logs detailed discrepancies and refuses to participate
+
+### (Coordinator) Phase 3: Initiate Signing
+
+- The Coordinator operator accepts their own sweep request to enable local signing participation
+- Once sufficient members have accepted the sweep request, the coordinator can initiate the FROST signing process
```
shumkov: How does the coordinator know that they accepted? I think it should just start singing on the `initiate` command and receive rejects ot accepts. When we have enough peers we will get it signed.

docs/design/emergency_wallet_sweep_spec.md
```diff
@@ -0,0 +1,699 @@
...
+
+### (All) Phase 4: Unified Threshold Signing
+
+- All members who have accepted the sweep request participate in the FROST signing session using the standard signing flow with emergency session validation. As soon as `t` signatures are collected, the transaction is finalized and broadcast.
```
shumkov: During this unified singing, we need to make sure that we are singing what we are accepted.

docs/design/emergency_wallet_sweep_spec.md
```diff
@@ -0,0 +1,699 @@
...
+---
+
+## 7. Security & Operational Considerations
+
+### **Security Measures**
+- **Coordinator Authority**: Cryptographic validation prevents unauthorized emergency initiation
+- **Session Isolation**: Emergency operations are completely isolated from normal pegout processing
+- **Message Integrity**: All coordination messages include cryptographic signatures and checksums
+- **Audit Trail**: Comprehensive logging of all consensus decisions and UTXO exclusions
+- **Replay Protection**: Session nonces prevent replay attacks on emergency endpoints
```
shumkov: Are we talking about collecting the state endpoint? Don't we trust peers since we are in a permissioned network?

shumkov (changes requested):

docs/design/emergency_wallet_sweep_spec.md
SatwikPrabhu: For now, the current effort is mainly for a happy scenario where things like this are not considered. Obviously we need to take care of these scenarios but not in the scope here

docs/design/emergency_wallet_sweep_spec.md
SatwikPrabhu: Yep. Forgot to add that. Thanks for reminding me.

docs/design/emergency_wallet_sweep_spec.md
SatwikPrabhu: Actually this is a placeholder from my side. Its open for discussion

docs/design/emergency_wallet_sweep_spec.md
SatwikPrabhu: Tbh I dont think we really need this and would probably be overengineering as emergency scenarios would be rare and this design is decently secure already

docs/design/emergency_wallet_sweep_spec.md
SatwikPrabhu: Makes sense. Am updating this. Thanks!

docs/design/emergency_wallet_sweep_spec.md
SatwikPrabhu: I guess both work similarly. But yeah, I get your point. byte to byte is more simpler

docs/design/emergency_wallet_sweep_spec.md
SatwikPrabhu: As per the previous comment https://github.com/botanix-labs/Macbeth/pull/876#discussion_r2244172913, nonce will be removed

docs/design/emergency_wallet_sweep_spec.md
```diff
@@ -0,0 +1,764 @@
...
+    2. **Query Reachable Members**: Directly queries each federation member's `btc-server` via the `GetMemberUtxoState` gRPC endpoint to collect current UTXO state:
+        - Reads the `FederationTomlConfig` from the specified path
+        - Extracts member socket addresses from `federation_member_public_key[].socket_addr` fields
+        - Establishes authenticated gRPC connections using JWT authentication
+        - Sends `GetMemberUtxoState` requests with configurable timeout (default: 30 seconds per member)
+        - Implements exponential backoff retry logic (3 attempts, 1s/2s/4s delays) for transient failures
+        - Marks persistently unreachable members as offline after retry exhaustion
+    3. **Compute Safe Subset**: Uses the collected UTXO lists to identify UTXOs recognized by at least the percentage of *reachable* members defined by `--consensus-threshold`. Note: UTXOs from offline members are not considered since their state is unknown.
+    4. **Generate Audit Report**: Creates a detailed `excluded_utxos.json` report listing every UTXO that did not meet the consensus threshold and which specific members reported it.
```
shumkov, line 66: Remove this, we already have it in SweepRequest

shumkov (approved):

PR #880, opened 2025-07-28
feat(pegouts): remove finalized pegout id from pending pegouts

scottmillner:

ticket: https://github.com/botanix-labs/botanix/issues/1067

When signing a psbt, if a finalized pegout id is included, remove it from the pending pegouts list before returning the error.  Otherwise, the coordinator will keep adding that finalized pegout id in the psbt.

## What was done?

- Remove finalized pegouts ids from the pending pegouts list.

## How Has This Been Tested?

- Refactored existing test to ensure finalized pegout id is removed from the pending pegouts list.

## Breaking Changes

- None

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Pending pegouts that are already finalized are now properly removed from the pending list when detected during validation.

* **Tests**
  * Enhanced tests to verify that finalized pegouts are removed from the pending list after validation.


DariusParvin (approved):

DariusParvin:
Looks good!  I am curious though, how does a pegout end up getting in a state where it's finalized but still pending?

SatwikPrabhu (approved):

SatwikPrabhu:
> Looks good! I am curious though, how does a pegout end up getting in a state where it's finalized but still pending?

Maybe when` add_tx_back_to_pending()` is called in `track_mempool` when it is detected to be not in the mempool?

lamafab (approved):

scottmillner:
> Looks good! I am curious though, how does a pegout end up getting in a state where it's finalized but still pending?

great question and this an unlikely edge case(s)

an example where this could happen:
- our staged headers logic syncs any missed pegouts while btc-server is down
- if we introduced a bug to staged headers, we could accidentally send already handled pegouts to btc-server
- btc-server would then store them in the pending pegouts list which would be included in the psbt in the next epoch block

and as @SatwikPrabhu mentioned, if there was faulty logic in how we track txs, we could have the same situation

this is very unlikely but if it happened, and we never removed the finalized pegout id from pending pegouts, signers would always reject the psbt and pegouts would halt

PR #886, opened 2025-08-01
feat(btc-server): RecoverMissingUtxos grpc endpoint

DariusParvin:
Addresses https://github.com/botanix-labs/botanix/issues/1089
We need a safe way to recover Utxos that are incorrectly removed from the db.

## What was done?
Added a new grpc endpoint on btc-server called `RecoverMissingUtxos`.
The plan is for this endpoint to be called by the reth server (TBD), which will read the data from a file.

The endpoint takes in an outpoint and and eth address (optional). Before adding the utxo to the db it will verify the following:
- the utxo is not already in the db
- the utxo exists on chain
- the scriptPubKey is as expected (matches the generated gateway address or change address, depending on if `eth_address` is provided)

## How Has This Been Tested?

- Integration tests covering the new method happy path and failure cases.

## Breaking Changes

- None

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Introduced a new gRPC API method to recover missing UTXOs by specifying them individually.
  * Added support for submitting UTXOs with optional Ethereum addresses for recovery.
  * The recovery response now reports the total number of UTXOs requested and successfully recovered.

* **Tests**
  * Added comprehensive tests to verify correct recovery behavior for various UTXO scenarios.


bin/btc-server/proto/btc_server.proto
```diff
@@ -144,6 +155,12 @@ message Utxo {
     string eth_address = 3;
 }
 
+message UtxoToRecover {
+    OutPoint outpoint = 1;
+    // empty eth_address for change utxo
+    string eth_address = 2;
+}
```
DariusParvin, line 162: @rwlockbg i think we had talked about using `Vec<u8>` but when it came to writing it, i felt like we might as well use the proper types. It shouldnt be much for the reth side to parse a json file into these types before sending request.

bin/btc-server/src/bin/main.rs
```diff
@@ -2106,6 +2110,143 @@ where
...
+            let Some(on_chain_utxo) = on_chain_utxo else {
+                warn!(
+                    "BtcServer::recover_missing_utxos: UTXO {} does not exist on chain, skipping.",
+                    outpoint
+                );
+                continue;
+            };
+
+            // parse the eth address if it is not empty
+            let eth_address: Option<[u8; 20]>;
```
scottmillner: could also write it this way to initialize it:
```
let eth_address: Option<[u8; 20]> = if req_utxo.eth_address.is_empty() {
                None
            } else {
                let parsed_eth_address = parse_eth_address(req_utxo.eth_address).map_err(|e| {
                    error!("Invalid ETH address format: {}", e);
                    tonic::Status::internal(format!("Invalid ETH address format: {}", e))
                })?;
                Some(parsed_eth_address)
            };
 ```

File: bin/btc-server/proto/btc_server.proto [hunk 3]
```diff
@@ -53,6 +53,8 @@ service BtcServer {
 
     rpc ResetAllUtxos(ResetAllUtxosRequest) returns (Empty) {}
 
+    rpc RecoverMissingUtxos(RecoverMissingUtxosRequest) returns (RecoverMissingUtxosResponse) {}
```
scottmillner, line 56: shouldn't we add this to `trait BtcServerExtendedApi` in `extended_client.rs` as well?

rwlockbg:
@DariusParvin There is a missing .bin file I believe in the PR which is also getting created when you change the defs of the .proto file. Make sure you also include it in the PR.

bin/btc-server/src/bin/main.rs
DariusParvin: nice, I've updated that

bin/btc-server/proto/btc_server.proto
DariusParvin: ah yes thanks for catching that, added now

DariusParvin:
@rwlockbg ah thanks for noticing, i didnt realize we should be committing the .bin file. I have included it now

bin/btc-server/src/bin/main.rs
```diff
@@ -2106,6 +2110,142 @@ where
...
+        for req_utxo in request.into_inner().utxos {
+            info!("Processing UTXO: {:?}", req_utxo.outpoint);
+
+            // convert the request outpoint to the database outpoint
+            let outpoint = req_utxo
+                .outpoint
+                .as_ref()
+                .map(|outpoint| bitcoin::OutPoint {
+                    txid: bitcoin::Txid::from_slice(&outpoint.txid).unwrap(),
+                    vout: outpoint.vout,
```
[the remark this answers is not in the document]
DariusParvin: removed .unwrap call

bin/btc-server/src/bin/main.rs
```diff
@@ -2106,6 +2110,142 @@ where
...
+
+            // add the utxo to the list of utxos to be added
+            info!("UTXO {} passed all validations, adding to recovery list", outpoint);
+            utxos_to_add.push(utxo);
+        }
+
+        // add the utxos to the database
+        let utxo_refs: Vec<&crate::database::Utxo> = utxos_to_add.iter().collect();
+        info!("Storing {} missing UTXOs.", utxo_refs.len());
+        self.db.store_utxos(&utxo_refs).to_status()?;
```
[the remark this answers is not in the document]
DariusParvin: added

scottmillner (dismissed):

bin/btc-server/src/bin/main.rs
```diff
@@ -2106,6 +2110,142 @@ where
     ) -> Result<tonic::Response<rpc::FinalizeSigningResponse>, tonic::Status> {
         panic!("Not used");
     }
+
+    async fn recover_missing_utxos(
+        &self,
+        request: tonic::Request<RecoverMissingUtxosRequest>,
+    ) -> Result<tonic::Response<RecoverMissingUtxosResponse>, tonic::Status> {
+        let total_requested = request.get_ref().utxos.len() as u64;
```
shumkov, line 2120: Shouldn't we validate jwt (`self.validate_jwt(&req)?`) as we do in other methods?

bin/btc-server/src/bin/main.rs
```diff
@@ -2106,6 +2110,142 @@ where
...
+
+        // Ensure we have a key package
+        let key_package = self
+            .db
+            .get_key_package()
+            .to_status()?
+            .ok_or(tonic::Status::internal("Missing key package"))?;
+
+        let mut utxos_to_add = Vec::new();
+        for req_utxo in request.into_inner().utxos {
```
shumkov, line 2141: How long do you think this could work? What are our default timeouts for gRPC calls?

bin/btc-server/src/bin/main.rs
```diff
@@ -2106,6 +2110,142 @@ where
     ) -> Result<tonic::Response<rpc::FinalizeSigningResponse>, tonic::Status> {
         panic!("Not used");
     }
+
+    async fn recover_missing_utxos(
```
shumkov, line 2114: The logic in reth that will call this method is matter of another PR?

DariusParvin: yeah exactly

bin/btc-server/src/bin/main.rs
DariusParvin: yes good catch, thank you! I just added it

bin/btc-server/src/bin/main.rs
DariusParvin: Yeah good point I'll see what timeouts we have if any for  gRPC calls. I expect this method to take  about a second for out ~300 utxos.

DariusParvin: in btc-server config.toml i see a timeout of 180 seconds [here](https://github.com/botanix-labs/Macbeth/blob/d6afae5eeb88e2bb0020de23b3b52c5d0eb8954b/bin/btc-server/config.toml#L14). This endpoint will be way faster than that. The only call to bitcoind is to get the utxo, but that is a fast in-memory lookup.

scottmillner (approved):

lamafab (approved):
Nice!

PR #891, opened 2025-08-04
fix(fee): fn effective tip per gas returns the final fee instead of the priority fee

scottmillner:

The base fee is added back to the tx cost and not in `fn effective_tip_per_gas()`.  This way, the gas oracle returns an accurate final fee.  Without this change, the base fee is returned from `fn effective_tip_per_gas()` which is incorrect.

## What was done?

- The base fee is added directly to the tx cost instead of in `fn effective_tip_per_gas()`.

## How Has This Been Tested?

- Existing unit and int tests passed.
- I've manually tested the following:
  - syncing from genesis
  - simple transfers for legacy and eip-1559 txs
  - pegins/pegouts
  - rpc endpoint `eth_gasPrice` returns the correct gas price (doesn't add base fee)

## Breaking Changes

- None

## Checklist

### DIFF
--- a/crates/ethereum/evm/src/execute.rs
+++ b/crates/ethereum/evm/src/execute.rs
@@ -299,8 +299,10 @@ where
             })?;
 
             // calculate the total transaction fee
-            let transaction_fee =
-                transaction.clone().effective_tip_per_gas(base_fee).expect("base fee is valid");
+            let mut transaction_fee =
+                transaction.clone().effective_tip_per_gas(base_fee).expect("base fee exists");
+            // Include the base fee so it's not burned
+            transaction_fee += base_fee.expect("base fee exists") as u128;
             total_block_fees += transaction_fee * u128::from(result.gas_used());
 
             // append gas used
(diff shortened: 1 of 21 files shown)


crates/ethereum/evm/src/execute.rs
```diff
@@ -299,8 +299,10 @@ where
             })?;
 
             // calculate the total transaction fee
-            let transaction_fee =
-                transaction.clone().effective_tip_per_gas(base_fee).expect("base fee is valid");
+            let mut transaction_fee =
+                transaction.clone().effective_tip_per_gas(base_fee).expect("base fee exists");
+            // Include the base fee so it's not burned
+            transaction_fee += base_fee.expect("base fee exists") as u128;
```
lamafab: I'm not sure if panicking here is safe, I could image some issues when considering historical sync or some other edge case. Maybe it's cleaner to just do this:

```rust
// calculate the total transaction fee
let mut transaction_fee =
    transaction.clone().effective_tip_per_gas(base_fee).expect("base fee is valid");

// prevent burn of base fee if provided.
if let Some(base_fee) = base_fee {
    transaction_fee += base_fee as u128;
}

total_block_fees += transaction_fee * u128::from(result.gas_used());
```

lamafab: Deprecated by: https://github.com/botanix-labs/Macbeth/pull/891/commits/7ff4f3f7e854b0ad082ecb68d7f2d81f174304f0

lamafab (approved):
Looks good, but lets get those sanity checks done live on testnet ^_^

PR #911, opened 2025-08-08
feat(reth): add missing utxos on reth startup

DariusParvin:
Part 2 of implementing UTXO recovery functionality. 
During reth federation node startup call the new `recover_missing_utxos` gRPC endpoint on the btc-server.

https://github.com/botanix-labs/botanix/issues/1089

## What was done?

- on fed node startup, just after the `btc-server` health check, a utxo_recovery.json file is read, and if it exists, the node will call `btc-server.recover_missing_utxos` with the utxo data. If an error is thrown while reading the file, or if it doesn't exist, it'll log the error and move on. 

- Added a new optional env argument `RETH_UTXO_RECOVERY_FILE` that should point to the path/filename for the utxo recovery json file. If the argument is provided, it will attempt to load it, if not, it will skip that step.

## How Has This Been Tested?

- Additional unit tests. 
- Manually tested that the file loads and btc-server is called by using the frost_e2e integration test

## Breaking Changes

- None

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Recover missing Bitcoin UTXOs during POA node startup by loading recovery entries from a user-provided JSON file (CLI/env configurable).
  * Startup recovery run emits logs with outcome details (requested vs. recovered counts).

* **Bug Fixes**
  * Invalid or malformed recovery entries are skipped safely with informative log messages.


scottmillner (dismissed):

bin/reth/src/commands/poa/mod.rs
```diff
@@ -103,6 +105,57 @@ use tokio::{
 };
 use tracing::{debug, error, info};
 
+use reth_fs_util;
```
shumkov: Let's move this logic to a separate module? It could be even bin/btc-server since all UTXO stuff is there.

DariusParvin: Hmm yeah previously in this issue https://github.com/botanix-labs/botanix/issues/1089 we decided on call the btc-server endpoint from the reth server for security reasons, but if we are loading the data in from a json file in the reth server, i guess we could just do that in the btc-server directly? It makes sense to me but I'd want a second opinion from @rwlockbg or @scottmillner in case there's some context i'm missing.

DariusParvin: ah okay i think i get what you mean, just moving this logic there, then we can import it here, instead of having it defined here 👍 i'll do that

bin/reth/src/commands/poa/mod.rs
```diff
@@ -103,6 +105,57 @@ use tokio::{
...
+    eth_address: String,
+}
+
+impl From<UtxoRecoveryData> for UtxoToRecover {
+    fn from(data: UtxoRecoveryData) -> Self {
+        UtxoToRecover {
+            outpoint: Some(OutPoint {
+                txid: hex::decode(&data.txid).unwrap_or_else(|_| {
+                    error!(target: "reth::cli", "Invalid txid hex: {}", data.txid);
+                    vec![0u8; 32] // Fallback to zeros
```
lamafab: Is that a good decision? Consider implementing [`TryFrom`](https://doc.rust-lang.org/std/convert/trait.TryFrom.html) instead.

bin/reth/src/commands/poa/mod.rs
```diff
@@ -398,6 +451,28 @@ impl<Ext: clap::Args + fmt::Debug> PoaNodeCommand<Ext> {
             })?;
             info!(target: "reth::cli", "Btc server authenticated");
 
+            info!(target: "reth::cli", "Recovering missing UTXOs...");
+
+            let utxos = read_utxos_from_file(Path::new("utxo_recovery.json"));
```
lamafab: So what's the idea here? We have the option to add this file to recover any Utxos? Do note that this **always** returns an error if the file does not exist - it should probably be optional?

bin/reth/src/commands/poa/mod.rs
scottmillner: yeah just moving that logic

scottmillner (dismissed):

bin/reth/src/commands/poa/mod.rs
DariusParvin: yeah it isnt ideal to return an error as normal behavior. I'll add an optional argument to indicate if it should be loaded

DariusParvin: added

bin/reth/src/commands/poa/mod.rs
DariusParvin: yeah you're right, better to throw an error than rely on btc-server rejecting the non existent utxo. I've had a go implementing `TryFrom` if you are able to take a look

scottmillner (approved):

lamafab (approved):

PR #922, opened 2025-08-12
fix(btc-server): ensure metrics server shuts down when btc-server shu…

scottmillner:
Sometimes when the btc-server process terminates the metrics server is still running giving the false impression that the btc-server is live and healthy.
This ensures the metrics server terminates when btc-server does.
I also added a btc-server hearbeat which is part of the `is_healthy` check.

Issue fixed: https://github.com/botanix-labs/botanix/issues/1095

The metrics server terminates when btc-server does.

## What was done?

- Added logic to terminate both servers when there's a panic or btc-server crashes.

## How Has This Been Tested?

- Current unit and int tests pass

## Breaking Changes

- None

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added an optional metrics HTTP server that starts when telemetry is enabled, serves on the configured port, and shuts down gracefully.

* **Refactor**
  * Centralized shutdown handling across components for consistent, orderly termination.
  * Graceful shutdown now triggered after termination requests and on unexpected errors, with clearer final logs.
  * Coordinated server lifecycle to ensure clean stops and reduced partial shutdown scenarios.


bin/btc-server/src/bin/main.rs
```diff
@@ -2299,6 +2299,20 @@ async fn main() -> anyhow::Result<(), Box<dyn std::error::Error>> {
 
     let config = btcserverlib::config::load_config()?;
 
+    // Create a shared shutdown signal that all components will monitor
+    let (shutdown_tx, _shutdown_rx) = broadcast::channel::<()>(1);
+
+    // Set up panic hook to ensure shutdown signal is sent even on panic
+    let shutdown_tx_for_panic = shutdown_tx.clone();
+    let original_hook = std::panic::take_hook();
+    std::panic::set_hook(Box::new(move |info| {
```
rwlockbg, line 2308: @scottmillner We already have `pub fn set_panic_hook()`,  under `panic_hook.rs`. You dont need any of this I would say.

rwlockbg (dismissed):

bin/btc-server/src/bin/main.rs
scottmillner: thanks for pointing that out!
i'd still need to refactor that method to take an optional Sender to send the message on the channel the metrics server is subscribed to so it's guaranteed the metrics server always terminiates when there's a panic which isn't currently happening

DariusParvin (dismissed):

PR #928, opened 2025-08-15
feat: create sweep psbt

SatwikPrabhu:

Feature implemented: Emergency wallet sweep PSBT creation functionality
This commit implements the ability to create Partially Signed Bitcoin Transactions (PSBTs) for emergency wallet sweeps in the Botanix federation.

## What was done?

- Added async PSBT creation function: Implemented create_psbt_async in the botanix-wallet-sweep crate that:
     - Fetches all UTXOs and tracked transactions from the BTC server
     - Filters out tracked/spent inputs to avoid conflicts
     - Creates a PSBT that sweeps all available funds to a specified destination
     
- Enhanced CLI command: Updated the Psbt subcommand in the sweep CLI to:
     - Use the new async PSBT creation function
     - Output the PSBT in base64 format to stdout for use with other tools
     - Automatically save the PSBT to a file with .psbt extension
     - Include overwrite protection with warning messages
   
- Improved error handling: Added proper error wrapping and more descriptive error messages for PSBT creation failures

## How Has This Been Tested?

- Have added comprehensive unit tests around the psbt creation logic

## Breaking Changes

-

## Checklist

## Summary by CodeRabbit

- New Features
  - Wallet sweep now produces a PSBT, prints it in base64, and saves a .psbt file next to the request (with overwrite warning).
  - Emergency sweep supports building PSBTs from available UTXOs with validation, fee/size limits, and embedded metadata; works online (async) and from provided data.
  - Added session tracking and configurable consensus parameters for sweeps.

- Improvements
  - Clearer logging and error handling during sweep and PSBT generation.

- Chores
  - Dependency adjustments to enable sweep and PSBT functionality.

### DIFF
--- a/bin/reth/src/commands/sweep.rs
+++ b/bin/reth/src/commands/sweep.rs
@@ -1,15 +1,14 @@
 //! Emergency wallet sweep command implementation
 
 use bitcoin::{address::NetworkUnchecked, FeeRate};
-use botanix_data_parser::{DataParser, SerializationType, DEFAULT_COMPRESSION_STRATEGY};
-use botanix_wallet_sweep::{request::DestinationConfig, WalletSweepRequest};
+use botanix_wallet_sweep::{create_psbt_async, request::DestinationConfig, WalletSweepRequest};
 use btc_server_client::{jwt::JwtSecret, BtcServerExtendedApi, Empty, GrpcClientFactory};
-use btcserverlib::{database::Db, dkg, federation_args::FederationTomlConfig};
+use btcserverlib::{dkg, federation_args::FederationTomlConfig};
 use clap::{Parser, Subcommand};
-use eyre::{OptionExt, WrapErr};
+use eyre::WrapErr;
 use reth_cli_runner::CliContext;
 use std::{fs, net::SocketAddr, path::PathBuf, str::FromStr};
-use tracing::{error, info, warn};
+use tracing::{info, warn};
 
 /// Emergency wallet sweep operations for the Botanix federation
 #[derive(Debug, Parser)]
@@ -72,7 +71,7 @@ pub enum SweepSubcommands {
 }
 
 #[derive(Debug, Parser)]
-struct DestinationOptions {
+pub struct DestinationOptions {
     /// Bitcoin network to use (mainnet, testnet, regtest)
     #[arg(long)]
     network: bitcoin::Network,
@@ -185,11 +184,26 @@ impl SweepCommand {
             SweepSubcommands::Psbt { request_file_path } => {
                 let request = WalletSweepRequest::from_json_file(request_file_path).await?;
 
-                let psbt = botanix_wallet_sweep::create_psbt(request).wrap_err_with(|| {
-                    format!("Failed to create PSBT from request file {:?}", request_file_path)
-                })?;
+                let psbt = create_psbt_async(request, &mut btc_server_client).await.wrap_err_with(
+                    || format!("Failed to create PSBT from request file {:?}", request_file_path),
+                )?;
 
-                // TODO: Save PSBT to file or write to std out if pipe is provided
+                // Print PSBT in base64 format for use with other tools
+                use bitcoin::base64::{engine::general_purpose, Engine as _};
+                let psbt_base64 = general_purpose::STANDARD.encode(&psbt.serialize());
+                println!("Emergency sweep PSBT:");
+                println!("{}", psbt_base64);
+
+                // Save to file with overwrite protection
+                let psbt_filename = request_file_path.with_extension("psbt");
+                if psbt_filename.exists() {
+                    warn!("PSBT file {:?} already exists, overwriting...", psbt_filename);
+                }
+
+                fs::write(&psbt_filename, psbt_base64.as_bytes()).wrap_err_with(|| {
+                    format!("Failed to write PSBT to file {:?}", psbt_filename)
+                })?;
+                info!("PSBT saved to: {:?}", psbt_filename);
             }
         }
 
--- a/crates/botanix-wallet-sweep/Cargo.toml
+++ b/crates/botanix-wallet-sweep/Cargo.toml
@@ -21,7 +21,7 @@ serde = "1.0.171"
 thiserror = "1.0.43"
 tokio = { version = "1.45.0", features = ["full"] }
 tracing = "0.1.40"
-#btc-server = { path = "../../bin/btc-server" }
+# btc-server = { path = "../../bin/btc-server" } # Circular dependency - btc-server depends on us
 btc-server-client = { path = "../../bin/btc-server/client" }
 eyre = { workspace = true }
 botanix-data-parser = { workspace = true }
@@ -38,3 +38,4 @@ futures = "0.3.31"
 
 [dev-dependencies]
 mockall = { workspace = true }
+# rand is already available from regular dependencies (version 0.6)
--- a/crates/botanix-wallet-sweep/src/psbt.rs
+++ b/crates/botanix-wallet-sweep/src/psbt.rs
@@ -1,6 +1,1716 @@
-use crate::request::WalletSweepRequest;
-use bitcoin::Psbt;
+//! Emergency wallet sweep PSBT creation
+//!
+//! This module provides functionality to create PSBTs (Partially Signed Bitcoin Transactions)
+//! for emergency wallet sweeps in the Botanix federation. Emergency sweeps collect all available
+//! UTXOs and create a single transaction to move funds to a secure destination.
+//!
+//! The module supports both synchronous and asynchronous PSBT creation, with automatic
+//! filtering of tracked UTXOs, proper fee calculation, and transaction size limits to ensure
+//! reliable Bitcoin network propagation.
+//!
+//! ## Fee Calculation Compatibility
+//!
+//! The fee calculations in this module are IDENTICAL to those used in btc-server's coin selection
+//! (bin/btc-server/src/wallet/util.rs and bin/btc-server/src/wallet/coin_selection.rs).
+//! Both systems use the same weight constants and calculation methods to ensure perfect
+//! consistency across the federation.
 
-pub fn create_psbt(_request: WalletSweepRequest) -> eyre::Result<Psbt> {
-    todo!("Implement PSBT creation logic based on the WalletSweepRequest");
+use crate::request::WalletSweepRequest;
+use bitcoin::hashes::Hash;
+use bitcoin::psbt::PsbtSighashType;
+use bitcoin::sighash::TapSighashType;
+use bitcoin::{Amount, FeeRate, OutPoint, Psbt, ScriptBuf, TxOut, Weight};
+use btc_server_client::{BtcServerExtendedApi, BtcServerExtendedClient, Empty};
+use std::collections::HashSet;
+use thiserror::Error;
+
+// Transaction size limits for reliable Bitcoin network propagation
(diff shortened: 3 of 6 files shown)


bin/reth/src/commands/sweep.rs
```diff
@@ -185,11 +184,25 @@ impl SweepCommand {
...
+                println!("{}", psbt_base64);
+                
+                // Save to file with overwrite protection
+                let psbt_filename = request_file_path.with_extension("psbt");
+                if psbt_filename.exists() {
+                    warn!("PSBT file {:?} already exists, overwriting...", psbt_filename);
+                }
+                
+                fs::write(&psbt_filename, psbt_base64.as_bytes())
+                    .wrap_err_with(|| format!("Failed to write PSBT to file {:?}", psbt_filename))?;
```
DariusParvin: What is the intended use for this base64 serialized psbt? I ask because the other nodes will also need the UTXO information, specifically the associated etherum addresses, to be able to sign the inputs. That's part of the `Psbt` but wont be included in the serialized psbt.

Also, i know base64 is more compact, but it might be easier to deal with hex encoded strings? in terms of making it easier to use with other tools like bitcoind or transaction parsers in general. No strong feelings for me either way.

crates/botanix-wallet-sweep/src/psbt.rs
```diff
@@ -1,6 +1,1330 @@
...
+    // Validate request parameters
+    validate_request_parameters(&request)?;
+    let destination_network = request.destination_network.parse::<bitcoin::Network>()
+        .map_err(|e| SweepError::NetworkParsing(e.to_string()))?;
+    let destination_address = request.destination_address
+        .require_network(destination_network)
+        .map_err(|e| SweepError::AddressValidation(e.to_string()))?;
+    
+    // Network validation (TODO: integrate with node configuration)
+    validate_network_consistency(destination_network)?;
```
DariusParvin: I dont think we gain anything from having a `destination_network` in the sweep request. The bitcoin address already encodes whether it is testnet or mainnet. So if anything I think there's less room for error by just checking that. That way we dont have a situation where the e.g. request destination_network is 'testnet' but the address is mainnet.

crates/botanix-wallet-sweep/src/psbt.rs
```diff
@@ -1,6 +1,1330 @@
...
+    
+    let fee_rate = FeeRate::from_sat_per_vb(request.fee_rate_sat_vb)
+        .ok_or_else(|| SweepError::InvalidFeeRate { 
+            fee_rate: request.fee_rate_sat_vb, 
+            min: MIN_FEE_RATE_SAT_VB, 
+            max: MAX_FEE_RATE_SAT_VB 
+        })?;
+    
+    // Filter and prepare UTXOs for spending
+    let spendable_utxos = filter_and_sort_utxos(available_utxos, &tracked_inputs)?;
```
DariusParvin, line 242: I can understand filtering out tracked inputs, but im not sure I see the benefit of sorting them

DariusParvin: ah okay, this is for the case where the number of inputs is larger than `MAX_EMERGENCY_SWEEP_INPUTS`, got it

crates/botanix-wallet-sweep/src/psbt.rs
```diff
@@ -1,6 +1,1330 @@
...
+/// Calculates the total value of UTXOs
+fn calculate_total_value(utxos: &[btc_server_client::Utxo]) -> Amount {
+    utxos
+        .iter()
+        .map(|utxo| {
+            utxo.output.as_ref()
+                .map(|output| Amount::from_sat(output.value))
+                .unwrap_or(Amount::ZERO)
+        })
+        .sum()
```
DariusParvin: Maybe worth having a checked sum? Just because these utxo values come from the db, and who knows if for some reason there's incorrect data in there.

crates/botanix-wallet-sweep/src/psbt.rs
```diff
@@ -1,6 +1,1330 @@
...
+    
+    let limited_utxos = apply_size_limits(spendable_utxos)?;
+    let total_input_value = calculate_total_value(&limited_utxos);
+    
+    if total_input_value == Amount::ZERO {
+        return Err(SweepError::InsufficientFunds { 
+            fee: Amount::ZERO, 
+            total_value: total_input_value 
+        });
+    }
```
DariusParvin: This minimum input value should be enough to cover the fees. I think a realistic scenario we hit this is if we have to split the sweep over multiple transactions, then the last sweep transaction could be spending low value - near dust level outputs. Combined with a high fee rate, it's possible that we reach this insufficient funds scenario. So yeah with that in mind i think it makes sense to calculate the absolute fee, then use that value for this check

crates/botanix-wallet-sweep/src/psbt.rs
```diff
@@ -1,6 +1,1330 @@
...
+}
+
+/// Internal representation of a UTXO input for PSBT creation
+#[derive(Debug, Clone)]
+struct SweepInput {
+    outpoint: OutPoint,
+    output: TxOut,
+    eth_address: Option<[u8; 20]>,
+    version: UtxoVersion,
+}
```
DariusParvin, line 152: I think it makes sense to use the existing `InputDTO` struct:
```
pub(crate) struct InputDTO {
    pub outpoint: OutPoint,
    pub output: TxOut,
    pub eth_address: Option<[u8; 20]>,
    pub version: UtxoVersion,
}
```

crates/botanix-wallet-sweep/src/psbt.rs
```diff
@@ -1,6 +1,1330 @@
...
+
+/// Validates Bitcoin value limits to prevent overflow
+fn validate_bitcoin_value_limits(value: Amount, operation: &str) -> Result<(), SweepError> {
+    if value.to_sat() > MAX_BITCOIN_VALUE_SATS {
+        return Err(SweepError::ValueOverflow {
+            operation: operation.to_string(),
+        });
+    }
+    Ok(())
+}
```
DariusParvin: nit: this is more like an 'invalid bitcoin amount' than an 'overflow'

crates/botanix-wallet-sweep/src/psbt.rs
```diff
@@ -1,6 +1,1330 @@
...
+    let inputs = convert_to_sweep_inputs(limited_utxos)?;
+    validate_bitcoin_value_limits(total_input_value, "total input value")?;
+    
+    // Calculate fee using preliminary PSBT
+    let preliminary_output = TxOut {
+        script_pubkey: destination_address.script_pubkey(),
+        value: total_input_value,
+    };
+    let preliminary_psbt = build_preliminary_psbt(&inputs, &preliminary_output);
+    let fee = calculate_transaction_fee(&preliminary_psbt, fee_rate)?;
```
DariusParvin, line 278: @SatwikPrabhu if it's okay with you I think it would be easier to maintain if we use the existing fee calculation methods from the btc-server. The logic is a bit tricky and the methods in btc-server have more extensive unit tests and have gone through review already. If it sounds good with you I can work on updating this bit since I worked with them recently.

crates/botanix-wallet-sweep/src/psbt.rs
DariusParvin: ah okay, so you do have the `final_value` check below 👌 . In that case I think you can remove this `Amount::ZERO` check here and instead just check that `limited_utxos` isn't empty? Or maybe it makes sense to check that after `utxos.truncate(final_count);`.

crates/botanix-wallet-sweep/src/psbt.rs
SatwikPrabhu: I thought of that as well, but the thing is this would create a circular dependency issue since the btc server has the sweep crate as a dependency. Only options are have a separate crate for fee utils or just duplicate the logic. I feel duplicating the logic makes more sense for now as I have I done. Let me double check whether its a perfect duplicate or not. Maybe we can hop on a call?

crates/botanix-wallet-sweep/src/psbt.rs
SatwikPrabhu: As mentioned in https://github.com/botanix-labs/Macbeth/pull/928#discussion_r2279963420, theres a circular dependency issue here

crates/botanix-wallet-sweep/src/psbt.rs
SatwikPrabhu: I think it makes sense to check that after the final count. However, I would still keep the zero check in case the truncated set is non-empty and has zero value. Does that make sense?

crates/botanix-wallet-sweep/src/psbt.rs
SatwikPrabhu: I've added a todo to consider making a new crate for fee utils so that reimplementation is at a minimum

crates/botanix-wallet-sweep/src/psbt.rs
SatwikPrabhu: Done!

crates/botanix-wallet-sweep/src/psbt.rs
SatwikPrabhu: Done!

crates/botanix-wallet-sweep/src/psbt.rs
SatwikPrabhu: Done!

crates/botanix-wallet-sweep/src/psbt.rs
SatwikPrabhu: I've implemented the logic to extract the network from the address in this commit for psbt creation and a todo to remove the `destination_network` in the sweep request since it is out of scope here

bin/reth/src/commands/sweep.rs
SatwikPrabhu: As discussed offline, base64 is also fine. Added a couple of tests to verify the same

crates/botanix-wallet-sweep/src/psbt.rs
```diff
@@ -1,6 +1,1716 @@
...
+/// This function retrieves all available UTXOs and tracked transactions from the database,
+/// then creates a comprehensive emergency sweep PSBT with proper size limits and fee calculation.
+///
+/// # Arguments
+/// * `request` - The wallet sweep request containing destination and fee information
+/// * `client` - Mutable reference to the btc-server client for database access
+///
+/// # Returns
+/// A PSBT that sweeps all available UTXOs to the specified destination address
+pub async fn create_psbt_async(
```
shumkov, line 165: Why do we need async postfix for the function name?

crates/botanix-wallet-sweep/Cargo.toml
```diff
@@ -21,7 +21,7 @@ serde = "1.0.171"
 thiserror = "1.0.43"
 tokio = { version = "1.45.0", features = ["full"] }
 tracing = "0.1.40"
-#btc-server = { path = "../../bin/btc-server" }
+# btc-server = { path = "../../bin/btc-server" } # Circular dependency - btc-server depends on us
```
shumkov, line 24: Yeah, we don't need this dependency. It's commented out. You can just remove this line.

shumkov (approved):

SatwikPrabhu:
Merging this now so that we can move forward with the signing logic. Will circle back to the lint/clippy errors and other smaller stuff in the next PR. Thanks for the review guys

PR #948, opened 2025-08-24
fix(wallet-sweep): ignore sweep in validate outputs

SatwikPrabhu:

Ignore sweep transactions when validating change outputs as the destination is not the federation address

## What was done?

- Use signing psbt type from the network instead of detecting psbt type
- Ignore sweep transactions when validating change outputs as the destination is not the federation address

## How Has This Been Tested?

-

## Breaking Changes

-

## Checklist

## Summary by CodeRabbit

* **New Features**
  * PSBT-type support added: clients can specify/receive a PSBT type (Pegout vs Sweep) and signing flows propagate this type end-to-end.

* **Bug Fixes**
  * Sweep transactions (single-output sweeps) skip change-output script checks to prevent valid sweeps from being rejected.

* **Tests**
  * Tests updated for PSBT-type conversion, backward compatibility, and type-aware validation/signing flows.

* **Chores**
  * Public RPC/messages and session metadata extended; API surfaces and test scaffolding adjusted accordingly.


bin/btc-server/src/util.rs
```diff
@@ -430,6 +430,8 @@ pub enum ValidateOutputsError {
 /// - pegouts have not already been finalized
 /// - there are no duplicate outputs
 pub(crate) fn validate_outputs(psbt: &Psbt, db: &database::Db) -> Result<(), ValidateOutputsError> {
+    let is_sweep = psbt.outputs.len() == 1 && psbt.pegout_ids().is_empty();
```
shumkov: This is hacky way to check this. Better to use enum param

shumkov (changes requested):

bin/btc-server/src/util.rs
SatwikPrabhu: Updated

scottmillner:
I'll let @shumkov take the lead on this as the main reviewer
When you and @SatwikPrabhu feel good about the pr and addressing feedback please merge

shumkov:
@SatwikPrabhu please resove conflicts

bin/btc-server/src/util.rs
```diff
@@ -497,18 +526,18 @@ pub(crate) fn validate_outputs(psbt: &Psbt, db: &database::Db) -> Result<(), Val
         return Err(ValidateOutputsError::DuplicateOutputs);
     }
 
-    // if a change output exists, check if it is valid
+    // if a change output exists, check if it is valid (ignore for sweep transactions)
     if let Some(idx) = change_output {
-        // TxOut scriptpubkey should be scriptpubkey derived from aggregated public key
-        let agg_pk = public_key_package.verifying_key().to_secp_pk().expect("valid secp pk");
-        let expected_script_pubkey = generate_taproot_change_scriptpubkey(&agg_pk);
+        if !is_sweep {
```
DariusParvin, line 531: If i understand correctly, the only check in `validate_outputs` that's relevant for a sweep is `if psbt.outputs.len() != psbt.unsigned_tx.output.len() {`. So i think it might be a bit easier to maintain if after that check, we just return early if it's a sweep.

bin/btc-server/src/coordinator/mod.rs
```diff
@@ -78,6 +80,7 @@ pub fn make_tx(
     outputs: Vec<(TxOut, PegoutId)>,
     fee_rate: FeeRate,
     change_script: ScriptBuf,
+    psbt_type: SigningPsbtType,
```
DariusParvin, line 83: ~Hmm, `make_tx` is a core part of the regular pegout flow, and the majority of the logic in it doesn't apply to sweeps. There are already going to be some merge conflicts with the version in develop too. I think a simpler approach would be to have a new method `make_sweep_tx` which would be a stripped down version of this method.~

Edit: Ah okay i think i misunderstood. It's not that the plan is to use `make_tx` for sweeps, it's just that `validate_outputs` now requires a new argument for `psbt_type`, so it needed to be passed in to `make_tx` as an argument. (is that right?) 

In that case, since the actual validation being done on the sweep output is pretty minimal (just `if psbt.outputs.len() != psbt.unsigned_tx.output.len()` as mentioned in the other comment), I think it might be a bit cleaner to create a new method `validate_sweep_output` that does just that, and you call it where it's needed.

PR #949, opened 2025-08-24
fix(rpc): Remove additional encoding for rpc utxo scripts

SatwikPrabhu:

There was a consistency in the format of the `script` field in `RpcUtxo` and `DbUtxo`. This would be potentially confusing down the line when using a grpc endpoint that uses `RpcUtxo`.

## What was done?

Now the format is consistent, so both structs store the scriptpubkey without the the length prefix. i.e. a p2tr script just starts with `OP_1`, not `0x22`.

To maintain backward compatibility, the method `TryFrom<RpcUtxo> for DbUtxo` will first interpret the RpcUtxo script with the new format, and if the result is not a valid p2tr script, fallback to interpreting it with the prefix. 

## How Has This Been Tested?

- Unit and integration tests

## Breaking Changes

- None

## Checklist

## Summary by CodeRabbit

- New Features
  - Improved compatibility when reading/writing Bitcoin scripts: supports both raw Taproot and legacy-encoded formats, reducing deserialization issues.
- Improvements
  - Added warning logs when legacy script formats are detected to aid operators.
- Bug Fixes
  - More reliable UTXO conversions with robust round-trip handling across script formats.
- Tests
  - Updated test suites to use proper scriptPubKey generation and direct byte encoding.
  - Added tests covering fallback parsing and round-trip conversions.

### DIFF
--- a/bin/btc-server/src/util.rs
+++ b/bin/btc-server/src/util.rs
@@ -746,10 +746,7 @@ mod tests {
         assert!(total_outputs == Amount::ZERO);
 
         let res = validate_psbt(&psbt, NO_FLAGS, 2, &db);
-        assert_eq!(
-            res.unwrap_err(),
-            ValidatePSBTError::FeeSanityCheck(Amount::from_sat(2332), Amount::ZERO)
-        );
+        assert!(matches!(res.unwrap_err(), ValidatePSBTError::FeeSanityCheck(_, Amount::ZERO)));
     }
 
     #[test]
--- a/bin/btc-server/src/util.rs
+++ b/bin/btc-server/src/util.rs
@@ -748,7 +748,7 @@ mod tests {
         let res = validate_psbt(&psbt, NO_FLAGS, 2, &db);
         assert_eq!(
             res.unwrap_err(),
-            ValidatePSBTError::FeeSanityCheck(Amount::from_sat(2346), Amount::ZERO)
+            ValidatePSBTError::FeeSanityCheck(Amount::from_sat(2332), Amount::ZERO)
         );
     }
 
--- a/bin/btc-server/src/database/mod.rs
+++ b/bin/btc-server/src/database/mod.rs
@@ -20,7 +20,7 @@ use bitcoin::{
 use btc_server_client::SigningStatus;
 use frost_secp256k1_tr as frost;
 use futures::Stream;
-use log::info;
+use log::{info, warn};
 use miniscript::psbt::PsbtExt;
 use serde::{Deserialize, Serialize};
 use sled::transaction::{ConflictableTransactionError, TransactionError};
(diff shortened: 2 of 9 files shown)


scottmillner:
will review when pipeline is green

shumkov:
@SatwikPrabhu please resove conflicts

DariusParvin:
I'm quite weary about making this change for the following reasons:
- `RpcUtxo` is used in new_consensus_checkpoint
- the format of scripts in the db is quite critical, so even if there is some inefficiency, it's safer to just leave it as is.
- there's no unit test for the type conversions

I'm guessing from the PR description that the expected behavior was a different encoding of the script when calling the `GetAllUtxos` struct. I don't believe this is being used right now, and was this probably came up while developing the sweep feature.

I think changing the conversion of the `From` conversion, but not the other direction seems problematic, because then there would be an inconsistent definition of what the `script` field in the struct should be. Maybe this fix is actually making it so that the types are consistent, but without unit tests it's hard to tell. 

I think the next best step here (that I will work on) is to add a unit test to do a round conversion, from db utxo -> rpc utxo and back -> db utxo, and verify that nothing was lost. If this change makes that correct, then we can merge, but if not, then maybe it would be better to just deprecate this struct and use a new definition when it's needed?

bin/btc-server/src/util.rs
```diff
@@ -748,7 +748,7 @@ mod tests {
         let res = validate_psbt(&psbt, NO_FLAGS, 2, &db);
         assert_eq!(
             res.unwrap_err(),
-            ValidatePSBTError::FeeSanityCheck(Amount::from_sat(2346), Amount::ZERO)
+            ValidatePSBTError::FeeSanityCheck(Amount::from_sat(2332), Amount::ZERO)
```
rwlockbg: Is this some sort of a magic number here ? Could we add some explanation as to how it is derived ?

rwlockbg (dismissed):

bin/btc-server/src/util.rs
DariusParvin: yeah this is from an old test before we had more proper fee calculation methods. By the way, I think this whole PR is probably not worth it. I'm going to comment on the PR and then close it.

DariusParvin:
In addition to the other comment I wrote above, i just think this change is introducing unnecessary risk for the sake of addressing a small inconsistency between the RpcUtxo and DbUtxo. 

This is effectively changing the grpc contract of what data should be passed in for the scriptPubkey field of the RpcUtxo. That's why a lot of the test inputs have to be updated in this PR, but there are some scenarios i can think of where this happens in prod:
- The btc server and reth server are running different versions
- we miss some part of the code that should have been updated, or we try to revert some changes, without considering how RpcUtxo had changed.

We _could_ add some extra validation or build in some backwards compatibility, but again I think it's not worth introducing that complexity or risk.

DariusParvin:
On second thought I think i'll just add some safety checks in the conversion, so that we dont end up with invalid scripts in the db. The pro of doing this now is that we avoid running into the same issue in the future

bin/btc-server/src/util.rs
DariusParvin: @rwlockbg  that value was the fee rate of the psbt, but that is outside the scope of the unit test. It was just included to match the error message that was returned when the fee is too high. There are other tests for calculating the exact fee rate. 

I've updated this test so that the assertion uses `matches!` on the error type returned, so that it doesnt look like there's a magic number there

scottmillner (approved):

PR #950, opened 2025-08-25
feat: wallet sweep

shumkov:

-

## What was done?

- [ ] integration tests -- Doing (Satwik)
- [x] we enable walletsweeptask only when specific flag is passed to reth poa command
- [x] validate that signing session is present by fed members
- [ ] fix tests, formatting, codespell, to make PR green -- in progress @shumkov 
- [ ] validate request in btc server's accept endpoint (signature, reply attack, etc)
- [ ] remove network from cli, request and session
- [ ] handle errors
- [ ] metrics
- [ ] unit tests
- [ ] docs
- [ ] instructions
- [ ] update implementation spec

## How Has This Been Tested?

-

## Breaking Changes

-

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Emergency Wallet Sweep end-to-end: initiate, accept, PSBT creation, live streaming updates, and abort.
  * New reth CLI commands: sweep initiate, accept-request, psbt, abort.
  * Federation nodes can auto-create/sign sweep PSBTs (controlled by --wallet-sweep-enabled).
  * Improved signing session typing and a FAILED signing status for clearer reporting.

* **Documentation**
  * Added comprehensive Emergency Wallet Sweep design spec.

* **Chores**
  * Storage/DB support and targeted migration list for wallet sweep sessions.


bin/btc-server/src/bin/main.rs
```diff
@@ -1285,25 +1390,42 @@ where
...
+
+        if signing_session_id.is_sweep_session() {
+            trace!(
+                "Signed sweep tx to be broadcast for session {}: {}",
+                signing_session_id,
+                hex::encode(&tx_bytes),
+            );
+        } else {
+            trace!(
+                "Signed pegout tx to be broadcas for session {}: {}",
```
scottmillner, line 1447: nit: typo - `broadcas` > `broadcast`

scottmillner:
just did first pass through the pr

awesome work!!

@shumkov @SatwikPrabhu let's sync thursday on the feature
almost there!!

docs/design/emergency_wallet_sweep_spec.md
lamafab, line 1: Should have led with this :)

docs/design/emergency_wallet_sweep_spec.md
```diff
@@ -0,0 +1,887 @@
...
+
+## 4. The Step-by-Step Emergency Workflow
+
+### (Coordinator) Phase 1: Collect States and Initiate Sweep
+
+- The designated Coordinator operator runs the `initiate` command with federation configuration:  
+  `reth sweep initiate --destination <SECURE_ADDR> --fee-rate <RATE> --consensus-threshold 80 --federation-config <PATH> --coordinator-key <PATH> --db-path <PATH>`
+- This command performs several critical actions:
+    1. **Validate Coordinator Authority**: Verifies that the operator holds the coordinator private key for the active key generation by requiring a cryptographic signature using the coordinator's federation private key.
+    2. **Query Reachable Members**: Directly queries each federation member's `btc-server` via the `GetMemberUtxoState` gRPC endpoint to collect current UTXO state:
```
lamafab, line 58: This makes it sound like the Fed members can directly contact other member's `btc-server` (**which it MUST NOT**). I think this document would generally benefit from a section specifically on the `WalletSweepTask`, and how it integrates into our existing Frost protocol.

crates/botanix-storage/src/models/wallet_sweep.rs
lamafab, line 1: Having to manually encode/decode this stuff is very tedious, and something I'm currently dealing with myself (for the TEM project). While we should keep this consistent for now, in the near future we should consider using something already-established and battle-tested - and yet simple - like the [`parity-scale-codec`](https://crates.io/crates/parity-scale-codec), for example (cc @shumkov).

bin/reth/src/commands/sweep.rs
```diff
@@ -0,0 +1,383 @@
...
+    /// Accept an emergency sweep request to participate in signing session
+    #[command()]
+    AcceptRequest {
+        /// Path to wallet sweep request JSON file
+        #[arg(long, value_parser = parse_file_exists)]
+        request_file_path: PathBuf,
+    },
+    /// Construct PSBT from an emergency sweep request for ingesting or manual signing
+    #[command()]
+    Psbt {
```
lamafab, line 76: Thanks for that!

bin/reth/src/commands/sweep.rs
```diff
@@ -0,0 +1,383 @@
...
+                )?;
+
+                info!(
+                    "Created wallet sweep request for coordinator ID: {}, notifying btc-server",
+                    session_request.coordinator_id
+                );
+
+                // Accept the request on the local btc-server (this creates the session and notifies
+                // other members)
+                session_request.accept(&mut btc_server_client).await?;
```
lamafab, line 250: Here I'd clarify how the `WalletSweepTask` gets involved, and how its' subscription model works (respectively just reference the appropriate type/module).

bin/reth/src/commands/sweep.rs
```diff
@@ -0,0 +1,383 @@
...
+}
+
+impl SweepCommand {
+    /// Validate if we are the coordinator by checking our federation configuration and private key
+    async fn validate_coordinator_authority(
+        request: &WalletSweepRequest,
+        btc_server_client: &mut impl BtcServerExtendedApi,
+    ) -> eyre::Result<bool> {
+        // Get our local identifier from btc-server configuration
+        let public_key_response = btc_server_client
```
lamafab, line 127: This does not work as expected; this is not the local identifier, but rather the aggregated (multisig) public key itself. So when you run `reth sweep accept`, this will attempt to verify the coordinator signature using the aggregated (multisig) public key. Meaning this will always fail.

bin/reth/src/commands/sweep.rs
```diff
@@ -0,0 +1,383 @@
...
+        let our_public_key_bytes: Vec<u8> = FromHex::from_hex(&public_key_response.publickey)
+            .wrap_err("Failed to decode our public key")?;
+        let our_public_key = bitcoin::secp256k1::PublicKey::from_slice(&our_public_key_bytes)
+            .wrap_err("Failed to parse our public key")?;
+
+        // Verify the coordinator signature to check if we are the coordinator
+        let secp = bitcoin::secp256k1::Secp256k1::new();
+
+        // Reconstruct the signature data that was signed
+        let mut signature_data = Vec::new();
```
lamafab, line 142: This signature data pattern is kind of vulnerable to replay-attacks. Since this is a manual process - and there's a `created_at` param, at least - that might be acceptable. But consider tying a session Id do this, if possible , including in `WalletSweepRequest`, and having that set and checked on the `btc-server` side. This applies to `crates/botanix-wallet-sweep/src/request.rs` as well. Your `Db::update_wallet_sweep_session` actually returns the generated session Id, but you don't do anything with it (yet).

Also, consider using `merlin` for signature data construction, it exists for this very purpose - which we already do use in the DKG process:

```rust
// In `bin/btc-server/src/dkg/encryption.rs`

let mut commit = vec![0; 32];
let fed_static = self.fed_members.get(&initiator.0).ok_or(Error::NotAFedMember)?;

// Compute the challenge bytes to be verified against the provided
// signature; notably, we commit the ephemeral key and the round1
// package.
let mut t = self.transcript.clone();
t.append_message(b"frost_id", initiator.0.serialize().as_slice());
t.append_message(b"eph_pub", eph_pub.serialize().as_slice());
t.append_message(b"round1_package", package.serialize()?.as_slice());
t.challenge_bytes(b"round1_commit", &mut commit);
std::mem::drop(t);

// Verify the signature using the public key of the fed member.
let msg = secp256k1::Message::from_digest_slice(&commit).expect("valid size");

if fed_static.verify(&self.secp, &msg, &signature).is_err() {
	return Err(Error::SignatureVerificationFailed);
}
```

crates/consensus/authority/src/frost_task.rs
```diff
@@ -699,13 +731,25 @@ where
...
                                 };
 
-                                if let Err(e) = validate_psbt_by_ids(
-                                    &self.storage.reth_database,
-                                    self.storage.btc_network,
-                                    &psbt_res,
-                                )
-                                .await
-                                {
+                                let validation_result = match signing_session_id.session_type() {
```
lamafab: There might be some risky implications here; this logic allows the sender of the message to control how a PSBT is validated, since the logic decodes the `SigningSessionType` as provided. But both functions are very strict, so it's probably fine.

Second opinions appreciated here.

crates/consensus/authority/src/utils.rs
```diff
@@ -785,6 +785,378 @@ pub fn validate_psbt_id_by_maximum_cutoff_age(
...
+
+    // Basic PSBT structure validation (strict)
+    if psbt.inputs.is_empty() {
+        error!(target: "consensus::authority::utils::validate_sweep_psbt", "Sweep PSBT has no inputs");
+        return Err(PsbtValidationError::FailedToValidatePsbtByIds(String::from(
+            "Sweep PSBT must have at least one input",
+        )));
+    }
+
+    if psbt.outputs.len() != 1 {
```
lamafab, line 829: You're **NOT** checking the actual `psbt.unsigned_tx.output` array, and the intended output destination is not checked either!

crates/consensus/authority/src/wallet_sweep_task.rs
```diff
@@ -0,0 +1,365 @@
...
+            }
+        };
+
+        self.storage
+            .botanix_database_factory
+            .update_wallet_sweep_session(session.clone())
+            .expect("failed to update wallet sweep session");
+    }
+
+    async fn initiate_signing_session(
```
lamafab: Unless I'm mistaken, it looks like everyone initiates this - not just the coordinator?

crates/consensus/authority/src/utils.rs
lamafab: Btw does this method even check whether a sweep request was accepted (via `SweepSubcommands::AcceptRequest`)?

lamafab (commented):
Alright, looks great! Left some comments.

Most of the heavy lifting and coordination happens via the `WalletSweepTask`, which is the most important and critical bit. It looks like it's heavily WIP, so I'm just leaving some higher-level thoughts here:

* First, I appreciate how you mimicked the existing `FroskTask` implementation.
* Is it a good idea that the `btc-server` database kicks-off the sweeping PSBT signing session? I have bad experiences with that _communicate-via-database_ approach.
* I'd like to have a little more theory on how the `WalletSweepTask` should manage the Utxo subset retrieval & communication, and of course the database migration & pegout/pegin backlog (which is probably the hardest part). I guess this will come together with time as the actual implementation progresses.

Seconds opinions appreciated here.

shumkov:
@lamafab Thanks a lot for you review:
1. Actually `WalletSweepTask` is mostly done. There are two main things: update database based on btc server state. Another routine is checking the database state and handle signing session using signing state manager from FrostTask. 
2. Unfortunately, that's the only way if we don't want introduce additional internal API for reth (I believe we don't). So now CLI updates btc server state. WalletSweepTask pulling (via stream) updates of btc server state and updates reth state.
3. WalletSweepTask creates PSBT (create_psbt_async ) and starts singing session using FrostTask's `SingingStateManager` so communicate is happening through FrostTask.

docs/design/emergency_wallet_sweep_spec.md
shumkov: Spec is outdated ))

crates/botanix-storage/src/models/wallet_sweep.rs
shumkov: Agree, bincode is also good candidate.

bin/reth/src/commands/sweep.rs
shumkov: Request goes to btc server's accept_wallet_sweep_session. This methods updates btc server database. WalletSweepTask connects to subscribe_to_wallet_sweep_session_update. This method listens for updates from btc server database and push it it the stream. WalletSweepTask receives the update and modify reth's database in turn.

crates/consensus/authority/src/frost_task.rs
shumkov: But if you manipulate with session ID it won't work since we won't have such session in session manager, isn't?

crates/consensus/authority/src/utils.rs
shumkov: It should but not implemented yet.

crates/consensus/authority/src/wallet_sweep_task.rs
shumkov: This is in #964

scottmillner:
closing as this has been moved to a separate repo

PR #973, opened 2025-09-02
feat(btc-utils): Init binary & implement export/import mechanism for Frost keys

lamafab:
NOTE: `main` branch!

This introduces a convenient mechanism that allows validators to export and import the Frost key package individually (cc @birozuru), primarily intended for backup purposes.

As discussed internally, we created a new binary package titled `btc-utils` - which can also serve as a basis to further build additional extensions and convenience functions ontop of it for anything `btc-server` related, such as adding missing UTXO's, extracting insights, etc (cc @DariusParvin).

```console
$ btc-utils --help
Usage: btc-utils <COMMAND>

Commands:
  export-key-package  Export encrypted key packages to a file
  import-key-package  Import encrypted key packages from a file
  help                Print this message or the help of the given subcommand(s)

Options:
  -h, --help  Print help
```

## Subcommands

```console
$ btc-utils export-key-package --help
Export encrypted key packages to a file

Usage: btc-utils export-key-package [OPTIONS] --db <DB> --output <OUTPUT>

Options:
      --db <DB>                  The path to the database containing key packages
      --output <OUTPUT>          Output file path for the encrypted export
      --passphrase <PASSPHRASE>  Passphrase as parameter instead of prompt
      --allow-empty-passphrase   Allow using an empty passphrase (NOT recommended)
  -h, --help                     Print help
```

```console
$ btc-utils import-key-package --help
Import encrypted key packages from a file

Usage: btc-utils import-key-package [OPTIONS] --db <DB> --input <INPUT>

Options:
      --db <DB>                  The path to the database to store key packages
      --input <INPUT>            Input file path containing the encrypted export
      --passphrase <PASSPHRASE>  Passphrase as parameter instead of prompt
      --force-overwrite          Overwrite existing key packages in the database
  -h, --help                     Print help
```

## Examples

We export the key package from the database to a given file and get prompted for an encryption passphrase:

```console
$ btc-utils export-key-package \
	--db=".local_env/federation/node-3/btc_server/db" \
	--output="exported_key_package.bin"
Passphrase: 
Confirm passphrase: 
Successfully exported encrypted key package to `exported_key_package.bin`
```

When importing, the default behavior is that an existing key package cannot be overwritten unless explicitly allowed:

```console
$ btc-utils import-key-package \
	--db=".local_env/federation/node-3/btc_server/db" \
	--input="exported_key_package.bin"
Error: Existing key package may not be overwritten unless explicitly allowed
```

By passing the `--force-overwrite` parameter, the operation succeeds:

```console
$ btc-utils import-key-package \
	--db=".local_env/federation/node-3/btc_server/db" \
	--input="exported_key_package.bin" \
	--force-overwrite
Passphrase: 
Confirm passphrase: 
Successfully imported decrypted key package
```

### DIFF
--- a/bin/btc-server/src/bin/utils.rs
+++ b/bin/btc-server/src/bin/utils.rs
@@ -1,0 +1,138 @@
+use btcserverlib::database;
+use clap::Parser;
+use std::path::PathBuf;
+use zeroize::Zeroizing;
+
+#[derive(Clone, Debug, Parser)]
+#[command(name = "btc-utils")]
+pub struct Cli {
+    #[command(subcommand)]
+    pub cmd: Commands,
+}
+
+#[derive(Clone, Debug, Parser)]
+pub enum Commands {
+    /// Export encrypted key packages to a file.
+    #[command(name = "export-key-package")]
+    ExportKeyPackage(ExportConfig),
+    /// Import encrypted key packages from a file.
+    #[command(name = "import-key-package")]
+    ImportKeyPackage(ImportConfig),
+}
+
+#[derive(Clone, Debug, Parser)]
+pub struct ExportConfig {
+    /// The path to the database containing key packages.
+    #[arg(long)]
+    pub db: PathBuf,
+    /// Output file path for the encrypted export.
+    #[arg(long)]
+    pub output: PathBuf,
+    /// Passphrase as parameter instead of prompt.
+    #[arg(long)]
+    pub passphrase: Option<Zeroizing<String>>,
+    /// Allow using an empty passphrase (NOT recommended).
+    #[arg(long, default_value_t = false)]
+    pub allow_empty_passphrase: bool,
+}
+
+#[derive(Clone, Debug, Parser)]
+pub struct ImportConfig {
+    /// The path to the database to store key packages.
+    #[arg(long)]
+    pub db: PathBuf,
+    /// Input file path containing the encrypted export.
+    #[arg(long)]
+    pub input: PathBuf,
+    /// Passphrase as parameter instead of prompt.
+    #[arg(long)]
+    pub passphrase: Option<Zeroizing<String>>,
+    /// Overwrite existing key packages in the database.
+    #[arg(long, default_value_t = false)]
+    pub force_overwrite: bool,
+}
+
+fn get_passphrase(
+    provided: Option<Zeroizing<String>>,
+    do_confim: bool,
+) -> anyhow::Result<Zeroizing<String>> {
+    match provided {
+        Some(p) => Ok(p.trim().to_string().into()),
+        None => {
+            // Prompt user if CLI was not provided.
+            let p1: Zeroizing<String> = rpassword::prompt_password("Passphrase: ")?.into();
+
+            if do_confim {
+                let p2: Zeroizing<String> =
+                    rpassword::prompt_password("Confirm passphrase: ")?.into();
+
+                if p1 != p2 {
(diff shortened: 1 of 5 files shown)


lamafab:
Actually; the master key should be wrapped in `Zeroize`, will update - **EDIT**: Done!

bin/btc-server/src/database/mod.rs
```diff
@@ -90,6 +92,23 @@ impl FinalizedPegout {
     }
 }
 
+/// Encrypted key package export format for secure backup and transfer.
+///
+/// This structure contains both secret and public key packages encrypted with a
+/// passphrase-derived key. A single random nonce is used for both encryption
+/// operations, with two separately derived keys.
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
+pub struct ExportedKeyPackage {
```
DariusParvin, line 104: is it worth including any information about the encryption algorithm in this struct, or a version number?

bin/btc-server/src/database/mod.rs
```diff
@@ -90,6 +92,23 @@ impl FinalizedPegout {
...
+    /// Random 96-bit nonce used for both encryption operations, in plaintext.
+    pub iv: [u8; 12],
```
DariusParvin, line 109: I'm asking more to improve my understanding of cryptography. Purely in terms of the naming - I would have expected this field to be called a 'nonce', which is then used to derive the salt - for the passphrase, and the iv - for the encryption

DariusParvin (dismissed):
Very nice! Just a couple of questions

bin/btc-server/src/database/mod.rs
lamafab: Yes, I think there should be a version number at least, we don't need info on encryption algorithm as of now. With a version number we don't have to hack around different struct messages in case we ever change it. Will update.

bin/btc-server/src/database/mod.rs
lamafab: Imo this is just about semantics; both "iv" (initialization vector) and "nonce" are often used interchangeably, but generally "nonce" implies that it's used as an incremental counter, meanwhile "iv" is just a random/unique value. In our case, we generate a purely random iv/nonce once, and keep it fixed - we never increment it (we derive individual subkeys instead).

Using it as a salt for the KDF is optional, but since we're already tracking it then we might just as well use it there as well :)

bin/btc-server/src/bin/utils.rs
```diff
@@ -0,0 +1,138 @@
...
+    #[arg(long)]
+    pub passphrase: Option<Zeroizing<String>>,
+    /// Overwrite existing key packages in the database.
+    #[arg(long, default_value_t = false)]
+    pub force_overwrite: bool,
+}
+
+fn get_passphrase(
+    provided: Option<Zeroizing<String>>,
+    do_confim: bool,
```
scottmillner: nit typo > `do_confim` > `do_confirm`

bin/btc-server/src/database/mod.rs
DariusParvin: ohhh yes i see that makes sense. thanks for explaining!

scottmillner (dismissed):
nice clean implementation!
easy to reason about and efficient 🔥

scottmillner (approved):

DariusParvin (approved):
🚀

scottmillner:
note: the failing minting.sol test has been fixed in the develop branch

PR #986, opened 2025-09-23
feat(btc-utils): Implement `compute-gateway-address` functionality

lamafab:
Mentioned in internal comms. Even if we don't end up using it for our specific use-case, it's still useful to have.

The test data is from a recent pegin issue we had to debug (`#bug-reports` channel).

## Subcommand

```console
$ btc-utils compute-gateway-address --help
Compute the gateway pegin address for a given Botanix address

Usage: btc-utils compute-gateway-address [OPTIONS] --botanix-address <BOTANIX_ADDRESS>

Options:
      --botanix-address <BOTANIX_ADDRESS>
          The Botanix (ETH) address to whom the BTC should be minted to
      --aggregated-key <AGGREGATED_KEY>
          Custom aggregated public key of a multisig federation. Uses the genesis federation key on mainnet by default
  -h, --help
          Print help
```

## Example

```console
$ btc-utils compute-gateway-address --botanix-address "0xE99F129Bb9d60a91f6d0Ae2d6bBC746C52A87220"
bc1pdxrrvdqunlrwnt4qrqs52djk0g4r39s0f5qsa0j9dcctq6jvaf5qu9na42
```

## Summary by CodeRabbit

- **New Features**
  - Added a CLI command to compute a Bitcoin Taproot gateway address from a Botanix (ETH) address and an optional aggregated key; uses a built-in genesis key when none is supplied. The command prints the computed gateway address and returns clear errors for invalid input.

- **Tests**
  - Added tests for default-key behavior, custom-key handling, and malformed-key error handling.

### DIFF
--- a/bin/btc-server/src/bin/utils.rs
+++ b/bin/btc-server/src/bin/utils.rs
@@ -1,7 +1,20 @@
-use btcserverlib::database;
+use btcserverlib::{
+    database,
+    util::parse_eth_address,
+    wallet::address::{generate_taproot_address, generate_tweaked_public_key},
+};
 use clap::Parser;
 use std::path::PathBuf;
 use zeroize::Zeroizing;
+
+/// The aggregated Frost public key of the genesis federation.
+///
+/// > GENESIS_AGGR_KEY =
+/// > hex::decode("03ae26f6152efa6e65619f436aae5076356cacab97bed10c294a38b777efa66e72")
+const GENESIS_AGGR_KEY: &[u8] = &[
+    3, 174, 38, 246, 21, 46, 250, 110, 101, 97, 159, 67, 106, 174, 80, 118, 53, 108, 172, 171, 151,
+    190, 209, 12, 41, 74, 56, 183, 119, 239, 166, 110, 114,
+];
 
 #[derive(Clone, Debug, Parser)]
 #[command(name = "btc-utils")]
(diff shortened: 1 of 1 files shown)


bin/btc-server/src/bin/utils.rs
```diff
@@ -132,7 +177,58 @@ async fn main() -> anyhow::Result<(), anyhow::Error> {
 
             println!("Successfully imported decrypted key package");
         }
+        Commands::ComputeGatewayAddress(c) => {
+            let gateway_address = c.compute_gateway_address()?;
+            println!("{}", gateway_address.to_string());
+        }
```
DariusParvin, line 183: is it worth getting the aggregate pubkey from the db, just to check it matches `GENESIS_AGGR_KEY` first? just in case with dynafed or something it changes in the future and we forget to update this.

DariusParvin (approved):
nice! I just one question

bin/btc-server/src/bin/utils.rs
lamafab: We do not know yet how pegins under DynaFed will work; simply pulling "the" key from the database is tricky since there will be multiple. However, if we do that comparison we'd at least spot that something is wrong, so yeah probably a good idea.

scottmillner (approved):

PR #990, opened 2025-09-25
feat(btc-server): add flag excluded_eth_addresses for coin selection

DariusParvin:
https://github.com/botanix-labs/botanix/issues/1108

## What was done?

- Added a flag to btc-server `excluded_eth_addresses` that takes in a list of eth addresses.
- When performing coin selection for creating pegout txs, utxos associated with that eth address will be excluded from coin selection

## How Has This Been Tested?

- Unit test locally and will test on devnet.

## Breaking Changes

- None

## Checklist

-   [] I have tested my changes running a local network, and they work as expected

## Summary by CodeRabbit

- **New Features**
  - CLI/config option to supply comma-separated Ethereum addresses to exclude from coin selection; invalid addresses are rejected with clear validation.
  - Transaction creation respects the exclude list, omitting UTXOs associated with excluded addresses and logging counts before/after filtering.
  - Startup logs the number of configured excluded addresses.

- **Tests**
  - Unit tests added for address parsing and UTXO-filtering behavior.

### DIFF
--- a/bin/btc-server/src/config.rs
+++ b/bin/btc-server/src/config.rs
@@ -197,6 +197,9 @@ pub struct CliConfig {
     /// http port
     #[arg(long)]
     metrics_port: Option<u16>,
+    /// Comma-separated list of Ethereum addresses to exclude from coin selection
+    #[arg(long)]
+    excluded_eth_addresses: Option<String>,
 }
 
 #[derive(Clone, Debug)]
@@ -239,6 +242,8 @@ pub struct Config {
     pub fee_rate_diff_percentage: u32,
     /// Fall back fee rate expressed in sat per vbyte
     pub fall_back_fee_rate_sat_per_vbyte: u64,
+    /// Ethereum addresses to exclude from coin selection
+    pub excluded_eth_addresses: Vec<String>,
 }
 
 pub fn load_config() -> Result<Config, Error> {
@@ -263,6 +268,10 @@ pub fn load_config() -> Result<Config, Error> {
         metrics_port: cli_config.metrics_port,
         fee_rate_diff_percentage: cli_config.fee_rate_diff_percentage.unwrap_or(2),
         fall_back_fee_rate_sat_per_vbyte: cli_config.fall_back_fee_rate_sat_per_vbyte.unwrap_or(10),
+        excluded_eth_addresses: cli_config
+            .excluded_eth_addresses
+            .map(|s| s.split(',').map(|addr| addr.trim().to_string()).collect())
+            .unwrap_or_default(),
     };
     Ok(config)
 }
--- a/bin/btc-server/src/config.rs
+++ b/bin/btc-server/src/config.rs
@@ -11,6 +11,8 @@ use thiserror::Error;
 use tokio::{fs::File, io::AsyncReadExt};
 use url::Url;
 
+use crate::util::{parse_eth_address, ParsingError};
+
 #[derive(Debug, DisplayDoc, Error)]
 pub enum Error {
     /// Open config file: {0}
@@ -27,6 +29,8 @@ pub enum Error {
     Confy(ConfyError),
     /// Missing config element: {0}
     MissingConfigElement(&'static str),
+    /// Invalid Ethereum address: {0}
+    InvalidEthereumAddress(ParsingError),
 }
 
 #[derive(Debug, Default, Deserialize, Clone)]
@@ -141,6 +145,11 @@ async fn read_to_string(path: impl AsRef<Path> + Send) -> Result<String, Error> 
     String::from_utf8(contents).map_err(Error::ParseUtf8)
 }
 
+/// Clap value parser for Ethereum addresses
+fn parse_ethereum_address_for_clap(s: &str) -> Result<[u8; 20], String> {
+    parse_eth_address(s.to_string()).map_err(|e| e.to_string())
+}
+
 // Cli args and config
 
 #[derive(Clone, Debug, Parser)]
@@ -198,8 +207,8 @@ pub struct CliConfig {
     #[arg(long)]
     metrics_port: Option<u16>,
     /// Comma-separated list of Ethereum addresses to exclude from coin selection
-    #[arg(long)]
-    excluded_eth_addresses: Option<String>,
+    #[arg(long, value_delimiter = ',', value_parser = parse_ethereum_address_for_clap)]
+    excluded_eth_addresses: Vec<[u8; 20]>,
 }
 
 #[derive(Clone, Debug)]
@@ -243,7 +252,7 @@ pub struct Config {
     /// Fall back fee rate expressed in sat per vbyte
     pub fall_back_fee_rate_sat_per_vbyte: u64,
     /// Ethereum addresses to exclude from coin selection
-    pub excluded_eth_addresses: Vec<String>,
+    pub excluded_eth_addresses: Vec<[u8; 20]>,
 }
 
 pub fn load_config() -> Result<Config, Error> {
@@ -268,10 +277,37 @@ pub fn load_config() -> Result<Config, Error> {
         metrics_port: cli_config.metrics_port,
         fee_rate_diff_percentage: cli_config.fee_rate_diff_percentage.unwrap_or(2),
         fall_back_fee_rate_sat_per_vbyte: cli_config.fall_back_fee_rate_sat_per_vbyte.unwrap_or(10),
-        excluded_eth_addresses: cli_config
-            .excluded_eth_addresses
-            .map(|s| s.split(',').map(|addr| addr.trim().to_string()).collect())
-            .unwrap_or_default(),
+        excluded_eth_addresses: cli_config.excluded_eth_addresses,
     };
     Ok(config)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_parse_ethereum_address_for_clap_valid() {
+        let result = parse_ethereum_address_for_clap("0x1234567890123456789012345678901234567890");
+        assert!(result.is_ok());
+
+        // Test without 0x prefix
+        let result = parse_ethereum_address_for_clap("1234567890123456789012345678901234567890");
+        assert!(result.is_ok());
+    }
+
+    #[test]
+    fn test_parse_ethereum_address_for_clap_invalid() {
+        // Test invalid hex
+        let result = parse_ethereum_address_for_clap("invalid");
+        assert!(result.is_err());
+
+        // Test wrong length
+        let result = parse_ethereum_address_for_clap("0x1234");
+        assert!(result.is_err());
+
+        // Test empty string
+        let result = parse_ethereum_address_for_clap("");
+        assert!(result.is_err());
+    }
 }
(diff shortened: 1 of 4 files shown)


bin/btc-server/src/coordinator/mod.rs
```diff
@@ -191,12 +230,13 @@ pub fn attempt_make_tx(
...
+            filtered_utxos
+                .get(op)
+                .ok_or_else(|| CoordinatorError::MissingUtxoForConflictingInput)
+                .map(|u: &Utxo| {
                     conflicting_utxos.insert(*op, u.clone());
                     u.clone()
-                },
-            )
+                })
         })
```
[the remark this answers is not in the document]
DariusParvin: this situation shouldn't happen, but i suppose it's possible. There's a slight performance benefit to using `filtered_utxos` but i guess it's safer to keep it as `uxto` to prevent double spends.

bin/btc-server/src/config.rs
```diff
@@ -263,6 +268,10 @@ pub fn load_config() -> Result<Config, Error> {
         metrics_port: cli_config.metrics_port,
         fee_rate_diff_percentage: cli_config.fee_rate_diff_percentage.unwrap_or(2),
         fall_back_fee_rate_sat_per_vbyte: cli_config.fall_back_fee_rate_sat_per_vbyte.unwrap_or(10),
+        excluded_eth_addresses: cli_config
+            .excluded_eth_addresses
+            .map(|s| s.split(',').map(|addr| addr.trim().to_string()).collect())
+            .unwrap_or_default(),
     };
```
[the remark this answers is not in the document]
DariusParvin: this is a good suggestion, i'll add a warning or log

rwlockbg (dismissed):

bin/btc-server/src/config.rs
```diff
@@ -263,6 +268,10 @@ pub fn load_config() -> Result<Config, Error> {
         metrics_port: cli_config.metrics_port,
         fee_rate_diff_percentage: cli_config.fee_rate_diff_percentage.unwrap_or(2),
         fall_back_fee_rate_sat_per_vbyte: cli_config.fall_back_fee_rate_sat_per_vbyte.unwrap_or(10),
+        excluded_eth_addresses: cli_config
+            .excluded_eth_addresses
+            .map(|s| s.split(',').map(|addr| addr.trim().to_string()).collect())
+            .unwrap_or_default(),
```
lamafab: Do note that `clap` has built-in support for that: https://docs.rs/clap/4.5.48/clap/struct.Arg.html#method.value_delimiter

bin/btc-server/src/coordinator/mod.rs
```diff
@@ -144,6 +180,9 @@ pub fn attempt_make_tx(
     debug!("utxos len = {:?}", utxos.len());
     debug!("utxos = {:?}", utxos);
 
+    // Exclude UTXOs that have been specifically requested to not be included in the coin selection
+    let filtered_utxos = filter_excluded_utxos(utxos.clone(), &config.excluded_eth_addresses);
```
lamafab: Why not just:
```rust
let utxos = filter_excluded_utxos(utxos, &config.excluded_eth_addresses);
```
Without the cloning and renaming?

bin/btc-server/src/config.rs
```diff
@@ -239,6 +242,8 @@ pub struct Config {
     pub fee_rate_diff_percentage: u32,
     /// Fall back fee rate expressed in sat per vbyte
     pub fall_back_fee_rate_sat_per_vbyte: u64,
+    /// Ethereum addresses to exclude from coin selection
+    pub excluded_eth_addresses: Vec<String>,
```
lamafab: Imo those should be validated entries after passing through `fn parse_eth_address`, and throw error immediately.

bin/btc-server/src/coordinator/mod.rs
```diff
@@ -25,6 +26,38 @@ pub mod error;
...
+    if excluded_addresses.is_empty() {
+        info!("No excluded eth addresses provided, returning all utxos");
+        return utxos;
+    }
+
+    info!("Filtering out excluded eth addresses: {:?}", excluded_addresses);
+
+    // Parse excluded addresses from hex strings to byte arrays for comparison
+    let excluded_address_bytes: Vec<[u8; 20]> =
+        excluded_addresses.iter().filter_map(|addr| parse_eth_address(addr.clone()).ok()).collect();
```
lamafab: Imo the parsed and validated ETH addresses should already be ready in `Config.excluded_eth_addresses` (see comment above), that way we spot early whether something is wrong. Additionally, this `.filter_map()` approach **quietly ignores invalid addresses**; the `Config` parser on startup should complain immediately if a passed-on address/input is invalid

bin/btc-server/src/coordinator/mod.rs
```diff
@@ -25,6 +26,38 @@ pub mod error;
...
+    // Filter out UTXOs associated with excluded Ethereum addresses
+    let filtered_utxos = utxos
+        .clone()
+        .into_iter()
+        .filter(|(_, utxo)| {
+            // Keep UTXO if it has no eth_address (change UTXO) or if its address is not in excluded
+            // list
+            utxo.eth_address.map_or(true, |addr| !excluded_address_bytes.contains(&addr))
+        })
+        .collect::<HashMap<_, _>>();
```
lamafab: Consider just using `HashMap::retain()`.

bin/btc-server/src/coordinator/mod.rs
DariusParvin: yeah the issue is that later on there's code regarding selecting conflicting inputs that technically _should_ use the full utxo list. There is an edge case where a psbt is created, then the utxo is added to the exclusion list, and now if the same pegout from before is attempted, the conflicting utxo logic will say there are no `required` utxos, and a double spend becomes possible. 

It is a super edge case where the excluded eth address is added (too late), and the pegout is retried.

I don't have much intuition about how much of a performance impact the `clone()` has. I'm happy to change it to the way you suggested if you think it's better.

lamafab: Nah not a big deal, better safe than sorry :)

bin/btc-server/src/coordinator/mod.rs
```diff
@@ -317,3 +349,58 @@ pub async fn finalize_signing(
...
+
+        println!("result = {:?}", result);
+
+        // The UTXO should be filtered out
+        assert_eq!(result.len(), 2);
+        assert!(!result.contains_key(&outpoint_to_filter));
+        assert!(result.contains_key(&outpoint_to_keep));
+        assert!(result.contains_key(&outpoint_change));
+    }
+}
```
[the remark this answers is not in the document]
DariusParvin: done

bin/btc-server/src/coordinator/mod.rs
```diff
@@ -82,6 +108,7 @@ pub fn make_tx(
     db: &Db,
     min_signers: u16,
     tracked_txs: Vec<Tx>,
+    config: &Config,
 ) -> Result<Psbt, CoordinatorError> {
```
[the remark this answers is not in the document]
DariusParvin: i think coderabbit is making a mistake here. `&self.config` is already being passed in. It seems like coderabbit is only recognising the first two arguments when there are a lot more for `make_tx` and `attempt_make_tx`

bin/btc-server/src/config.rs
```diff
@@ -27,6 +29,8 @@ pub enum Error {
     Confy(ConfyError),
     /// Missing config element: {0}
     MissingConfigElement(&'static str),
+    /// Invalid Ethereum address: {0}
+    InvalidEthereumAddress(ParsingError),
```
lamafab: Afaik this is not used anywhere?

bin/btc-server/src/coordinator/mod.rs
```diff
@@ -82,6 +111,7 @@ pub fn make_tx(
     db: &Db,
     min_signers: u16,
     tracked_txs: Vec<Tx>,
+    config: &Config,
```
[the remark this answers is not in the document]
DariusParvin: same as the other comment, i think coderabbit is making a mistake here as it's not recognising the full list of arguments, which _does_ include `&self.config`

bin/btc-server/src/coordinator/mod.rs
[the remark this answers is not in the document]
lamafab: Wtf?

bin/btc-server/src/coordinator/mod.rs
```diff
@@ -25,6 +26,34 @@ pub mod error;
...
 const MIN_RELAY_FEE_RATE_SAT_VB: u64 = 1;
 
+/// Filters out UTXOs associated with excluded Ethereum addresses
+fn filter_excluded_utxos(
+    utxos: &HashMap<OutPoint, Utxo>,
+    excluded_addresses: &[[u8; 20]],
+) -> HashMap<OutPoint, Utxo> {
+    if excluded_addresses.is_empty() {
+        info!("No excluded eth addresses provided, returning all utxos");
+        return utxos.clone();
```
lamafab, line 36: You could avoid this unnecessary clone by checking `excluded_addresses.is_empty()` **before** calling `filter_excluded_utxos()`. Just a nitpick, of course.

lamafab (dismissed):

DariusParvin:
these changes should be merged into `hotfix`, so this PR is being replaced by https://github.com/botanix-labs/Macbeth/pull/991

bin/btc-server/src/coordinator/mod.rs
DariusParvin: oh yes good point!

bin/btc-server/src/coordinator/mod.rs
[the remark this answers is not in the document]
DariusParvin: yeah lol it's confusing how the AI could make such an error? Especially when at the same time it was able to point out the very edge case about doing the conflict utxo logic on filtered utxos.

bin/btc-server/src/coordinator/mod.rs
DariusParvin: actually, i tried refactoring it and it seems like the clone is still necessary because `utxos` is still needed in addition to the filtered list (because of the conflicting inputs logic). There probably is a bigger refactor possible to remove the cloning, like doing the conflicting input logic before the filtering, but that would be a bigger change. There are already conflicts in the part of the code between `develop` and `main` so yeah probably better to just leave it as is.

DariusParvin:
actually, i'll just close this PR and re-create it, cherry picking from what's going into the hotfix branch.

bin/btc-server/src/config.rs
```diff
@@ -141,6 +143,11 @@ async fn read_to_string(path: impl AsRef<Path> + Send) -> Result<String, Error>
     String::from_utf8(contents).map_err(Error::ParseUtf8)
 }
 
+/// Clap value parser for Ethereum addresses
+fn parse_ethereum_address_for_clap(s: &str) -> Result<[u8; 20], String> {
+    parse_eth_address(s.to_string()).map_err(|e| e.to_string())
+}
+
```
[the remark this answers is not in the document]
DariusParvin: maybe we can update `parse_eth_address` in the future to allow `0X` as a prefix, but it's not a priority for now.

scottmillner (approved):

PR #1019, opened 2025-10-27
fix(pegouts): remove all finalized pegouts from pending pegouts list

scottmillner:

When the coordinator is creating a psbt, it checks if any of the outputs are already finalized, and returns an error if so.  It also removes the first occurrence of a finalized pegout being found from the pending pegout list.  We need to remove all occurrences found instead of just the first one.

## What was done?

- Remove all occurrences of finalized pegouts found in the psbt

## How Has This Been Tested?

- All current unit and int tests pass

## Breaking Changes

- None

## Checklist

### DIFF
--- a/bin/btc-server/src/util.rs
+++ b/bin/btc-server/src/util.rs
@@ -1642,4 +1642,65 @@ mod tests {
         assert!(res_error
             .contains("error validating outputs: found already finalized psbt pegouts in db"));
     }
+
+    #[test]
+    fn test_validate_outputs_should_remove_all_finalized_pegouts_from_pending_pegouts() {
+        let (db, _temp_dir) = setup_db();
+        let (shares, pk_package) = trusted_dealer_setup(2, 2);
+        let key_package = frost::keys::KeyPackage::try_from(shares[&frost_id!(1)].clone())
+            .expect("valid key package");
+
+        // Add the key packages
+        db.set_pubkey_package(pk_package.clone()).expect("set public key package");
+        db.set_key_package(key_package.clone()).expect("set key package");
+
+        // create pegouts and store them as finalized and pending pegouts
+        let mut pending_pegouts = Vec::new();
+        let mut finalized_pegouts = Vec::new();
+        let mut valid_pending_pegouts_count = 0;
+        for i in 0..5 {
+            let pegout_id = PegoutId::new(rand::thread_rng().gen::<[u8; 32]>(), 0);
+            let pending_pegout = PegoutRequest {
+                spk: ScriptBuf::new(),
+                value: Amount::from_sat(1000),
+                id: pegout_id,
+                botanix_height: 0,
+            };
+            pending_pegouts.push(pending_pegout);
+            // Store every odd pegout as finalized so some pending pegouts are valid and should not
+            // be removed during validate_outputs()
+            if i % 2 == 0 {
+                valid_pending_pegouts_count += 1;
+                continue;
+            }
+            let finalized_pegout = FinalizedPegout { id: pegout_id, block_number: 100 };
+            finalized_pegouts.push(finalized_pegout);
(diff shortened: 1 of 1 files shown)


thobbyAk (dismissed):

bin/btc-server/src/util.rs
```diff
@@ -469,20 +469,22 @@ pub(crate) fn validate_outputs(psbt: &Psbt, db: &database::Db) -> Result<(), Val
...
+    // If any matching finalized pegouts were found, remove them all at once and return error
+    if !matching_finalized_pegouts.is_empty() {
+        for id in &matching_finalized_pegouts {
             info!("Finalized pegout id txid hash: {}", hex::encode(id.txid));
             info!("Finalized pegout id tx receipt log index: {}", id.idx);
-            db.remove_pending_pegout(&[*id]).ok();
-            return Err(ValidateOutputsError::AlreadyFinalizedPegouts(vec![*id]));
         }
+        db.remove_pending_pegout(&matching_finalized_pegouts).ok();
+        return Err(ValidateOutputsError::AlreadyFinalizedPegouts(matching_finalized_pegouts));
```
DariusParvin, line 487: this return is in the for loop, so I think it would still error out early and only remove one pending pegout?

DariusParvin (changes requested):
Just because of how critical/consequential this part of the code is, i think we should have at least a unit test for it. We also dont want to make any mistakes here when merging this change back into develop. I can add a test for it if you want

lamafab (dismissed):

bin/btc-server/src/util.rs
lamafab: Looks okay to me? It clearly removes all matching finalized pegouts - **outside** the for-loop.

DariusParvin: ah yeah you're right

scottmillner:
> Just because of how critical/consequential this part of the code is, i think we should have at least a unit test for it. We also dont want to make any mistakes here when merging this change back into develop. I can add a test for it if you want

yeah good point, i'll add the test, thanks!

DariusParvin (dismissed):
🙌

DariusParvin (approved):

thobbyAk (approved):

bin/btc-server/src/util.rs
```diff
@@ -1640,4 +1642,65 @@ mod tests {
...
+            .expect("to store pending pegouts");
+        let finalized_pegouts_refs: Vec<&FinalizedPegout> = finalized_pegouts.iter().collect();
+        db.store_finalized_pegout_ids_atomically(finalized_pegouts_refs.as_slice())
+            .expect("to store finalized pegouts");
+
+        // create a psbt with the finalized pegout ids
+        let mut psbt = create_psbt(1, finalized_pegouts.len(), None);
+        for (i, finalized_pegout) in finalized_pegouts.iter().enumerate() {
+            psbt.outputs[i].set_pegout_id(finalized_pegout.id.as_bytes());
+        }
```
lamafab, line 1690: Technically, this still creates outputs where **all** pegouts are finalized, not just every odd index. Consider using `pending_pegouts.len()` when calling `fn create_psbt`, instead.

lamafab (approved):

bin/btc-server/src/util.rs
scottmillner: this test checks two things:
a psbt fails validation if an output is in the finalized pegouts list
and all and only finalized pegouts are removed from the pending pegouts lists

i believe this test achieves both and the suggested change doesn't impact this

this comment:
```
// Store every odd pegout as finalized so some pending pegouts are valid and should not
// be removed during validate_outputs()
```
is referring to the stored pending pegouts not the outputs in the psbt

PR #1023, opened 2025-11-07
fix(pegout-scheduler): prevent duplicate pegout indices for same spk

DariusParvin:
…ination

## What was done?

-

## How Has This Been Tested?

-

## Breaking Changes

-

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where multiple pegout transactions with identical destinations and amounts could be incorrectly matched to the same output. The system now properly validates both destination and amount, ensuring each transaction output is used only once during processing.

### DIFF
--- a/bin/btc-server/src/pegout_scheduler/mod.rs
+++ b/bin/btc-server/src/pegout_scheduler/mod.rs
@@ -308,20 +308,32 @@ impl PegoutScheduler {
         timestamp: SystemTime,
     ) -> &Tx {
         let pegout_idxs = {
-            let mut ret = Vec::with_capacity(pegouts.len());
             info!(
                 "PegoutScheduler::add_tx: tx output length: {}, pegouts length: {}",
                 tx.output.len(),
                 pegouts.len()
             );
-            // TODO: the same pegout could be in the tx multiple times
+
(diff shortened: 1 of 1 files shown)


bin/btc-server/src/pegout_scheduler/mod.rs
```diff
@@ -308,20 +308,32 @@ impl PegoutScheduler {
...
+            let mut used_indices = HashSet::new();
+
             for pegout in pegouts {
                 let pegout_txout = pegout.txout();
                 let idx = tx
                     .output
                     .iter()
-                    .position(|txout| txout.script_pubkey == pegout_txout.script_pubkey)
+                    .enumerate()
+                    .find(|(idx, txout)| {
```
rwlockbg, line 328: I believe you might want to use `filter_map` here - it is more convenient

DariusParvin:
When working on this PR i realised a couple of things:
1 - there's another part of the code that needs to be updated for the indexing of Outpoint -> Pegout ID to work properly.
2 - This is all essentially dead code that is not used (other than printing misleading warnings). So the better solution is probably to just remove it. I'll close this PR and update the issue.

